PR #235, opened 2014-06-16
Ability to read binary keys and messages.

eirc:
Extended the ability of reading ASCII armored keys and messages to reading binary gpg files too. There are a couple issues here I would like some help addressing.
- `mocha_phantomjs` tests fail
  - It looks like the `require('fs')` I did at the beginning of the test suite returns an empty object and `fs` does not have the `readFile` method. I'm a total nodejs noob so I'm not sure what's going on here...
- Code duplication
  - The `readBinary` methods I added are mostly duplicated code from the `readArmored` methods since they do the exact same thing with the exception of the dearmoring part at the beginning. Extracting the common code is not a trivial task unfortunately though and would require some small refactoring due the heavy usage of exceptions in these methods. Also there are some TODOs left in the `readArmored` of `Message` which note the fact that that exception pattern is not used there. So before jumping in refactoring I'd like some feedback on whether there are some thoughts on how it should be done.
- No tests for `Message`
  - There are no tests for the Message class at all so I wasn't sure if there's some reason for that.
- Fixture files
  - Since no other fixture files were used in the project I didn't know where to add them so maybe their location at `test/general/fixtures/` should be changed?

PS: I have already used the methods I added by monkey patching them in my code in a project I'm working on  so just want to shamelessly plug it :) https://github.com/eirc/pass.js


barrydegraaff:
It is really a shame this was never merged, reading binary messages would be great for interoperability.

emadow:
So there's zero support to read binary messages? Has anyone come up with a workaround?

bartbutler:
Yes, there is support for binary messages. This PR is ancient.

emadow:
My apologies @bartbutler -- the documentation leaves this out. Any resource you can point me to so I can better understand the implementation?

barrydegraaff:
//get html file select elements (with multiselect)
      var fileInputs = document.getElementsByClassName("fileInputPgpAttach");
      for (var inputIndex = 0; inputIndex < fileInputs.length; inputIndex++) 
      {                
         for (var multiselectIndex = 0; multiselectIndex < fileInputs[inputIndex].files.length; multiselectIndex++)           
         {
            (function(file) {
               var name = file.name;
               var reader = new FileReader();  
               reader.onload = function(e) 
               {  
                  options = {
                     data: new Uint8Array(reader.result), 
                     publicKeys: pubKeys,   // for encryption
                     privateKeys: privKey,  // for signing (optional)
                     armor: false 
                  };
                  
                  openpgp.encrypt(options).then(function (message) {
                     // the result is in:         
                     var result = message.message.packets.write();                           
                  }, 
                  function(err) {
                   // some error handler
                   });
               }
               reader.readAsArrayBuffer(file);
            })(fileInputs[inputIndex].files[multiselectIndex]);               
         }                      
      }

barrydegraaff:
The above example shows how to encrypt a binary. The trick is to use readAsArrayBuffer, if you use something else, you must eventually cast it as Uint8Array in `options` to openpgp.js. Using Uint8Array and set armor to false will instruct openpgp.js to do a binary.

barrydegraaff:
Decrypt example:

      options = {
          message: message,           // parse encrypted bytes
          publicKeys: pubKey,         // for verification (optional)
          privateKey: privKey,        // for decryption
          format: 'binary'
      };
      openpgp.decrypt(options).then(function (binary) {
          /result is here:
         var result = binary.data;
      });

This was all added in the 2.x release.

barrydegraaff:
@emadow see above.

tanx:
Also see the examples in the current current for full roundtrip handling of binary data.

emadow:
Thanks, I eventually got it to work. The problem was that I wasn't casting the message as a Uint8Array. I appreciate it!

PR #272, opened 2014-12-02
Compression Support

laktak:
I've added compression support (for zip & zlib) using https://github.com/imaya/zlib.js.

This also fixes decompression for large data (images) that jxg.js couldn't handle (and a compData vs compdata typo).

I'm not sure how you want to handle the dependency so I've included the source like you did with jxg.


derverzweifler:
Absolutely awesome! Is someone working on adding compression support for outgoing messages to high level API? Some kind of config switch?

puzrin:
I'd recomment to take a look at https://github.com/nodeca/pako for inflate/deflate

Issue #269, opened 2014-11-17
Unclear Dependency on Web Workers


bantu:
It is unclear whether Web Workers are an actual requirement or not. The Dependencies section in README.md (see below) makes it sound like Workers are not required when `window.crypto.subtle` is available. Even though subtle is available in my browser, I get an "You need to set the worker path!" error. I think this should be clearified.

> Dependencies
OpenPGP.js only supports browsers that implement window.crypto.getRandomValues. Also, if the browsers support native WebCrypto via the window.crypto.subtle api, this will be used. Though this can be deactivated by setting config.useWebCrypto = false. In this case the library will fall back to Web Worker operations if the initWorker(workerPath) is set.

tanx:
Worker is purely optional. Please see the README and docs

PR #261, opened 2014-10-24
fix worker instance misconfiguration

ratik:
Hi

I've added passing config into the worker. Could you please review.

Thanks

PR #274, opened 2014-12-15
Add getter method for worker to high level API. Initialization options for AsyncProxy: path and worker.

toberndo:
Currently if you initialize a worker with `openpgp.initWorker` you can't directly access the newly created AsyncProxy object. Access to methods like `AsyncProxy.prototype.decryptKeyPacket` therefore requires a second worker instantiation.

With this PR:

- the new method `openpgp.getWorker` returns the AsyncProxy if it was initialized with `openpgp.initWorker`
- `openpgp.initWorker` returns a Boolean which indicates successful Worker initialization
- `openpgp.initWorker`: the path parameter was replaced with an options parameter:
```
options.path relative path to the worker scripts, default: 'openpgp.worker.js'
options.worker initialized AsyncProxy object to be used as a worker
```
This breaks API compatibility, but with https://github.com/openpgpjs/openpgpjs/pull/261 on the way we will have here a third parameter therefore I think it makes sense to use options.
- removes the warning "You need to set the worker path", see also #269
- a helper method is added to get a Keyid object from a hex keyid

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -37,6 +37,7 @@ module.exports = {
   compression: enums.compression.zip,
   integrity_protect: true,
   rsa_blinding: true,
+  useWebCrypto: true,
 
   show_version: true,
   show_comment: true,
--- a/src/openpgp.js
+++ b/src/openpgp.js
@@ -46,14 +46,29 @@ if (typeof Promise === 'undefined') {
   require('es6-promise').polyfill();
 }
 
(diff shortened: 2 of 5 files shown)


toberndo:
Discussed with @tanx and https://github.com/mailvelope/openpgpjs/commit/712f807e503ffd1aa0a22618824bec5bfd92a84a changes this PR for API compatibility.

PR #281, opened 2015-01-14
Add option for encryption iteration/octet count

geraintluff:
I wanted to be able to alter the iteration count when encrypting the private-key, so I added an option for it.

Existing tests still pass - I wasn't sure where to put a new test or what it should be, though.

### DIFF
--- a/src/packet/secret_key.js
+++ b/src/packet/secret_key.js
@@ -184,16 +184,16 @@ SecretKey.prototype.encrypt = function (passphrase, options) {
   }
 
   var s2k = new type_s2k(),
-    symmetric = 'aes256',
-    cleartext = write_cleartext_mpi('sha1', this.algorithm, this.mpi),
-    key = produceEncryptionKey(s2k, passphrase, symmetric),
-    blockLen = crypto.cipher[symmetric].blockSize,
-    iv = crypto.random.getRandomBytes(blockLen);
-
   options = options || {};
   if (options.count) {
     s2k.set_count(options.count);
   }
+  
+  var symmetric = 'aes256';
+    cleartext = write_cleartext_mpi('sha1', this.algorithm, this.mpi),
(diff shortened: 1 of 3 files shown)


src/type/s2k.js
```diff
@@ -57,6 +57,23 @@ S2K.prototype.get_count = function () {
...
+  while (upperBits < 15 && count > (31 << (upperBits + expbias))) {
+    upperBits++;
+  }
+  
+  var lowerBits = 0;
+  while (lowerBits < 15 && count > (16 + lowerBits) << (upperBits + expbias)) {
+    lowerBits++;
+  }
+
+  this.c = lowerBits + (upperBits << 4);
```
tanx: I honestly cannot judge with the naked eye if this is correct or not. Perhaps add some unit test with several test vectors and please document your reasoning behind the code.

src/type/s2k.js
```diff
@@ -57,6 +57,23 @@ S2K.prototype.get_count = function () {
...
+  return this.get_count();
```
tanx: No need to return the getter in a setter function.

src/packet/secret_key.js
```diff
@@ -175,7 +175,7 @@ SecretKey.prototype.write = function () {
  * This can be used to remove passphrase protection after calling decrypt().
  * @param {String} passphrase
  */
-SecretKey.prototype.encrypt = function (passphrase) {
+SecretKey.prototype.encrypt = function (passphrase, options) {
```
tanx: just using a second `count` parameter here is fine. No need to use an options object.

tanx:
Thanks for the PR. I made some comments. Please rebase to the current master and push your branch against to initiate a new travis build. Thanks.

bartbutler:
Closing due to inactivity.

PR #287, opened 2015-02-09
Set preferred digest algos on generated keys in descending order of strength

anm:


tanx:
SHA-256 is currently considered secure. And our SHA-256 implementation is much faster than the others. Closing this in favor of performance.

PR #292, opened 2015-02-14
Specify a filename in encrypt fromBinary [Closes #284]

### LINKED ISSUE #284: Ability to encrypt file content/message with a specified filename
While [Pull Request 176](https://github.com/openpgpjs/openpgpjs/pull/176) added the ability to specify a Filename for the literal data packets and there was mention of adding the ability to specify it in the higher level api at the message level, at this point I am not aware of a way to use that feature easily without modifying the library or copying the processes in the existing encryptMessage function.  

Would it be possible to add an optional filename parameter to the openpgp.encryptMessage function or were there security reasons for not exposing the filename at these higher interface layers?

I've modified my own repository to allow me to set the filename easily, however I want to make it easy to keep my implementation up to date with any updates that openpgp.js may have going forward.

Thanks in Advance!

mapmeld:
- Follows the original suggestion in #176 to add a filename option to openpgp.message.fromBinary
- Adds a filename to the performance test to make sure this doesn't cause significant problems

I would be open to adding another test to verify the output file has the right filename, but I'm not sure how it'd work

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -327,11 +327,15 @@ function fromText(text) {
 /**
  * creates new message object from binary data
  * @param {String} bytes
+ * @param {String} filename
  * @return {module:message~Message} new message object
  * @static
  */
-function fromBinary(bytes) {
+function fromBinary(bytes, filename) {
   var literalDataPacket = new packet.Literal();
+  if (filename) {
+    literalDataPacket.setFilename(filename);
+  }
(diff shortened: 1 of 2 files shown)


evilaliv3:
:+1:

it wont' be bad to add the test for this but anyhow the patch seems fine to me.

PR #298, opened 2015-03-03
Check validity of key packet before signature verification

toberndo:
OpenPGP.js does not validate sub keys before using them for signature verification.

Attack scenario:
Alice uses insecure channel to send public key to Bob. MITM attacker adds subkey to Alice's key. Bob verifies fingerprint of Alice's key but not of the subkeys. Attacker can send messages to Bob signed with the fake subkey, pretending to come from Alice.

With this fix primary and subkey packets are validated before used in signature verification.

Bug reported by: @cure53


tanx:
Thanks @toberndo and @cure53

Released as [v0.11.1](https://github.com/openpgpjs/openpgpjs/releases/tag/v0.11.1)

PR #299, opened 2015-03-05
Escape email before using as regular expression in emailCheck method.

toberndo:
Email addresses are currently not escaped before using in a RegEx and therefore the check fails as e.g. '.' matches any single character.

### DIFF
--- a/src/keyring/keyring.js
+++ b/src/keyring/keyring.js
@@ -123,7 +123,9 @@ KeyArray.prototype.getForAddress = function(email) {
  * @return {Boolean} True if the email address is defined in the specified key
  */
 function emailCheck(email, key) {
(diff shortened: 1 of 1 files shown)


tanx:
Looks good. Let me know when you are done patching stuff from this week's audit. Then I'll just release all the changes in v1.0.0.

tanx:
Wait a second. Are you sure this works? It may be good to add some unit tests.

toberndo:
The escaping is from: https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Regular_Expressions

```
function escapeRegExp(string){
  return string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
}
```

PR #296, opened 2015-02-28
Support for crypto API keygen in IE11 on Windows 7 and 8

bartbutler:

PR #302, opened 2015-03-16
Password-protected session key packets

bartbutler:
1. Fixed a bunch of bugs with symmetrically-encrypted session key class.
2. Changed encryption method so RFC4880-compliant,
3. Added an optional 'passwords' argument to encryptMessage and decryptMessage to allow encryption with both public key packets and passwords for the same session key. 
4. Added tests to verify encryption/decryption with passwords, and also tested interoperability with GnuPG.

Public keys now optional if passwords present. Function signatures unchanged for now ('passwords' is just an additional parameter) though the more logical arrangement would be for the data to be the first argument.
Includes previous PR regarding IE11 WebCrypto support for key generation.

### DIFF
--- a/src/crypto/cfb.js
+++ b/src/crypto/cfb.js
@@ -247,7 +247,12 @@ module.exports = {
     var pos = 0;
     var cyphertext = '';
     var tempBlock = '';
-    blockc = iv.substring(0, block_size);
+    if (iv === null)
+      for (i = 0; i < block_size; i++) {
+        blockc += String.fromCharCode(0);
+      }
+    else
+      blockc = iv.substring(0, block_size);
     while (plaintext.length > block_size * pos) {
       var encblock = cipherfn.encrypt(util.str2bin(blockc));
       blocki = plaintext.substring((pos * block_size), (pos * block_size) + block_size);
--- a/src/message.js
+++ b/src/message.js
@@ -85,31 +85,54 @@ Message.prototype.getSigningKeyIds = function() {
 
 /**
  * Decrypt the message
- * @param {module:key~Key} privateKey private key with decrypted secret data
+ * @param {module:key~Key|String} privateKey private key with decrypted secret data or password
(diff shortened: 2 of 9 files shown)


tanx:
@bartbutler I've merged https://github.com/openpgpjs/openpgpjs/pull/296 so you can now rebase this PR on the current master to remove redundant commits.

@toberndo have you looked at this? Would this implementation fit your use case for private key sync?

bartbutler:
Will do, thanks.

toberndo:
My use case is already implemented with: https://github.com/openpgpjs/openpgpjs/blob/master/src/message.js#L186
This PR adds another possibility to create encryted session key packets.

bartbutler:
The main thing this does that the other does not is allow the same message to be encrypted with arbitrary numbers of password and public key packets.

bartbutler:
Rebase complete.

bartbutler:
Included in typed arrays migration.

tanx:
> Included in typed arrays migration.

I guess it's fine to have two ways to create sym. encrypted session key packets. But since we're going to break the api with the new refactoring, we might want to think about removing [the old api](https://github.com/openpgpjs/openpgpjs/blob/master/src/message.js#L180-L232) as it seems @bartbutler's approach is more flexible and we should reduce redundancy wherever possible.

toberndo:
> we might want to think about removing the old api

Sounds good.

tanx:
I have reverted the addition of the symEncrypt/symDecryp apis from https://github.com/openpgpjs/openpgpjs/commit/4d6a34faa1f960bfef1f28173c8451e519e7580d in favor of the new apis: https://github.com/openpgpjs/openpgpjs/commit/c221ee079f5507e88b229e3fc8c30ace17fb4d50. I'll make a proposal for a new options based api considering @evilaliv3's comment from https://github.com/openpgpjs/openpgpjs/pull/334#commitcomment-12365702.

PR #309, opened 2015-04-21
Implement Symmetric-Key Message Encryption using a passphrase

tanx:
See https://tools.ietf.org/html/rfc4880#section-3.7.2.2

This new api can be used together with a high entropy backup code in our [private key sync proposal](https://github.com/whiteout-io/mail-html5/wiki/Secure-OpenPGP-Key-Pair-Synchronization-via-IMAP) to securely sync a user's private key between devices.

The standard ascii armored output will replace the custom packet structure in the current version of the key sync spec and allow the usage of existing ciphers already in OpenPGP implementations today. This should ease adoption for other mail user agents that are interested in supporting key sync. I tested this against GPG v2.0.27 and it was able to decrypt the generated format.

Note: the key sync RFC will recommend to use the AES-256 symmetric cipher as well as SHA-256 for generating the passphrase via S2K.

Thanks @koto and Werner Koch for your feedback!

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -178,6 +178,60 @@ Message.prototype.encrypt = function(keys) {
 };
 
 /**
+ * Encrypt the message symmetrically using a passphrase.
+ *   https://tools.ietf.org/html/rfc4880#section-3.7.2.2
+ * @param {String} passphrase
+ * @return {Array<module:message~Message>} new message with encrypted content
+ */
+Message.prototype.symEncrypt = function(passphrase) {
+  if (!passphrase) {
+    throw new Error('The passphrase cannot be empty!');
+  }
+
+  var algo = enums.read(enums.symmetric, config.encryption_cipher);
+  var packetlist = new packet.List();
+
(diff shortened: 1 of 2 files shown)


src/message.js
```diff
@@ -178,6 +178,60 @@ Message.prototype.encrypt = function(keys) {
...
+    throw new Error('The passphrase cannot be empty!');
+  }
+
+  var algo = 'aes256';
+  var packetlist = new packet.List();
+
+  // create a Symmetric-key Encrypted Session Key (ESK)
+  var symESKPacket = new packet.SymEncryptedSessionKey();
+  symESKPacket.sessionKeyAlgorithm = algo;
+  symESKPacket.decrypt(passphrase);
```
tanx: The `decrypt` invocation is a little confusing here, but correct using of the current `SymEncryptedSessionKey` api. See https://github.com/openpgpjs/openpgpjs/blob/master/src/packet/sym_encrypted_session_key.js#L108-L117

toberndo:
Looks good. Merging.

Issue #314, opened 2015-05-05
Put unit testing under code coverage monitoring with Coveralls


evilaliv3:
Monitoring of code coverage is a valuable indicator of unit testing maturity and It would be relevant to put unit-tests for OpenPGP.js under coverage monitoring by means of [Coveralls](https://coveralls.io)

given the fact that tests are implemeneted by means of Mocha relevant tools are [coveralls](https://github.com/cainus/node-coveralls) combined with [grunt-mocha-istanbul](https://github.com/pocesar/grunt-mocha-istanbul)

evilaliv3:
I've implemented what is was suggesting; now code coverage is under monitoring and shows that currently only 79% of the library is covered; Coveralls.io offers a useful interface to review what are the crytical parts no covered in order to evaluate what tests write in order to rise up the coverage fastly.

![screenshot from 2015-05-06 01 15 13](https://cloud.githubusercontent.com/assets/217034/7484896/ac618a5e-f38e-11e4-80ee-a6b111f0e5b7.png)
![screenshot from 2015-05-06 01 15 24](https://cloud.githubusercontent.com/assets/217034/7484899/b404f264-f38e-11e4-9a8b-027fcfcef62f.png)
![screenshot from 2015-05-06 01 16 24](https://cloud.githubusercontent.com/assets/217034/7484902/bfebf5f0-f38e-11e4-9d39-f9cd5882e63f.png)

evilaliv3:
bam, implemented and integrated in master! enjoy!

https://coveralls.io/github/openpgpjs/openpgpjs

![screenshot from 2015-11-28 18 50 56](https://cloud.githubusercontent.com/assets/217034/11453453/77cab4a8-960b-11e5-83f7-01e5403ee694.png)

tanx:
Congrats 😊👍🏼

tanx:
I'm guessing only the travis tests are used as a hard quality gate for pull requests since not all sauce labs tests pass for all browsers?

That would be fine and definitely an improvement over the status quo with the results pic visibale in the github readme.

bartbutler:
Yes, fantastic I'll try to find some time soon to work on typed array stuff.

Issue #313, opened 2015-05-05
Implement cross-compatibility testing by means of SauceLabs


evilaliv3:
It would be relevant to implement some cross-browser compatibility testing in order to validate the implementation of OpenPGP.js among a wide range of Browsers, Versions and OS.

A complete in-the-cloud solution for this is offered by [https://saucelabs.com/](Saucelabs) that offers free accounts for OpenSource projects so that OpenPGP.js could benefit of it.

given the fact that openpgp.js makes us of Grunt, relevant library that would be interesting to integrate is [grunt-saucelabs](https://github.com/axemclion/grunt-saucelabs)

evilaliv3:
I've implemented what is was suggesting; now on every commit each test is executed on the relevant browsers (now selected randomly) and results are published through a badge inside the [Readme.md](https://github.com/evilaliv3/openpgpjs/tree/feature/unittesting)

![screenshot from 2015-05-06 01 32 54](https://cloud.githubusercontent.com/assets/217034/7484993/11fe3898-f390-11e4-99cf-d54053341a9e.png)
![screenshot from 2015-05-06 01 33 09](https://cloud.githubusercontent.com/assets/217034/7484995/1751ffc8-f390-11e4-97b6-c7dbdd940112.png)

AlexanderWillner:
very nice!

evilaliv3:
![screenshot from 2015-11-28 18 50 56](https://cloud.githubusercontent.com/assets/217034/11453441/12cd31ac-960b-11e5-829b-b38b0f9c7421.png)

bam, integrate in master! enjoy!

PR #315, opened 2015-05-05
Improve unit testing by means of Coveralls and Saucelabs

evilaliv3:
This pull request implement support for issues #313 and #314

- #313: Each unit test is now performed on SauceLabs.com in order to do on-commit cross-browser testing.
- #314 Code coverage is now monitored in the cloud on-commit by means of Coveralls.io;

An execution of the current .travis.ci script is available at: https://travis-ci.org/evilaliv3/openpgpjs


tanx:
Very cool. Thanks!

Any idea why the travis build fails?

evilaliv3:
let me check

evilaliv3:
sure tanx!

grunt saucelabs return false if there is at least one browser that is failing the test.
in the short term we can put "grunt saucelabs || true"

the real good solution would be to monitor only the  browsers/version/os for openpgp.js and put only them under monitoring.
e.g. using caniuse.com/#search=crypto

screenshot from 2015-05-11 11 40 15

as you can see:
- i did put ie8, ie9, ie10 just for test but we already know that webcrypto is not supported here.
- safari is failing and we should see why: https://saucelabs.com/tests/95a439f819a348cd88079077c7e7dad7 and eventually opening a issue and fix it
- iphone is failing
- i didnt put any ipad (saucelabs offers also this)
- i did not put any android (it would be really valuable to test against it)

here is an omnicomprensive list of the tests supporde by saucelabs, i'm going to investigate the relevant ones for openpgp.js an i will be back here: https://github.com/globaleaks/GlobaLeaks/blob/536435c60ad4ccb47d60094a556852ed9c9a499f/scripts/travis_saucelabs.sh

tanx:
iOS and Safari are failing because the native RSA keygen via WebCrypto will not accept a keysize smaller than 2048. If you set the keysize to 2048 in the test it should work. Although we might have to increase the timeout for that test:

https://github.com/openpgpjs/openpgpjs/blob/master/test/general/basic.js#L12

Also, is there any way you can give me access to the saucelabs instance so I can maintain changes in the future?

tanx:
And feel free to remove IE version older than 11.

evilaliv3:
right suggestions, i'm working on them and i will be back when all is ready for the clean integration.

evilaliv3:
sure tankred, when finished setting it up allo i'll pass the saucelab credentials.

evilaliv3:
ok @tanx  we are all set:

https://travis-ci.org/evilaliv3/openpgpjs
https://saucelabs.com/u/openpgpjs

in the end i've decided to not put safari under unit testing for the moment.
the reason is that as you said the key bit len should be raised to 2048.
this makes in general all unit tests to last longer, and in the specific cases of IE and Safari it makes saucelabs provide errors due to very big timeouts.

the latest selection is so the follwing:
tested browsers: IE, Chrome, Firefox
the stable version is tested on all OS
all the other relevant versions are tested only in the main native OS

screenshot from 2015-05-12 13 14 13

tanx:
> in the end i've decided to not put safari under unit testing for the moment.
> the reason is that as you said the key bit len should be raised to 2048.
> this makes in general all unit tests to last longer, and in the specific cases of IE and Safari it makes saucelabs provide errors due to very big timeouts.

I've enabled 2048 bit keys only in case of WebCrypto support here: https://github.com/openpgpjs/openpgpjs/commit/85133393e3f79c21bd51a2e1a72730a9138ce2de

This will not slow down browsers without webcrypto support like IE. If you rebase your PR on top of the current master you should be able to add iOS, Android and Safari to the test suite once again.

tanx:
Any update on this? Looking forward to cross-browser CI. Thanks again for your work.

evilaliv3:
i think we are done https://saucelabs.com/u/openpgpjs
https://github.com/evilaliv3/openpgpjs/tree/feature/unittesting

i've now commited the addition of all the browsers that support WebCrypto and that are on Saucelabs; let's see the results.

given the fact that we can add to the test whatever configuration (due to time limits) what are the rilevant browsers/version that we want to test among the one that does not support WebCrypto?

evilaliv3:
screenshot from 2015-05-21 14 38 09

Ok @tanx i think we are done and this is the status:
- tests succeed on every browser version that support webcrypto (and that is present on saucelabs) for: firefox, chrome, ie
- tests fails due to browser exploding on Android 5.1 (the one with chrome 42) runned on the linux emulator in one of the first tests (screencas available here: https://saucelabs.com/tests/a68371abf5e24fc7a41c713d093df537)
- tests fails on all OSX/IOS due probably to a current outage of saucelabs.

i've put all this tests in .travis.yml with a || true, this way we can track them without not impacting on the global badge failure/success.

fine to proceed for the merging for me.

\cc @fpietrosanti

evilaliv3:
@tanx @AlexanderWillner: except from the browsers that support webcrypto on wich other is OpenPGP expected to work? 

because  probably we should write the tests so that we test also the "config.useWebCrypto == false" condition so to test also what happens (on all the browsers) whenever the web crypto is not enabled or not present. if you can go over mergin this patch i promise to tackle this task! as it's also needed from us in globaleaks so i'm really intested to work on this topic for openpgpjs

\cc @fpietrosanti

fpietrosanti:
@tanx Do you have a list of OpenPGP.js practically supported with/webcrypto browsers/osx pair based on your experience? So that we can declare officially a supported listed list and invest time required to make all tests pass on those platforms where we know that OpenPGP.js works but the tests, for some reasons, need to re-organized

tanx:
Whiteout runs on Chrome, Firefox, Safari 8+, IE11, Android 4.0+ (Crosswalk WebView) and iOS 8.0+. We've got a pre-alpha Windows 8 and Firefox OS App running in our lab.

evilaliv3:
this cause it probably enforces the use of webCrypto.

by using "config.useWebCrypto == false"  is it still compatible with some browser? if yes what are them?

tanx:
> this cause it probably enforces the use of webCrypto.
> 
> by using "config.useWebCrypto == false" is it still compatible with some browser? if yes what are them?

Sorry I don't quite understand. Do you want to run the tests once with webcrypto activated and once without? Yes, you could do that by setting `openpgp.config.useWebCrypto = false`.

The only hard dependency for OpenPGP.js is `window.crypto.getRandomValues()`.

tanx:
> i've put all this tests in .travis.yml with a || true, this way we can track them without not impacting on the global badge failure/success.

GitHub tells me I can't merge automatically. Could you rebase on top of the current master?

evilaliv3:
rebased but concerning your comment:
> Sorry I don't quite understand. Do you want to run the tests once with webcrypto activated and once without? Yes, you could do that by setting openpgp.config.useWebCrypto = false.

> The only hard dependency for OpenPGP.js is window.crypto.getRandomValues().

let's wait for today and i will make a patch that will make us possible to test openpgp even with only this requirement: window.crypto.getRandomValues()

evilaliv3:
ah ok, no changes are needed in the unit tests. already openpgp does backword compatibility atuomagically.

screenshot from 2015-05-21 19 57 22

evilaliv3:
here we are @tanx !
no change is needed to the test, and thanks to saucelabs we spotted some failures on Safari, and Android

i do not know why github say that the build is failing: https://travis-ci.org/evilaliv3/openpgpjs/builds/63523435

ok for me to proceed with the merge.

tanx:
Don't know why the build is failing either but I'd like to wait with the merge until we have CI working again:
https://travis-ci.org/openpgpjs/openpgpjs/builds/63523438

N.B. I will be off the grid for the next two weeks. I'll look at this once I'm back or perhaps @toberndo can take a look in the meantime? thx

evilaliv3:
i've pusshed not a commit in order to trigger a need build.

evilaliv3:
ah! it is failing due to the fact that the keys for saucelabs where still encrypted for my repository. now it should work like a charm.

tanx:
Still seems to be failing: https://travis-ci.org/openpgpjs/openpgpjs/builds/63622271#L327

tanx:
FYI we finally have WebCrypto support for IE11 in https://github.com/openpgpjs/openpgpjs/pull/296 so the cross browser tests should really come in handy :+1:

evilaliv3:
i know @tanx let me redo a push just to trigger the build an other time. 
if it wont work it means that i should ask saucelabs to move the saucelabs account to openpgp/js repository (now credentials are binded to mine)

evilaliv3:
ok @tanx i've undestood why the build was failing.
travis does not support encrypted keys on pull request; now all unit tests should work.

when merging revert last commit and put the following inside .travis.yml:

    env:
      global:
        - secure: a3JyzGFVcoK6OqTf6g6Esf1OZlj7qhe9yv/dJPN9QEfHcEztvJAUHkpxvugsZPf2eTTZoFdD+stkLWvmAjPEqE9SLGrqAp9foQLF/MwvEd6gywO/c8Yx1rZ2JwQiTvFy9RKzKXHjx50h8fMTsrvyBhsd8WCkWXXn66FW+oZwnro=
        - secure: KMpqNgNDV9URoyxVNFHWgMPHB/P9Qb+dvX1ZVLh7J4LaVrLwkwBcwR94zkY9EnAtjK7gJd9lLDNOVhfWEayH6OGaxpu28vDeLLvsSLIw49bpYhdj6xVpheaaBv5MWz2itpkN+PmBAZwjgyuoLyriBNLgVjJaHVsh2vUj5C54D1Y=

the patch is already rebased for current master

https://saucelabs.com/u/openpgpjs

PR #324, opened 2015-06-09
Optimizations regarding string handling

plaincode:
Hi,

here some optimizations. They lead to a faster execution on Chrome and Safari (not on firefox however), and specifically improve the memory footprint when encrypting/decrypting PGP files on all browsers (machines), which is especially relevant on mobile devices.

Would love if you approve the changes.

Attached are some screenshots of a little test app to check encryption / decryption with openpgpjs found here:

http://www.lool.net/browser/ (uses openpgpjs with the changes)
http://www.lool.net/browser3/ (uses original openpgpjs)

The screenshots show encryption and decryption times of 10 MB random characters (lower numbers is the new version) on safari.

Best regards

Peter

10mb decrypt safari before
10mb decrypt safari after
10mb encrypt safari before
10mb encrypt safari after


tanx:
Thanks for the PR. Although I thought string concat is faster than join-with-push. See here:
http://stackoverflow.com/questions/7299010/why-is-string-concatenation-faster-than-array-join

plaincode:
Hi,

yes, for small strings this is fine, memory footprint will not be an issue and then it IS faster. 

However as soon as you are dealing with longer strings, memory consumption might grow very fast.

Let's you concatenate a char to a 20 MB string with the + operator, you will probably consume 40 MB temporarily e.g.:

// assume s is a 20MB string
var c = s + 'x';
// now at this point you WILL consume 40MB

Probably a JS engine could optimize it when there is no other reference to s and you do a:

s = s + 'x'

However in the current openpgpjs implementation it is an issue with the browsers we testet: Just watch browser memory consumption when encrypting/decrypting 20 MB or more.

Some more refactoring would be needed to further reduce consumption and really reference as much as needed.

Further work could include a streaming approach to be able to encrypt/decrypt PGP regardless of main memory (let's say a 1GB PGP even on a mobile device).

Thanks and best regards

Peter

fpietrosanti:
@plaincode We've been also doing work on "full streaming" by @evilaliv3 & @hellais for our use within GlobaLeaks project https://github.com/globaleaks/globaleaks that do require to handle the maximum level of uploading/downloading of files in streaming.

Our JS client need also to be able to download a file, decrypt it on the fly in streaming, re-encrypt it, upload it to another server (in order to integrate GlobaLeaks with 3rd party case management system), so every memory and performance optimization is very relevant to us.

plaincode:
A similar pull request is now sent from our 1&1 account, so i close this one.

PR #327, opened 2015-06-10

plaincode:


src/crypto/public_key/rsa.js
```diff
@@ -210,17 +199,6 @@ function RSA() {
...
-      return new Promise(function(resolve, reject) {
-        keyop.onerror = function (err) { 
-          reject(new Error(errmsg));
-        }
-        keyop.oncomplete = function (e) {
-          resolve(e.target.result);
-        }
-      });
-    }
-
```
tanx: Have your rebased on top of the current OpenPGP.js master? ;)

PR #328, opened 2015-06-10

plaincode:


src/util.js
```diff
@@ -202,11 +202,19 @@ module.exports = {
...
-    return result;
+    return result.join('');
+    // probably a faster alternative as follows
+    // however this function is obviously called with non UInt's at the end sometimes, so this won't pass the unit tests
+    //var CHUNK_SZ = 0x8000;
+    //var result = [];
+    //for (var i = 0; i < bin.length; i += CHUNK_SZ) {
+    //  result.push(String.fromCharCode.apply(null, bin.subarray(i, CHUNK_SZ)));
+    //}
+    //return result.join("");
```
tanx: Please remove this comment as well. Sorry I did not see this before.

tanx:
One more minor thing. FYI there is no need to create a new PR for every change if you don't work on the master but rather on a feature branch. Then you can always `git rebase -i <hash>` and `git push -f` to squash your changes into one commit.

plaincode:
Hi Tankred,

yes, well now i got some routine in the meantime :-) but next time i will definitely try it with a branch!

Ok, last issue regarding the comment is now fixed too.

Thanks

Peter

PR #330, opened 2015-06-12
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "openpgp",
-  "version": "1.1.0",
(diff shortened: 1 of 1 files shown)


tanx:
Well the version is set automatically in a grunt job when running the release.sh script. So the bower version is not maintained manually and will be set during release automatically. This PR won't change that. We could remove it from the grunt job, but I wouldn't loose sleep over it ;)

PR #336, opened 2015-06-29
Support multiple user IDs for key generation

toberndo:
- userID parameter of openpgp.generateKeyPair now supports type array
- first entry in array set as primary user ID


evilaliv3:
:+1:

PR #337, opened 2015-06-30
Support string message in verifyClearSignedMessage

adjohnson916:
This updates `verifyClearSignedMessage` to also accept cleartext messages as armored string, making the method symmetric with `signClearMessage` which already returns an armored string. I think this a more typical use case for consumers of this library than having a wrapping cleartext message object already constructed, so this makes things easier.

Let me know if there is interest, then I can add some unit tests.


tanx:
We use the message object as input for decrypt and verify operations.

PR #340, opened 2015-07-28
Update README.md

stphivos:
Include additional examples for accessing and modifying keys in the keyring.


tanx:
Thanks but the READMe is just supposed to give a very broad overview of the api. All the details are contained in the documentation and unit tests.

PR #344, opened 2015-08-11
If you attempt to store an empty keylist then delete the storage key

Gaff:
I guess this is a minor point, but if you remove all the keys from a keyring and then save it to storage, the storage key remains but with a value of "[]". This small fix deletes the key itself so that there's no empty keys left behind.

Hope this is ok.


evilaliv3:
good correction and thanks for adding the unit tests

PR #347, opened 2015-08-19
Fix running in Node.js (and io.js) with enforced strict mode

arzeth:
( ```$ node --use_strict``` )


evilaliv3:
good patch @arzeth!

PR #348, opened 2015-08-20
Cosmetic changes

arzeth:


evilaliv3:
:+1:

Mithgol:
> By the way, if a bit of performance (especially RAM usage) is preferred over
unification, then, obviously, conditions like
> ```js
> if (/SIGNED MESSAGE/.test(header[1])) {
> ```
> can be changed to
> ```js
> if (header[1].indexOf('SIGNED MESSAGE') !== -1) {
> ```

They can also be further changed to use [`String.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) from ECMAScript 6. Example:

```js
if (header[1].includes('SIGNED MESSAGE')) {
```

(But only after `openpgpjs` starts using ECMAScript 6.)

evilaliv3:
Thanks @Mithgol 

let's keep this in mind in relation to https://github.com/openpgpjs/openpgpjs/issues/385 and with the ES6 pull request by https://github.com/openpgpjs/openpgpjs/pull/395

PR #352, opened 2015-09-08
Keyring: support plain email addresses (without name) as user IDs.

toberndo:
A user ID which is not in the form `Name <abc@def.gh>` but instead only consists of an email address is currently not recognized in the `getForAddress()` methods of the keyring.
This PR fixes the email check function.


evilaliv3:
:+1:

PR #355, opened 2015-09-24
Fix: importScript is not defined in browser

lokeshjainRL:


evilaliv3:
@lokeshjainRL thanks for the patch, but can you please elaborate why this patch would be needed and what is the use case?

let me clarify my doubts:
the importScripts function is function supported by quite all browsers/versions (https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts) expecially the one targetted by the library.

in case it is not present simply avoid to use it will anyhow make the code fail as we will not able to import the needed libraries and all the next code will fail, so eventually instead of avoiding silently to import one should log an exception or something; anyhow as explained above i would not find this really useful.

let me explain. the function "importScripts" is used

bartbutler:
Your problem is the importScripts is only defined within a web worker. You must be using the web worker script in the main thread, which is not a bug, it's user error. Or you are using a browser that doesn't support web workers, in which case you should turn off the web worker feature, not hide the error message. Closing this and not merging it.

PR #357, opened 2015-09-28
Safe guard

lokeshjainRL:


bartbutler:
Isn't this the same as the other importScripts PR? As explained earlier, this is user error, not a bug.

tanx:
Gotta agree with @bartbutler on this one. Closing...

PR #360, opened 2015-10-07
Add option for deprecation of Symmetrically Encrypted Data Packet (Tag 9)

toberndo:
There is a recent analysis of chosen-ciphertext attacks to downgrade a Sym. Encrypted Integrity Protected Data Packet (Tag 18) packet to a plain Symmetrically Encrypted Data Packet (Tag 9).

The author Jonas Magazinius points out: "The implications are among others, that an encrypted and signed message can be stripped of its signature and modified arbitrarily, with certain restrictions, by an attacker without knowing the key."

Full details here: http://www.metzdowd.com/pipermail/cryptography/2015-October/026685.html

This seems to be known issue for years. GPG triggers a warning when a tag 9 packet (without integrity protection) is used. But this requires the user to understand the implications. And from a practical perspective tag 9 is never used as all relevant PGP implementations create messages only with integrity protection (tag 18). Then why keep this hole open?

I propose to deprecate tag 9. With this PR decryption of a message with tag 9 will throw an exception.

For legacy use cases there is the config.enforce_integrity_protection which can be set to false and bring back tag 9.

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -35,7 +35,10 @@ module.exports = {
   prefer_hash_algorithm: enums.hash.sha256,
   encryption_cipher: enums.symmetric.aes256,
   compression: enums.compression.zip,
+  // use integrity protection for symmetric encryption
   integrity_protect: true,
+  // only allow integrity protected messages to decrypt
+  enforce_integrity_protection: true,
   rsa_blinding: true,
   useWebCrypto: true,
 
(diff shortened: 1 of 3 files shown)


toberndo:
If no feedback, I would merge this PR tomorrow.

cm1313:
I am getting this error Symmetrically Encrypted Data Packet tag 9 deactivated can anyone help? Idn what to do

toberndo:
Seems like completely dropping support for tag 9 raises too many compatibility issues. We could either restrict this fix to messages with Symmetric-Key Encrypted Session Key Packets. Or leave it completely to the application to handle the issue. Closing this PR for now.

PR #364, opened 2015-10-23
Off by one in jsbn methods

0ndorio:
Just played with your library and discovered an off by one in jsbn.js bnpFromInt() through some unit tests.

#### Problem
A *-1* as input parameter is not properly converted and stored as *BigInteger {0: 0, t: 0, s: -1}* but it should be *BigInteger {0: 67108863, t: 1, s: -1}* (always asuming BigInteger.DB == 26).

```javascript
var bigInt =  new BigInteger(null);
bigInt.fromInt(-1);
var byteArray = bigInt.toByteArray();       /// [-1] instead of the expected [255, 255, 255]
var binString = util.bin2str(byteArray);     /// '￿' instead of the expected 'ÿÿÿ'
```

#### Additions
This error also ocures in the original library published by Tom Wu and the related npm package by Andy Perlitch.

### DIFF
--- a/src/crypto/public_key/jsbn.js
+++ b/src/crypto/public_key/jsbn.js
@@ -171,7 +171,7 @@ function bnpFromInt(x) {
   this.t = 1;
   this.s = (x < 0) ? -1 : 0;
   if (x > 0) this[0] = x;
-  else if (x < -1) this[0] = x + this.DV;
+  else if (x <= -1) this[0] = x + this.DV;
   else this.t = 0;
 }
 
--- a/src/crypto/public_key/jsbn.js
+++ b/src/crypto/public_key/jsbn.js
@@ -423,7 +423,7 @@ function bnpSubTo(a, r) {
     c -= a.s;
   }
   r.s = (c < 0) ? -1 : 0;
-  if (c < -1) r[i++] = this.DV + c;
+  if (c <= -1) r[i++] = this.DV + c;
   else if (c > 0) r[i++] = c;
   r.t = i;
   r.clamp();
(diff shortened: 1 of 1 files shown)


Rhynodeeds:
Lol I'm unfamiliar with how the Site or the system works I do have a proton
mail account I was trying to download their mobile app for Android I am
brushed up and pretty sound at some of the more technical aspects of
Android A.K.A.  --Linux platform ?? I have also added some phones mostly
older ones come with the tile Rood some old Samsung's that I had to plug
into the computer I have an HTC One m8 that I'm dying to root it just a
little out of my comfort level or skill set if you could help me in point
me in the right direction as well as helping at this app I would be very
grateful I'm guessing you are more than well rounded and Most areas of this
type of thing thank you for your time

0ndorio:
It seems like their is an conceptual bug also in bnpClamp() but i'm not totally sure about this:

```javascript
function bnpClamp() {
  var c = this.s & this.DM;
  while (this.t > 0 && this[this.t - 1] == c)--this.t;
}
```

If the integers value is negativ (represented by *this.s =  -1*) the first row will evaluate to *c = 67108863*.
This is exactly the value of an integer with value *-1*. So an bnpClamp() on an integer with value *-1* (represented by *BigInteger {0: 67108863, t: 1, s: -1}*) is always reduced to *BigInteger {0: 0, t: 0, s: -1}* which should be -0 and not -1 as far as i understand.

I would suggest to change the second row to:

```javascript
  while (this.t > 1 && this[this.t - 1] == c)--this.t;
```

PR #365, opened 2015-10-28
Ignore mdc error flag

toberndo:
This is a refinement of https://github.com/openpgpjs/openpgpjs/pull/360

Instead of completely deprecating tag 9, MDC is by default only required for modern ciphers (AES, except for Twofish).

This is aligned with the `ignore_mdc_error` flag in GPG: http://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=commit;h=625e292108cc0fd9077769587a8c22abe7805e33
Starting with Release 2.1.9 GPG will:
```
* gpg: Fail with an error instead of a warning if a modern cipher
   algorithm is used without a MDC.
```
From: https://lists.gnupg.org/pipermail/gnupg-announce/2015q4/000380.html

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -35,7 +35,10 @@ module.exports = {
   prefer_hash_algorithm: enums.hash.sha256,
   encryption_cipher: enums.symmetric.aes256,
   compression: enums.compression.zip,
+  // use integrity protection for symmetric encryption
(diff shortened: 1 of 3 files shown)


evilaliv3:
:+1:

PR #368, opened 2015-11-12
Fix for bad instance name in browsers

niieani:
I just spent 6 hours scratching my head, analyzing why openpgp wouldn't work in my setup. 
It turns out it has something to do either with the previous polyfill of Promise in `core.js` or the way `System.js` resolves dependencies - either way I don't have the will to test which anymore. 
What would happen is when I started generating a key pair, the promise would be dispatched and then... nothing. No error, no key pair. Anyway, after stepping through the Promise generation I noticed that the culprit is the `if (!(keys instanceof Promise))` which was evaluating incorrectly. For some reason the `typeof keys` was returning `object` instead of `Promise`. After some googling around it seems this is also required for Safari compatibility.
Following this [thread](http://stackoverflow.com/questions/27746304/how-do-i-tell-if-an-object-is-a-promise), I've made the necessary corrections (only two lines). The code executes correctly now. I believe it has no security implications so it should be safe to pull.

### DIFF
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -167,7 +167,7 @@ function RSA() {
         };
         
         keys = webCrypto.generateKey(keyGenOpt, true, ['sign', 'verify']);
-        if (!(keys instanceof Promise)) { // IE11 KeyOperation
+        if ('function' !== typeof keys.then) { // IE11 KeyOperation
           keys = convertKeyOperation(keys, 'Error generating RSA key pair.');
         }
       }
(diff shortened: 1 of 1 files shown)


evilaliv3:
@toberndo @tanx i think this patch is right.

the suggested generally check to do to see if an object is a promis is if it has a member called 'then' that is a function.
depending infact on the promise implementation of the browser instanceof Promise will result false.

thanks @niieani

evilaliv3:
@niieani : the issue reported is right but yourr commit was buggy.
i had to remove it and implement a slightly different.

the original check was:
```
if (!(keys instanceof Promise))
```

your correction was:
```
if (keys.then !== typeof 'function')
```

but the correct change is:
```
if (!(keys.then === typeof 'function'))
```

take care of booleans! :P

evilaliv3:
the fixed commit is here: https://github.com/openpgpjs/openpgpjs/commit/63c520b76596594e6babe69da275b516be355f49

niieani:
Oh, that's right! good spot and sorry for the mistake :)

PR #374, opened 2015-11-28
Add https://pgp.help to Readme

Gaff:
I'd humbly like to add my pgp.help webapp to this list of resources. It's a modern web-app for doing pgp. Similar to hanewin (etc) but more secure.


tanx:
How does this apply to https://github.com/openpgpjs/openpgpjs/pull/374/files#diff-04c6e90faac2675aa89e2176d2eec7d8R122

PR #375, opened 2015-11-29
Test against new node v4.2 LTS release

tanx:
See https://nodejs.org/en/blog/release/v4.2.0/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
   - "0.12.2"
+  - "4.2"
 before_script:
   - npm install -g grunt-cli
 script:


tanx:
The travis build for node v4.2 fails since the tests take more than 50 minutes to complete. The build for v0.12 also takes 42 minutes, which could lead to it failing as well if the travis machines are busy/slow.

Maybe we can optimize the unit tests a bit. Specifically generating a new keypair is probably not required in so many test cases. We could reuse one keypair for several tests.

toberndo:
Travis job currently runs on legacy infrastructure. https://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade might improve the build time. Switch done with by adding `sudo: false` to travis.yml.

Mithgol:
@toberndo That's right. See #376 for details.

evilaliv3:
it's right the travis infrastructure we are using is still the old one.

i was already evaluating to use the new one and caching some of the packets also. this would speed the build time.

in additional as i've implemented in globaleaks my idea is to split the browser testing among multiple runs and this would definetly reduce the time as travis does not limit the parallelization time but simply the total run; so that in 1hour limit of the total run one can get even 10000hours of parallel time.

i will do exactly the same that you see here: https://travis-ci.org/globaleaks/GlobaLeaks

evilaliv3:
merged manually, see: https://github.com/openpgpjs/openpgpjs/commit/46d2b6e66da9021e3e18d01a443aa3d4940ce402

tanx:
Awesome @evilaliv3 thanks!

tanx:
Travis is still saying that it's running on the old infrastructure. I guess you have to add 'sudo: false' to the travis.yml as well:

https://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

evilaliv3:
you are right man!

PR #377, opened 2015-12-01
Handle carriage-return-only newlines

hmarr:
Newlines are normalised to \r\n, but that previously assumed that only \r\n and \n newline characters could be present. Even though \r newlines are rarely used in the wild, it's be helpful to support them - currently, they're just removed entirely.

On related note - does anyone know *why* newlines are normalised before data is encrypted?

### DIFF
--- a/src/packet/literal.js
+++ b/src/packet/literal.js
@@ -48,7 +48,7 @@ export default function Literal() {
  */
 Literal.prototype.setText = function(text) {
   // normalize EOL to \r\n
-  text = text.replace(/\r/g, '').replace(/\n/g, '\r\n');
+  text = text.replace(/\r\n/g, '\n').replace(/\r/g, '\n').replace(/\n/g, '\r\n');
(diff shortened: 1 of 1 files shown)


tanx:
@hmarr thanks for the PR. Could you please rebase to the current master and `git push -f` to your `support-cr-eols` branch. I'd like to see if this breaks I current unit tests. Thanks.

hmarr:
@tanx done, tests are looking happy!

tanx:
@hmarr ok thanks. Btw here's a fun read on the difference between `\r\n` `\n ` and `\r `: 

http://stackoverflow.com/a/1279802

So with this patch I'm happy to finally announce support for Mac OS 9 in OpenPGP.js!

evilaliv3:
well done!

bartbutler:
@hmarr They're normalized because the RFC says they should be normalized.

PR #380, opened 2015-12-11
Add basic HKP lookup and upload support

tanx:
This adds basic HKP support by using the fetch api. Unlike promises the polyfill isn't included in the build though. They are provided in the unit tests as dev dependencies for node and browsers though. So users can see how to use them there. I added a comment to the readme about this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ matrix:
   include:
     - node_js: "0.12.2"
       env: OPENPGPJSTEST='unit'
-    - node_js: "4.2"
-      env: OPENPGPJSTEST='unit'
     - node_js: "0.12.2"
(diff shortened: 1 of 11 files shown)


tanx:
@toberndo please take a look and merge if it's ok. I will release v1.4.0 so we can use it in Mailvelope.

@bartbutler feel free to also take a look if the api fits your need for potential future use in Protonmail.

Thanks

toberndo:
Looks good to me.

fpietrosanti:
Which public keyserver do support already HKP for publishing and retrieval?

tanx:
All SKS server using v1.1.5+ should work out of the box. Just check the version string e.g.:

https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x360E65AE21F4E07C

PR #388, opened 2016-01-22
Build cleanup

tanx:
So I've found some more bugs in the current build process:

1. It's true that as of grunt-browserify 3.x you should use the `browserifyOptions` attribute as stated [here](https://github.com/jmreidy/grunt-browserify/blob/master/README.md#30-release), but that does not seem to apply to the `external` attribute. Which means we are currently building `'node-localstorage'` into our browser bundle. This PR fixes that issue

2. The package.json now points to our browserified build under `dist/openpgp.js` instead of `src/index.js`. This is a requirement for our the upcoming babelify build. Since we want to support older node.js versions that don't offer ES6 support yet. Point to the ES5 transpiled version with the `main` attribute solves this issue.

3. I removed the unused `npm_pack` grunt task.

@evilaliv3 please take a look and feel free to merge it if you're happy with the changes. Thanks!


evilaliv3:
okay @tanx i'm near to have found the bug, i'm just retesting ;)

Issue #387, opened 2016-01-21
Questions on current build process


tanx:
So I've started work on https://github.com/openpgpjs/openpgpjs/issues/385 on the [es6 branch](https://github.com/openpgpjs/openpgpjs/compare/es6) but unfortunately I'm having some questions about the current build.

* I get this error message when calling `grunt` after a clean checkout of the master branch:

````
Running "browserify:unittests" (browserify) task
>> Bundle test/openpgp.js created.
>> Error: Cannot find module 'openpgp' from '/Users/tanx/Projects/openpgpjs/test/general'
Warning: Error running grunt-browserify. Use --force to continue.
````

* Why are we building `test/openpgp.js` [here](
https://github.com/openpgpjs/openpgpjs/blob/79034f8ddeecf5d1ad4fe0193d22e2b95b5c879e/Gruntfile.js#L48)?

* Why are we copying the complete contents of `src/` to `test/` [here](https://github.com/openpgpjs/openpgpjs/blob/79034f8ddeecf5d1ad4fe0193d22e2b95b5c879e/Gruntfile.js#L154-L160)?

Thanks for clarifying!

evilaliv3:
hello there @tanx ! cool to see you are giving some love to openpgpjs :)

i'm not remembering exactly (and i will be back to you tonight with a precise answer ) but both the two things have been introduced in order to copy sources in the test directory in order to be able to instruct them with istambul and track code coverage.

evilaliv3:
ok just reviewed;

it seems that both are not needed anymore; i think that was something i needed in the first step when i was implementing this and then probably i forgotten to remove them.

i'm pushing a new branch called "coverage-retesting" where to see that all is able to work without issues removing both all the code you highligthed.

evilaliv3:
yep i confirm those lines are totally unneded: https://travis-ci.org/openpgpjs/openpgpjs/jobs/103898393

i'm merging their removal on the master branch @tanx; you can rebase your branch on it!

evilaliv3:
the merged patch is the following: https://github.com/openpgpjs/openpgpjs/commit/b92efd6087c7f5bff855a4d60e0f7df4b9620661

tanx:
Awesome! Thanks for the quick response. I rebased the es6 branch on top of the current master.

[...]

PR #389, opened 2016-01-22
Fix/387

evilaliv3:
This commit fixes issues and doubts as by ticket https://github.com/openpgpjs/openpgpjs/issues/387

The pull request deprecate https://github.com/openpgpjs/openpgpjs/pull/388 of which cherry-picks few commits.

@tanx: seems it is your turn my friend!

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -228,31 +228,12 @@ module.exports = function(grunt) {
   }
 
   grunt.registerTask('default', 'Build OpenPGP.js', function() {
-    grunt.task.run(['clean', 'copy:zlib', 'browserify', 'replace', 'uglify', 'npm_pack']);
+    grunt.task.run(['clean', 'copy:zlib', 'browserify', 'replace', 'uglify']);
     //TODO jshint is not run because of too many discovered issues, once these are addressed it should autorun
     grunt.log.ok('Before Submitting a Pull Request please also run `grunt jshint`.');
   });
 
   grunt.registerTask('documentation', ['jsdoc']);
-
-  // Alias the `npm_pack` task to run `npm pack`
-  grunt.registerTask('npm_pack', 'npm pack', function () {
-    var done = this.async();
-    var npm = require('child_process').exec('npm pack ../', { cwd: 'dist'}, function (err, stdout) {
-      var package = stdout;
-      if (err === null) {
-        var install = require('child_process').exec('npm install dist/' + package, function (err) {
-          done(err);
-        });
-        install.stdout.pipe(process.stdout);
-        install.stderr.pipe(process.stderr);
-      } else {
-        done(err);
-      }
-    });
-    npm.stdout.pipe(process.stdout);
-    npm.stderr.pipe(process.stderr);
-  });
 
   grunt.event.on('coverage', function(lcov, done){
     require('coveralls').handleInput(lcov, function(err){
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -16,9 +16,9 @@ module.exports = function(grunt) {
         },
         options: {
           browserifyOptions: {
-            standalone: 'openpgp',
-            external: [ 'crypto', 'node-localstorage' ]
-          }
+            standalone: 'openpgp'
+          },
+          external: [ 'crypto', 'node-localstorage' ]
         }
       },
       openpgp_debug: {
@@ -28,9 +28,9 @@ module.exports = function(grunt) {
         options: {
           browserifyOptions: {
             debug: true,
-            standalone: 'openpgp',
-            external: [ 'crypto', 'node-localstorage' ]
-          }
+            standalone: 'openpgp'
+          },
+          external: [ 'crypto', 'node-localstorage' ]
         }
       },
       worker: {
@@ -48,9 +48,7 @@ module.exports = function(grunt) {
           'test/lib/unittests-bundle.js': [ './test/unittests.js' ]
         },
         options: {
-          browserifyOptions: {
-            external: [ 'openpgp', 'crypto', 'node-localstorage']
-          }
+          external: [ 'openpgp']
         }
       }
     },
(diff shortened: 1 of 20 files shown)


evilaliv3:
@tanx but why you say it does not fix the issue?

doesnt this work for you?
```bash
git clone git@github.com:openpgpjs/openpgpjs.git
cd openpgpjs
git checkeout fix/387
npm install -d
grunt
grunt test
```

the same the tests on travis are working: https://travis-ci.org/openpgpjs/openpgpjs/builds/104202739

do not mind to the browsertesting that has the tunnels not working due of too much test runned, but in few minuts i will restart them.

tanx:
See my explanation [here](https://github.com/openpgpjs/openpgpjs/issues/387#issuecomment-174082166). I've been using this workaround with pointing directly to `../../dist/openpgp/` as well because I have not been able to identify why the `'openpgp'` module cannot be resolved locally anymore. Like I said in the linked explanation. I'm pretty sure the answer lies in the difference between the cached node_modules folder in travis and our local installatiom after a clean `npm install`.

Do you if there is anyway to download the contents of a travis cache?

evilaliv3:
not an official way, but eventually this one: https://github.com/evilaliv3/travis-ci-reverse-shell :)

Gruntfile.js
```diff
@@ -228,32 +226,13 @@ module.exports = function(grunt) {
...
 
   grunt.registerTask('documentation', ['jsdoc']);
 
-  // Alias the `npm_pack` task to run `npm pack`
-  grunt.registerTask('npm_pack', 'npm pack', function () {
-    var done = this.async();
-    var npm = require('child_process').exec('npm pack ../', { cwd: 'dist'}, function (err, stdout) {
-      var package = stdout;
-      if (err === null) {
-        var install = require('child_process').exec('npm install dist/' + package, function (err) {
```
tanx: Aha! It seems the travis cache just had an installation of openpgp from the previous build 😄

We probably had this in our local node_modules folder as well after the first successful grunt buil 😉

evilaliv3:
anyway i got all your points, and i'm sure you are right but it's also true that the npm_pack is hiding this possible bug of packaging since a lot.

to find the breaking point for that we shall do a bisect over the commits using this test case:
```
npm install -d
grunt
rm node_modules/openpgp
grunt test
```

an we shall find the in which we do not get an error

Gruntfile.js
```diff
@@ -48,9 +48,7 @@ module.exports = function(grunt) {
           'test/lib/unittests-bundle.js': [ './test/unittests.js' ]
         },
         options: {
-          browserifyOptions: {
-            external: [ 'openpgp', 'crypto', 'node-localstorage']
-          }
+          external: [ 'openpgp']
```
tanx: We should keep 'crypto' 'node-localstorage' external for the browser unit tests. Any reason you added them here?

evilaliv3:
@tanx: for a matter of curiosity i retested on a clean v1.2.0 removing the nodemodules/openpgp and i get the same issue there.

Gruntfile.js
evilaliv3: do you mean why i removed them here (you said added)?

i was about to tell you, i removed them as they are included as external by 'openpgp'; am i wrong?

Gruntfile.js
evilaliv3: exactly :P this is what i'm trying to repeat to you in all the varius pull :)

sorry for the confusion of today; i'm also tired but now i think you can understand what i was saying here and there

Gruntfile.js
evilaliv3: ah! now i understand why you are talking in reversed;
by the way the openpgpjs module automatically; probably yes, this line should be  "external: [ 'openpgp', 'crypto', 'node-localstorage']" but i let this to you as you have more experience in this.

tanx:
> @tanx: for a matter of curiosity i retested on a clean v1.2.0 removing the nodemodules/openpgp and i get the same issue there.

Yeah I think you're right. I guess it is not normal behavior for node to resolve 'openpgp' using the project's own package.json. I was only assuming this since there was a local installation hiding under node_modules/openpgp which resulted from the npm_pack grunt job.

Anyway it seems your fix seems to work. BUT we need to keep in mind that we need to test dist/openpgp.js instead of src/index.js because of the babelify build. So let me try to use that locally instead get back to you with a suggestion after testing that locally.

evilaliv3:
dist/openpgp.js is tested in all browsers tests: https://github.com/openpgpjs/openpgpjs/blob/master/test/unittests.html

instread for tracking code coverage we run tests on sources

this mixed approach shoould work i think

tanx:
> this mixed approach shoould work i think

Yes it will right now. But as soon as we have ES6 syntax in src/ it will break e.g. for node v0.12. See my PR here: https://github.com/openpgpjs/openpgpjs/pull/390

evilaliv3:
perfect! :) so looking forward for this mix:

ES6 + ArrayBuffers + streaming patch for encryption + updated selection of crypto libraries

i'm dreaming of this but in the short term i would need from you simple the review of streaming patch that is really selfcontained and unittested; aside from strong passion this is why i'm trying to put all my effort in support you when a bug pops out. :)

tanx:
> ES6 + ArrayBuffers + streaming patch for encryption + updated selection of crypto libraries
> 
> i'm dreaming of this but in the short term i would need from you simple the review of streaming patch that is really selfcontained and unittested; aside from strong passion this is why i'm trying to put all my effort in support you when a bug pops out. :)

I hear you! That's why I'm keeping the diff with ES6 on it's own branch for now. The use of `dist/openpgp` is just to stay future proof. Nothing more nothing less. Having said that, I will probably get back to the ES6 branch after reviewing/merging the ArrayBuffers and streaming patch for encryption PRs. I will probably look at ArrayBuffers first though.

evilaliv3:
perfect for me to proceed on the array buffer this way i wil support you on the retesting cause i'm also interested in the possible performance/memory improvement that may follow.

in the last days i did a little of review of it but aside from requiring to be cleaned form minor mistakes and rebased is a simple mapping one-to-one.

PR #390, opened 2016-01-23
Fix/387 dist

tanx:
Same as @evilaliv3's PR only with dist instead of src for node unit tests.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -48,7 +48,7 @@ module.exports = function(grunt) {
           'test/lib/unittests-bundle.js': [ './test/unittests.js' ]
         },
         options: {
-          external: [ 'openpgp']
+          external: ['crypto', 'node-localstorage', 'openpgp', '../../../dist/openpgp', '../../dist/openpgp']
         }
       }
     },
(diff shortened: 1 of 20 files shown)


evilaliv3:
@tanx: i just retested a cleak pull + build + grunt test on my pc and it works! ok for mergin this patch if it passes travis!

my only doubt on the question you did here, but i let it to you as i've not experience on this: https://github.com/openpgpjs/openpgpjs/pull/389#discussion-diff-50606116

Gruntfile.js
```diff
@@ -48,9 +48,7 @@ module.exports = function(grunt) {
           'test/lib/unittests-bundle.js': [ './test/unittests.js' ]
         },
         options: {
-          browserifyOptions: {
-            external: [ 'openpgp', 'crypto', 'node-localstorage']
-          }
+          external: ['crypto', 'node-localstorage', 'openpgp', '../../../dist/openpgp', '../../dist/openpgp']
```
tanx: @evilaliv3 to answer your question... this is how things should be excluded just to be sure the browserified unittest.js build does not contain any src files in any of the unit tests. E.g. `test/crypto/cipher/aes.js` is one level deeper than `test/crypto/crypto.js`.

evilaliv3: clear! thanks so much for the clarification

tanx:
Ok. The travis build works! Merging...

PR #393, opened 2016-01-24
Codeclimate

evilaliv3:
As done for [dchest/scrypt-async-js](https://github.com/dchest/scrypt-async-js/pull/21) i'm here proposing this improvement:

This pull requests:
- Add tracking of code quality with Codeclimate
- replace Coveralls with Codeclimate coverage

since the integration it will be possible to have the following outputs:
screenshot from 2016-01-24 20 26 18
screenshot from 2016-01-24 20 26 37
screenshot from 2016-01-24 20 27 59

it will be easy then fix issues in the code related to not standard/interoperable syntaxes

After integration the codeclimate token should be replaced with a token specific for the dchest/scrypt-async-js to be taken loggin onto https://codeclimate.com/dashboard

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 language: node_js
 addons:
     code_climate:
-        repo_token: tanxchangeme!46ecd492907f8224b0ae64890d4befb4bbd9ebbeba55b5acb38
+        repo_token: $CODECLIMATE_REPO_TOKEN
 matrix:
   fast_finish: true
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
 language: node_js
+addons:
+    code_climate:
+        repo_token: tanxchangeme!46ecd492907f8224b0ae64890d4befb4bbd9ebbeba55b5acb38
 matrix:
   fast_finish: true
   include:
@@ -52,6 +55,6 @@ matrix:
     - env: OPENPGPJSTEST='end2end-12' BROWSER='iphone 7.0'
     - env: OPENPGPJSTEST='end2end-13' BROWSER='iphone 9.1'
 before_script:
-  - npm install -g grunt-cli
+  - npm install -g grunt-cli codeclimate-test-reporter
 script:
   - $TRAVIS_BUILD_DIR/travis.sh
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -136,15 +136,6 @@ module.exports = function(grunt) {
           root: '.',
           timeout: 240000,
         }
-      },
-      coveralls: {
-        src: ['test'],
-        options: {
-          root: '.',
-          timeout: 240000,
-          coverage: true,
-          reportFormats: ['cobertura','lcovonly']
-        }
       }
     },
     mochaTest: {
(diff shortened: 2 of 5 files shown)


evilaliv3:
Here are visible the outputs that will be available after the integration of the patch:
https://codeclimate.com/github/openpgpjs/openpgpjs
https://codeclimate.com/github/openpgpjs/openpgpjs/src/crypto/public_key/jsbn.js

.travis.yml
```diff
@@ -1,5 +1,8 @@
 sudo: false
 language: node_js
+addons:
+    code_climate:
+        repo_token: tanxchangeme!46ecd492907f8224b0ae64890d4befb4bbd9ebbeba55b5acb38
```
tanx: Which token do we need here?

tanx:
Apart from my question regarding the repo_token, looks good from my side.

.travis.yml
evilaliv3: you need to access to code climate and as you are the owner you should read the token needed for openpgpjs/openpgpjs.

you find it under openpgpjs/openpgpjs section inside the code coverage subsection.

evilaliv3:
you need to access to code climate and as you are the owner you should read the token needed for openpgpjs/openpgpjs.

you find it under openpgpjs/openpgpjs section inside the code coverage subsection.

tanx:
Wow. Now I know what made our travis builds so slow for node.js v4 and v5 ... I disable coveralls since it broke the tests for the typed array branch: https://github.com/openpgpjs/openpgpjs/commit/4ba60956304a894df85d1c97ba2e4d9bd4e6beeb

https://travis-ci.org/openpgpjs/openpgpjs/builds/106153867

evilaliv3:
the code coverage surely make it more slow, but i think other issues causes that big latency. let's see now that you are testing it's removal.

ok remove it, i will try to fix it in a separate branch later on.

.travis.yml
```diff
@@ -1,5 +1,8 @@
 sudo: false
 language: node_js
+addons:
+    code_climate:
+        repo_token: $CODECLIMATE_REPO_TOKEN
```
tanx: I set the repo token in travis-ci using the `CODECLIMATE_REPO_TOKEN` env var. CodeClimate only checks the master branch though. So things will only work once the PR is merged.

tanx:
> the code coverage surely make it more slow, but i think other issues causes that big latency. let's see now that you are testing it's removal.

The travis build for node.js finish in about 2 minutes by running just `npm test`. See:
https://travis-ci.org/openpgpjs/openpgpjs/builds/106153867

> ok remove it, i will try to fix it in a separate branch later on.

I would prioritize fast and reliable build above displaying code coverage. That way we could add node v4.2 and node v5 to the required builds and get notified when those break. Displaying coverage is a nice plus though, so it would be great if we could have both ;)

.travis.yml
evilaliv3: where is this variable set?

codeclimate checks the code quality (with linters), but the code is instrumented and subject to code coverage in every repo, as we are calculating it in travis. then codeclimate will show the last one published.

what i suggest is then also to instrument the code and publish it only on the master branch; i will modify travis.sh accordingly if you are ok.

tanx: > where is this variable set?

https://travis-ci.org/openpgpjs/openpgpjs/settings

> codeclimate checks the code quality (with linters), but the code is instrumented and subject to code coverage in every repo, as we are calculating it in travis. then codeclimate will show the last one published.
> 
> what i suggest is then also to instrument the code and publish it only on the master branch; i will modify travis.sh accordingly if you are ok.

Sure. Do you want to merge this PR to master then? I will continue to work without coverage on the typed array branch and have set the node v4.2 and v5 travis jobs as required there.

evilaliv3: Sure. Im out now but i will see It later  on this morning

tanx: Ok. Just FYI I rebase to the HEAD of master, so do a pull if you intend to work in this branch locally.

evilaliv3:
:+1:

PR #394, opened 2016-01-24
Remove ripped and outdated jdoc template

evilaliv3:
This pull request removes the ripped and outdated jdoc template included in https://github.com/openpgpjs/openpgpjs/tree/master/jsdoc.template

There is no need to include it and it is also outdated; in addition it may lead to copyright infringement.


tanx:
Looks good to me.

PR #395, opened 2016-01-25
Enable ES6 support by using babelify transform in browserify build

tanx:
* Add babelify as dev dependency
* Add babelify transform to browserify build
* Enable es6 support for jshint
* Ignore minified js files in babel transform

**N.B. This PR just enables support for ES6 in our browserify build. It does not change any source code. It does however change some semantics e.g. all code is now automatically executed in ES5 strict mode as explained [here](http://blog.shinetech.com/2015/07/07/es6-with-babel-js/). Any ES6 module patches are currently maintained in the [es6-modules branch](https://github.com/openpgpjs/openpgpjs/compare/es6...es6-modules) until after we review/merge some other open PRs.**

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -16,6 +16,7 @@
 	"trailing": true,
 	"undef": true,
 	"unused": true,
+	"esnext": true,
 
 	"predef": [
 			"console",
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -18,7 +18,13 @@ module.exports = function(grunt) {
           browserifyOptions: {
             standalone: 'openpgp'
           },
-          external: [ 'crypto', 'node-localstorage' ]
+          external: [ 'crypto', 'node-localstorage' ],
+          transform: [
+            ["babelify", {
+              ignore: ['*.min.js'],
+              presets: ["es2015"]
+            }]
+          ]
         }
       },
       openpgp_debug: {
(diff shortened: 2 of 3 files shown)


tanx:
@evilaliv3 if you're ok with merging this now, it would allow me to use ES6 syntax while reviewing/cleaning up @bartbutler's TypedArray branch. Thanks!

evilaliv3:
The pull does what it says and works! mergin it;

@tanx the only consideration is that you added dependencies using ^ while until now we used mostly ~.
i think we should end in betterly understanding and documenting these decisions in order to try to do it in the proper way.

tanx:
^ updates minor versions: 1.4.0 -> 1.50
~ updates bugfixes: 1.4.0 -> 1.4.1

I used ^ here because babel seems to practice semantic versioning with breaking changes only in major updates. Plus it's also quite a mature project with lots of users. I'm less hesitant to use ^ in that case, since more people notice when something breaks v.s. smaller libs where I use ~ more often.

tanx:
@evilaliv3 it seems all the browser tests are failing.
https://saucelabs.com/u/openpgpjs

Is it because of this PR?

tanx:
So it seems that this commit does indeed break the browser tests:

Before: https://travis-ci.org/openpgpjs/openpgpjs/builds/104534357

After: https://travis-ci.org/openpgpjs/openpgpjs/builds/104605735

So perhaps we should revert it until we figure out why.

evilaliv3:
yes, it seem so.

in general tests reflect always the latest build;
to have them sable we should eventually run them only on master but this way they won't give use valuable input during development.

evilaliv3:
@tanx i'm wondering that this may be the cause?

https://github.com/openpgpjs/openpgpjs/blob/master/test/unittests.html#L13

evilaliv3:
ok just rechecked and it's nothing related to https://github.com/openpgpjs/openpgpjs/blob/master/test/unittests.html#L13 that is correct

PR #396, opened 2016-01-26
Revert "Enable ES6 support by using babelify transform in browserify build"

tanx:
Reverts openpgpjs/openpgpjs#395

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -16,7 +16,6 @@
 	"trailing": true,
 	"undef": true,
 	"unused": true,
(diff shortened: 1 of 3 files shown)


evilaliv3:
Ok let's wait for the tests to finish and then merge it.

tanx:
So this PR reverts the addition of babelify. Let's keep this on the es6-modules branch until that we understand why the browser tests are failing.

evilaliv3:
Okay, the revert appear to work.

PR #398, opened 2016-01-30
Promise polyfill

tanx:
This PR goes back to using v1.0.0 of the es6-promise polyfill since the newer version break the phantom.js tests on hoodiecrow. It also adds node v5 to the travis build as an optional build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ matrix:
       env: OPENPGPJSTEST='unit' OPENPGP_NODE_JS='0.12'
     - node_js: "4.2"
       env: OPENPGPJSTEST='unit' OPENPGP_NODE_JS='4.2'
+    - node_js: "5"
+      env: OPENPGPJSTEST='unit' OPENPGP_NODE_JS='5'
     - node_js: "0.12"
       env: OPENPGPJSTEST='end2end-0' BROWSER='firefox 38'
     - node_js: "0.12"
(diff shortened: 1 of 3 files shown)


evilaliv3:
Ok for this revert if this breaks hoodiecrow

Waiting for the test results and i will take care of merging this.
In the meantime do you have the error with phantomjs? this way we can open an issue on the es6-promise.

tanx:
> Ok for this revert if this breaks hoodiecrow
> 
> Waiting for the test results and i will take care of merging this.

Thanks! I know this is kind of an application layer issue, but since we ship the polyfill in the build, others might experience this issue in phantomjs as well.

> In the meantime do you have the error with phantomjs? this way we can open an issue on the es6-promise.

The problem is that newer versions of es6-promise simply don't seem to polyfill Promise.resolve() behavior correctly under Phantom.js so that my tests never finish. It's really wierd and I wish I could explain what the reason is. I'll open an issue on es6-promise once I find out what's going on.

In the meantime v1.0.0 has been working like a charm on all relevant platforms. So no reason to upgrade right now.

tanx:
> Waiting for the test results and i will take care of merging this.

Note that I haven't updated the shrinkwrap file in this PR. I'll do that in the upcoming v1.5.6 release after this PR is merged. In the meantime I've created a test build using an updated shrinkwrap here:

https://travis-ci.org/openpgpjs/openpgpjs/builds/105851609

tanx:
Travis build worked... merging and releasing v1.5.6

evilaliv3:
perfect!

Tests are passing except the same timeout on node 4 and 5 as usual; i would suggest to keep this timeout as memento mori without raising the threshold :)
this way we can use this as comparison in the uintarray branch.

safari is failing to connect; maybe we should try to switch simply the tcp port from the 3000 to the 9000 (that is the one i'm using in globaleaks)

the 3000 as listed here should work, but maybe they did some mistake in recent updates: https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+FAQS

PR #400, opened 2016-01-30
Add linting to grunt test job

tanx:
* Add grunt-jscs and .jscsrc (lint trailing whitespaces and indentation length)
* Cleanup .jshintrc (use globals instead of deprecated predefs)
* Add Gruntfile.js and *.json linting to npm test

N.B. This does not lint the whole src/ directory during grunt test but
this should definitely be our goal. Please try to de-lint any js files
you touch during development using a texteditor that support linting
(Sublime Text 3 and SublimeLinter 3 with jshint/jscs plugins work great)


evilaliv3:
Works for me. Let's see what travis will say.

PR #401, opened 2016-02-01
Add Chrome, Firefox and IE11 to required tests

tanx:
These run pretty reliably and fast. I would add them to required tests so that we get email notifications not just for node v12.2.


evilaliv3:
good to go! right!

i hope to be able to fix soon safari and understand what is not working in edge as well

Issue #399, opened 2016-01-30
Safari and iOS not working


tanx:
This is an observation I've made during testing... The travis builds always fail for Safari on Mac OS and iOS. I've been able to reproduce a failure locally for Safari 9.0.3 running on Mac OS 10.11.3 but haven't found the time to debug the underlying reason. Patches welcome from anyone who finds the time ;)

Thanks.

PR #404, opened 2016-02-03
Fix unit tests on Safari, Closes openpgpjs/openpgpjs#399

tanx:


evilaliv3:
great!maybe we would like to test 4096 on all browsers except on safari?

tanx:
> great!maybe we would like to test 4096 on all browsers except on safari?

Please not on travis. I'm glad they're running really fast now. 43 seconds for node v5 ;)

https://travis-ci.org/openpgpjs/openpgpjs/builds/106728025

evilaliv3:
okay ;)

tanx:
Beautiful. The current version of all platform are now green! iOS and Android included:

https://github.com/openpgpjs/openpgpjs

evilaliv3:
yes probably iOS was failing due of safary;
i dot know instead what fixed android.

PR #405, opened 2016-02-05
ES6 modules

### LINKED ISSUE #385: Use ES6 module syntax
Hey,

I've been experimenting with ES6 and babel in hoodiecrow and am really liking ES6 modules. The main advantage is that you don't need a build step at development time since the browser can resolve the modules natively. It also works well with browserify using the babelify so that you can just consume common.js modules like before using the new syntax.

May proposal would be to refactor the following:

    var lib = require('lib');

to

    import lib from 'lib';

and

    module.exports = lib;

to

    export default lib;

It's really easy to integrate babel into the browserify grunt build. See my setup here:

https://github.com/tanx/hoodiecrow/blob/master/Gruntfile.js#L15-L19

The cool thing is developers can choose to activate experimental ES6 features in chrome to work without a build step during development:

https://github.com/nodejs/node-v0.x-archive/wiki/ES6-%28a.k.a.-Harmony%29-Features-Implemented-in-V8-and-Available-in-Node

But this is purely optional and devs can keep using the current browserify build workflow as well if they wish.

Publishing to npm works like so:

http://mammal.io/articles/using-es6-today/

Let me know what you think. It's not a lot of work to refactor from common.js to es6 but I wanted to get you input before doing it.

Thanks

tanx:
This PR merges to the v2.x branch and resolves #385 by using ES6 modules in the `src/` tree. It seems cleaning up the code to use strict mode solved the build problems with babel.


VIPv6:
Why was ES6 abandoned?  I'm in need of OpenPGP signature verification in the browser.  If OpenPGPjs were in ES6 instead of commonJS, then the module could be used in either NodeJS or the browser.  ES6 modules seem to be the ideal universal module format.  Please reconsider ES6.

PR #406, opened 2016-02-05
v2.x merge to master

tanx:
This branch builds on @bartbutler's `typed_arrays_and_speed` branch but goes quite a bit beyond. The goal is to be the foundation of the upcoming OpenPGP.js v2.0.0 release.

Test it out with:

    npm install git://github.com/openpgpjs/openpgpjs.git#v2.0.0-dev

# v2.0.0 cliff notes
* Completely new public api (breaking change to v1.x). You can review it [here](https://github.com/openpgpjs/openpgpjs/blob/v2.x/src/openpgp.js) and see some usage examples in the [README]( https://github.com/openpgpjs/openpgpjs/blob/v2.x/README.md#examples).
* Refactor to ES6 modules instead of common.js
* Uint8Array all the things! (kudos go to @bartbutler for the initial refactoring).
* Native node.js crypto for AES-CFB and all hashing (SHA-1, SHA-256, SHA-384, SHA-512, MD5, RIPE).
* Use asm.js implementations of AES-CFB, SHA-1, SHA-256 when no native crypto is available. Uses [Rusha](https://github.com/srijs/rusha) and [asmCrypto Lite](https://www.npmjs.com/package/asmcrypto-lite) (a minimal subset of asmCrypto.js built specifically for OpenPGP.js).
* All code under `src/` has been cleaned up and now runs in strict mode (with a few exceptions). This prevents a large class of bugs and lets JS runtimes to optimize more easily. Grunt jobs for JShint and JSCS (code style linting) now run automatically before each travis build so that PRs from outside must adhere to strict mode and our code style e.g. indentations using 2 spaces.

# Future roadmap
* Native WebCrypto apis for SHA hashing. This would require refactoring of most private apis from sync to async/promise style.
* Use WebCrypto or at least asmCrypto for RSA en/decrypt. Replacing current jsbn/rsa implementation (big performance bottleneck for larger RSA keys).
* Add resync-mode support to asmCrypto AES-CFB for non-integrity-protected symmetric encrypted packets. Replacing current AES/CFB implementation (right have two atm).


evilaliv3:
@tanx let us know when you consider this pull requests finalized from your side an you would require us to start reviwing them.

tanx:
IMHO It's ready to merge. These changes have been disconnected from the master branch long enough. I agree that there is work to be done on cleaning up the apis. But that work can happen in a different PR and should be kept outside of the scope of this discussion. This PR does not break the v1.x api.

bartbutler:
Let's do it.

tanx:
> IMHO It's ready to merge.

Sorry I meant ready to review ;)

tanx:
Has anyone gotten a chance to reciew this yet? If not I am nearing completion of the `new_api` branch. Just need to fix the rest of the old tests.

I guess I will merge that to the v2.x branch and this PR will contain all changes. Reminds me of:

https://twitter.com/iamdevloper/status/397664295875805184

evilaliv3:
i was about to review something tonight. what do you suggest as order?

tanx:
> i was about to review something tonight. what do you suggest as order?

1. This PR (`v2.x` is rebased on top of `master`)
2. The `es6-modules` PR (is rebased on top of `v2.x`)
3. The upcoming `new_api` PR (is rebased on top of `es6-modules`)

You can also wait and review everything all at once. That might make more sense as well. Since the `v2.x` branch still contains a lot of bugs that are fixed in `new_api`. It will just be a bigger change set. But I've tried to keep all commits very modular so you could go through them one by one. So I guess I'll merge `new_api` to `v2.x` tomorrow and update this PR.

evilaliv3:
okay! so tomorrow night i will try to do my best in review it all.

2016-02-10 0:24 GMT+01:00 Tankred Hase <notifications@github.com>:

> i was about to review something tonight. what do you suggest as order?
>
>
>    1. This PR (v2.x is rebased on top of master)
>    2. The es6-modules PR (is rebased on top of v2.x)
>    3. The upcoming new_api PR (is rebased on top of es6-modules)
>
> You can also wait and review everything all at once. That might make more
> sense as well. Since the v2.x branch still contains a lot of bugs that
> are fixed in new_api. It will just be a bigger change set. But I've tried
> to keep all commits very modular so you could go through them one by one.
> So I guess I'll merge new_api to v2.x tomorrow and update this PR.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openpgpjs/openpgpjs/pull/406#issuecomment-182131348>.
>



-- 
Ing. Giovanni Pellerano - Founding Member and CEO
giovanni.pellerano@logioshermes.org | +39.328.9590046

HERMES - Center for Transparency and Digital Human Rights
Associazione No Profit | Via Aretusa 34, IT-20129 Milan, Italy
t. +39-02-87186005 (voicemail) | f. +39-02-87162573
TaxCode: IT-97621810155 | EuropeAid: IT-2012-AOD-0806909254
w. http://logioshermes.org | m. info@logioshermes.org

tanx:
@evilaliv3 @bartbutler this PR now contains all the changes  and its's ready for review. Thanks in advance for any feedback!

tanx:
@bartbutler @evilaliv3 @toberndo I created a test release for v2.0.0. You can install it with:

    npm install git://github.com/openpgpjs/openpgpjs.git#v2.0.0-dev

tanx:
See the updated README for examples of how to use the api:

https://github.com/openpgpjs/openpgpjs/blob/v2.x/README.md#examples

evilaliv3:
fine for me, i've not particular comments but i'm new to es6 to understand if all the changes are correct but it seems so :)

nice the adition of the zerocopy with transferable objects; what i do not like is that the feature is to be enabled manually by the user, and demand to the user to write the browserdetection logic and setting of openpgp.config.zeroCopy accordingly.
i would really like more if we could add simply a test for it and set the variable zeroCopy automagically:

http://www.hyperlounge.com/blog/up-the-workers-2/

```javascript
if (worker.webkitPostMessage) {
    transferableSupported = true;
    worker.postMessage = worker.webkitPostMessage;
} else {
    try {
        var ab = new ArrayBuffer(1);
        worker.postMessage(ab, [ab]);
        transferableSupported = ab.byteLength == 0;
    } catch(error) {
        transferableSupported = false;
    }
}
```

it will work by the charm and with all the automatic browsertesting that we have we will see if it is working or not.

tanx:
> fine for me, i've not particular comments but i'm new to es6 to understand if all the changes are correct but it seems so :)
> 
Exploring ES6 is a good online book that I can recommend. I happen to know the author in Munich.
> nice the adition of the zerocopy with transferable objects; what i do not like is that the feature is to be enabled manually by the user, and demand to the user to write the browserdetection logic and setting of openpgp.config.zeroCopy accordingly.
> i would really like more if we could add simply a test for it and set the variable zeroCopy automagically
> 
Unfortunately that won't work. It depends on the usecase and how the app uses openpgp.js. E.g if the app relies on Message ciphertext buffers to be there after decryption, it will break the app.

Like Bart said, let's keep that behind a flag for now. We can optimize behavior in future releases.

tanx:
So I've thought about how to proceed with this PR, given that it's so big and no one seems to have the time for a full review.

1. The initial array_buffer PR was written by Bart and reviewed and cleaned up my me. So it seems like we've already had 4 eyes on that part of the change set.

2. The ES6 module refactoring isn't very exciting as it only changes syntax. If there was a bug in that change set, the unit tests would have failed due to a module not having been found. That doesn't seem to be the case since everything runs fine.

3. The new api has already been discussed in depth here and I've already integrated all your feedback. We've also received positive feedback on the mailing list on the new api and that's good enough for me.

So I'm going to go ahead and merge this PR. New PRs are already being rebased against the 2.x branch. So we're basicallly treating it like the main branch already anyway.

I'll also release v2.0.0. Any bugs will be resolved as future bug fixes like always.

If someone depends on v1.x and is not ready to upgrade, please use the v1.x branch and cherry-pick any upstream bug fixes to that branch.

evilaliv3:
i just seen that the hash utils export the sha libraries allowing only to perform the digest at once.

this limits not only the possibilities for me in doing the streaming encryption (that i can do somehow anyway in a fork as discussed) but would prevent possible extensions of openpgp.js were it is needed to do sha block by block instead of at once.

evilaliv3:
and other doubts i've; if the idea is to keep the codebase clean with few dependencies, what is the need of having 3 different sha libraries (rusha, asmcrypto, jsSHA)

do we have important performance issues that prevent to include and trust only one?

tanx:
I don't like the fact that we use three different sha libs either. But since rusha and asmCrypto-lite are both very small, are included via npm, and have an enourmous performance impact, it seems to justify the overhead in the shortterm.

In the mid-to-longterm, since openpgp.js relies on several cipher libs, the direction I would like to go in the future is to include ciphers via npm and have openpgp.js provide only the high level orchestration and parsing functions of the OpenPGP standard.

I would like to remove rusha, the old aes-cfb, and rsa/jsbn and rely on native node.js/webcrypto where possible and fall back to asmCrypto-lite as our only cipher suite. This suite is maintained in a seperate github repository so that we could switch engines in the future e.g if a faster webassembly library pops up. I think it's better to reach that goal in baby steps though. see 2.0 release notes and future roadmap:

https://github.com/openpgpjs/openpgpjs/releases/tag/v2.0.0

evilaliv3:
perfect thanks for the clarification @tanx

PR #409, opened 2016-02-18
Added compression libraries to the repo

pzduniak:
This is needed in order to include openpgpjs directly in a webpack build. After it will be merged, it will be possible to do `import * as openpgp from 'openpgp/src/openpgp'` in a Webpack ES6 build.


evilaliv3:
@pzduniak that could be interesting but would it be possible to add it using npm/package.json instead of ripping the libraries?

pzduniak:
http://stackoverflow.com/questions/24942161/does-npm-ignore-files-listed-in-gitignore

Do we want to complicate it that much?

tanx:
We currently do not support web pack. Adding minified libs in the src dirs for a crypto lib is not gonna happen.

dwolfand:
@tanx We would love to re-open this issue since this is something that is more and more common these days given server-side minification which we do for performance on AWS Lambda. Thoughts?

bartbutler:
Should probably make an new issue with your exact problem. The library has changed quite a bit from when this discussion occurred and it's not obvious to me that what you want to do isn't already supported.

PR #410, opened 2016-02-22
Only use required files in bower main array

iDuuck:
This is because of `wiredep`. Don't know if this fits your repository requirements.


tanx:
People need to be able to choose if they want minified, worker files.

Issue #106, opened 2013-09-23
Simple speed-up for Chrome in jsbn.js


maxtaco:
There is some pretty ancient code in jsbn.js:

``` javascript
if(j_lm && (compat.get_navigator().appName == "Microsoft Internet Explorer")) {
       BigInteger.prototype.am = am2;
       dbits = 30;
}
else if ( j_lm && (compat.get_navigator().appName != "Netscape")) {
       BigInteger.prototype.am = am1;
       dbits = 26;
}
else { // Mozilla/Netscape seems to prefer am3
       BigInteger.prototype.am = am3;
       dbits = 28;
}
```

For modern browsers (Chrome/Safari/Mozilla), you wind up in the 2nd branch, but I find it faster in Chrome to wind up in the third (`dbits=28`).  For me, the fix is just to eliminate the first two branches. 
##

mahrud:
Closing this, as we've removed jsbn in favor of bn.js: #654.

PR #415, opened 2016-03-12
Resolves openpgpjs/openpgpjs#106 by introducing suggested fix

tanx:
Resolves openpgpjs/openpgpjs#106 by introducing suggested fix

**NEEDS profiling and verification**


acornejo:
Any reason this is not merged?

bartbutler:
I think it just needs profiling to make sure it is actually faster.

bartbutler:
I think the right move here is to eliminate jsbn in favor of one of the number of faster BN JS libraries around now. In the interest of not making anything any worse in the mean time, I'm going to close this PR.

Issue #97, opened 2013-08-24
FeatureRequest: Key Signing


ruebox:
Openpgp.js is really a great source for PGP support in the web. Thanks a lot.

Unfortunately, it is not yet able to sign a key with the private key. From my point of view, this feature is mandatory if you want to participate in the web of trust ;-)

Do you plan to integrate this in the future.

I am open to support the development in case you need some help :-)

Best, ruebox

alokmenghrajani:
I was able to sign keys using the following code. I'm not sure if it's correct or not, but it could be a good starting point?

```javascript
function signPublicKey(public_key, private_key) {
  // Returns a new public key which is public_key + signature(public_key)
  var dataToSign = {};
  dataToSign.userid = public_key.users[0].userId;
  dataToSign.key = public_key.primaryKey;

  var signaturePacket = new openpgp.packet.Signature();
  signaturePacket.signatureType = openpgp.enums.signature.cert_generic;
  signaturePacket.publicKeyAlgorithm = 1;
  signaturePacket.hashAlgorithm = 2;
  signaturePacket.keyFlags = [openpgp.enums.keyFlags.certify_keys | openpgp.enums.keyFlags.sign_data];
  signaturePacket.preferredSymmetricAlgorithms = [];
  signaturePacket.preferredSymmetricAlgorithms.push(openpgp.enums.symmetric.aes256);
  signaturePacket.preferredSymmetricAlgorithms.push(openpgp.enums.symmetric.aes192);
  signaturePacket.preferredSymmetricAlgorithms.push(openpgp.enums.symmetric.aes128);
  signaturePacket.preferredSymmetricAlgorithms.push(openpgp.enums.symmetric.cast5);
  signaturePacket.preferredSymmetricAlgorithms.push(openpgp.enums.symmetric.tripledes);
  signaturePacket.preferredHashAlgorithms = [];
  signaturePacket.preferredHashAlgorithms.push(openpgp.enums.hash.sha256);
  signaturePacket.preferredHashAlgorithms.push(openpgp.enums.hash.sha1);
  signaturePacket.preferredHashAlgorithms.push(openpgp.enums.hash.sha512);
  signaturePacket.preferredCompressionAlgorithms = [];
  signaturePacket.preferredCompressionAlgorithms.push(openpgp.enums.compression.zlib);
  signaturePacket.preferredCompressionAlgorithms.push(openpgp.enums.compression.zip);
[...]

PR #417, opened 2016-03-13
Added ability to sign and verify public keys

zartdinov:
Hi, I found the time to try to implement this feature in my favorite lib (Issue #97)!
Have tried to do it properly by RFC and keep code style. Watched other language openpgp implementations, tested by gpg and added some base tests.
Always ready to edit, rebase, ... thanks!

### DIFF
--- a/src/key.js
+++ b/src/key.js
@@ -639,83 +639,61 @@ Key.prototype.revoke = function() {
 };
 
 /**
- * Signs the public key
+ * Signs primary user of key
  * @param  {Array<module:key~Key>} privateKey decrypted private keys for signing
  * @return {module:key~Key} new public key with new certificate signature
  */
-Key.prototype.sign = function(privateKeys) {
-  var primaryUser, primaryKey, dataToSign, signingKeyPacket, signaturePacket, user, key;
-  if (this.isPrivate()) {
-    throw new Error('Only public key can be signed');
-  }
-  primaryUser = this.getPrimaryUser();
-  if (!primaryUser) {
+Key.prototype.signPrimaryUser = function(privateKeys) {
+  var {index, user} = this.getPrimaryUser() || {};
+  if (!user) {
     throw new Error('Could not find primary user');
   }
-  primaryKey = this.primaryKey;
-  dataToSign = {};
-  dataToSign.userid = primaryUser.user.userId || primaryUser.user.userAttribute;
-  dataToSign.key = primaryKey;
-  user = new User(primaryUser.user.userId || primaryUser.user.userAttribute);
-  user.otherCertifications = [];
-  privateKeys.forEach(function(privateKey) {
-    if (privateKey.isPublic()) {
-      throw new Error('Need private key for signing');
-    }
-    if (privateKey.primaryKey.getKeyId().equals(primaryKey.getKeyId())) {
-      throw new Error('No need to self signinig');
-    }
-    signingKeyPacket = privateKey.getSigningKeyPacket();
-    if (!signingKeyPacket) {
-      throw new Error('Could not find valid signing key packet');
-    }
-    if (!signingKeyPacket.isDecrypted) {
-      throw new Error('Private key is not decrypted.');
-    }
-    signaturePacket = new packet.Signature();
-    // Most OpenPGP implementations use generic certification (0x10)
-    signaturePacket.signatureType = enums.write(enums.signature, enums.signature.cert_generic);
-    signaturePacket.keyFlags = [enums.keyFlags.certify_keys | enums.keyFlags.sign_data];
-    signaturePacket.hashAlgorithm = privateKey.getPreferredHashAlgorithm();
-    signaturePacket.publicKeyAlgorithm = signingKeyPacket.algorithm;
-    signaturePacket.signingKeyId = signingKeyPacket.getKeyId();
-    signaturePacket.sign(signingKeyPacket, dataToSign);
-    user.otherCertifications.push(signaturePacket);
-  });
-  user.update(primaryUser.user, this.primaryKey);
-  key = new Key(this.toPacketlist());
-  key.users[primaryUser.index] = user;
+  user = user.sign(this.primaryKey, privateKeys);
+  var key = new Key(this.toPacketlist());
+  key.users[index] = user;
   return key;
 };
 
 /**
- * Verifies public key
+ * Signs all users of key
+ * @param  {Array<module:key~Key>} privateKeys decrypted private keys for signing
+ * @return {module:key~Key} new public key with new certificate signature
+ */
+Key.prototype.signAllUsers = function(privateKeys) {
+  var users = this.users.map(user => user.sign(this.primaryKey, privateKeys));
+  var key = new Key(this.toPacketlist());
+  key.users = users;
+  return key;
+};
+
+/**
+ * Verifies primary user of key
  * @param  {Array<module:key~Key>} keys array of keys to verify certificate signatures
  * @return {Array<({keyid: module:type/keyid, valid: Boolean})>} list of signer's keyid and validity of signature
  */
-Key.prototype.verify = function(keys) {
-  var primaryKey, primaryUser, user;
-  if (this.isPrivate()) {
-    throw new Error('Only public key can be verified');
-  }
-  primaryKey = this.primaryKey;
-  primaryUser = this.getPrimaryUser();
-  if (!primaryUser) {
+Key.prototype.verifyPrimaryUser = function(keys) {
+  var {user} = this.getPrimaryUser() || {};
+  if (!user) {
     throw new Error('Could not find primary user');
   }
-  user = primaryUser.user;
-  if (!user.otherCertifications) {
-    return [];
-  }
-  return user.otherCertifications.map(function(signaturePacket) {
-    var keyPacket = keys.find(function(key) {
-      return key.getSigningKeyPacket(signaturePacket.issuerKeyId);
-    }) || null;
-    return {
-      keyid: signaturePacket.issuerKeyId,
-      valid: keyPacket && signaturePacket.verify(keyPacket.primaryKey, {userid: user.userId || user.userAttribute, key: primaryKey})
-    };
-  });
+  return user.verifyAllSignatures(this.primaryKey, keys);
+};
+
+/**
+ * Verifies all users of key
+ * @param  {Array<module:key~Key>} keys array of keys to verify certificate signatures
+ * @return {Array<({userid: String, keyid: module:type/keyid, valid: Boolean})>} list of userid, signer's keyid and validity of signature
+ */
+Key.prototype.verifyAllUsers = function(keys) {
+  return this.users.reduce((signatures, user) => {
+    return signatures.concat(
+      user.verifyAllSignatures(this.primaryKey, keys).map(signature => ({
+        userid: user.userId.userid,
+        keyid: signature.keyid,
+        valid: signature.valid
+      }))
+    );
+  }, []);
 };
 
 /**
@@ -805,6 +783,64 @@ User.prototype.isValidSelfCertificate = function(primaryKey, selfCertificate) {
     return true;
   }
   return false;
+};
+
+/**
+ * Signs user
+ * @param  {module:packet/secret_key|module:packet/public_key} primaryKey The primary key packet
+ * @param  {Array<module:key~Key>} privateKeys decrypted private keys for signing
+ * @return {module:key~Key} new user with new certificate signatures
+ */
+User.prototype.sign = function(primaryKey, privateKeys) {
+  var user, dataToSign, signingKeyPacket, signaturePacket;
+  dataToSign = {};
+  dataToSign.key = primaryKey;
+  dataToSign.userid = this.userId || this.userAttribute;
+  user = new User(this.userId || this.userAttribute);
+  user.otherCertifications = [];
+  privateKeys.forEach(privateKey => {
+    if (privateKey.isPublic()) {
+      throw new Error('Need private key for signing');
+    }
+    if (privateKey.primaryKey.getKeyId().equals(primaryKey.getKeyId())) {
+      throw new Error('Not implemented for  self signing');
+    }
+    signingKeyPacket = privateKey.getSigningKeyPacket();
+    if (!signingKeyPacket) {
+      throw new Error('Could not find valid signing key packet');
+    }
+    if (!signingKeyPacket.isDecrypted) {
+      throw new Error('Private key is not decrypted.');
+    }
+    signaturePacket = new packet.Signature();
+    // Most OpenPGP implementations use generic certification (0x10)
+    signaturePacket.signatureType = enums.write(enums.signature, enums.signature.cert_generic);
+    signaturePacket.keyFlags = [enums.keyFlags.certify_keys | enums.keyFlags.sign_data];
+    signaturePacket.hashAlgorithm = privateKey.getPreferredHashAlgorithm();
+    signaturePacket.publicKeyAlgorithm = signingKeyPacket.algorithm;
+    signaturePacket.signingKeyId = signingKeyPacket.getKeyId();
+    signaturePacket.sign(signingKeyPacket, dataToSign);
+    user.otherCertifications.push(signaturePacket);
+  });
+  user.update(this, primaryKey);
+  return user;
+};
+
+/**
+ * Verifies all user signatures
+ * @param  {module:packet/secret_key|module:packet/public_key} primaryKey The primary key packet
+ * @param  {Array<module:key~Key>} keys array of keys to verify certificate signatures
+ * @return {Array<({keyid: module:type/keyid, valid: Boolean})>} list of signer's keyid and validity of signature
+ */
+User.prototype.verifyAllSignatures = function(primaryKey, keys) {
+  var dataToVerify = { userid: this.userId || this.userAttribute, key: primaryKey };
+  return this.selfCertifications.concat(this.otherCertifications).map(signaturePacket => {
+    var keyPacket = keys.find(key => key.getSigningKeyPacket(signaturePacket.issuerKeyId)) || null;
+    return {
+      keyid: signaturePacket.issuerKeyId,
+      valid: keyPacket && signaturePacket.verify(keyPacket.primaryKey, dataToVerify)
+    };
+  });
 };
 
 /**
--- a/src/key.js
+++ b/src/key.js
@@ -802,8 +802,8 @@ User.prototype.sign = function(primaryKey, privateKeys) {
     if (privateKey.isPublic()) {
       throw new Error('Need private key for signing');
     }
-    if (privateKey.primaryKey.getKeyId().equals(primaryKey.getKeyId())) {
-      throw new Error('Not implemented for  self signing');
+    if (privateKey.primaryKey.getFingerprint() === primaryKey.getFingerprint()) {
+      throw new Error('Not implemented for self signing');
     }
     signingKeyPacket = privateKey.getSigningKeyPacket();
     if (!signingKeyPacket) {
--- a/src/key.js
+++ b/src/key.js
@@ -834,8 +834,9 @@ User.prototype.sign = function(primaryKey, privateKeys) {
  */
 User.prototype.verifyAllSignatures = function(primaryKey, keys) {
   var dataToVerify = { userid: this.userId || this.userAttribute, key: primaryKey };
-  return this.selfCertifications.concat(this.otherCertifications).map(signaturePacket => {
-    var keyPacket = keys.find(key => key.getSigningKeyPacket(signaturePacket.issuerKeyId)) || null;
+  var certificates = this.selfCertifications.concat(this.otherCertifications || []);
+  return certificates.map(signaturePacket => {
+    var keyPacket = keys.filter(key => key.getSigningKeyPacket(signaturePacket.issuerKeyId))[0] || null;
     return {
       keyid: signaturePacket.issuerKeyId,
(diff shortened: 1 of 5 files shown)


zartdinov:
I know about failed tests. I used only node.js stable, will try to fix soon.

tanx:
Thanks for the PR. I haven't looked at the key signing code in detail yet, but it should not be necessary to add this functionality to the high level api in openpgp.js. It should be totally ok to call these functions from the key.js module.

bartbutler:
Agreed.

zartdinov:
No problem! Can you give your opinion at next moments
1) Should `Key.proto.sign` return new `Key` object or just modify current like `Key.proto.decrypt`?
2) Should `Key.verify` return signatures statuses like `Message.proto.verify`  or just `true`, `false` or `enums.keyStatus.valid`, `enums.keyStatus.invalid` like `SubKey.proto.verify`?

In both cases, I want to sign and verify with one key instead of array, so modifiying key will be faster if someone will decide make it in the loop

tanx:
> 1) Should Key.proto.sign return new Key object or just modify current like Key.proto.decrypt?

I would suggest to have it behave like `Message.prototype.sign` returning a new Key object.

> 2) Should Key.verify return signatures statuses like Message.proto.verify or just true, false or enums.keyStatus.valid, enums.keyStatus.invalid like SubKey.proto.verify?

From a business logic standpoint `enums.keyStatus` seems to cover different semantics. So having verify return either true/false/null like `Message.prototype.verify` seems to make sense to me. @toberndo what do you think?

> In both cases, I want to sign and verify with one key instead of array, so modifiying key will be faster if someone will decide make it in the loop

I actually like your current api choice passing in an array of keys. Again this would be consistent with the Message api.

src/key.js
```diff
@@ -623,6 +623,86 @@ Key.prototype.revoke = function() {
...
+    signaturePacket.hashAlgorithm = privateKey.getPreferredHashAlgorithm();
+    signaturePacket.publicKeyAlgorithm = signingKeyPacket.algorithm;
+    signaturePacket.signingKeyId = signingKeyPacket.getKeyId();
+    signaturePacket.sign(signingKeyPacket, dataToSign);
+    user.otherCertifications.push(signaturePacket);
+  });
+  user.update(primaryUser.user, this.primaryKey);
+  key = new Key(this.toPacketlist());
+  key.users[primaryUser.index] = user;
+  return key;
```
tanx: Returning a new key obj here looks like the right approach here.

src/key.js
```diff
@@ -623,6 +623,86 @@ Key.prototype.revoke = function() {
...
+  }
+  return user.otherCertifications.map(function(signaturePacket) {
+    var keyPacket = keys.find(function(key) {
+      return key.getSigningKeyPacket(signaturePacket.issuerKeyId);
+    }) || null;
+    return {
+      keyid: signaturePacket.issuerKeyId,
+      valid: keyPacket && signaturePacket.verify(keyPacket.primaryKey, {userid: user.userId || user.userAttribute, key: primaryKey})
+    };
+  });
```
tanx: The return type looks fine to me.

src/key.js
```diff
@@ -623,6 +623,86 @@ Key.prototype.revoke = function() {
...
+  key.users[primaryUser.index] = user;
+  return key;
+};
+
+/**
+ * Verifies public key
+ * @param  {Array<module:key~Key>} keys array of keys to verify certificate signatures
+ * @return {Array<({keyid: module:type/keyid, valid: Boolean})>} list of signer's keyid and validity of signature
+ */
+Key.prototype.verify = function(keys) {
```
tanx: The function api looks good, passing in an array of Key objects.

src/key.js
```diff
@@ -623,6 +623,86 @@ Key.prototype.revoke = function() {
 };
 
 /**
+ * Signs the public key
+ * @param  {Array<module:key~Key>} privateKey decrypted private keys for signing
+ * @return {module:key~Key} new public key with new certificate signature
+ */
+Key.prototype.sign = function(privateKeys) {
```
tanx: The function api looks good, passing in an array of Key objects.

src/openpgp.js
```diff
@@ -344,6 +344,65 @@ export function decryptSessionKey({ message, privateKey, password }) {
...
+  }
+
+  return execute(() => ({
+
+    signatures: publicKey.verify(publicKeys)
+
+  }), 'Error verifying signed public key');
+}
+
+
```
tanx: Like mentioned before, the changes to this file can be reverted. Signing and verifying public keys is a very specific use case that can be done by invoking a private api. We should try to keep the public api as clean as possible to make it easy for newcomers to use the library. Advanced use cases will require private api use anyway.

src/openpgp.js
```diff
@@ -344,6 +344,65 @@ export function decryptSessionKey({ message, privateKey, password }) {
...
+  checkKey(publicKey, 'publicKey');
+  privateKeys = toArray(privateKeys);
+
+  if (asyncProxy) { // use web worker if available
+    return asyncProxy.delegate('signPublicKey', { publicKey, privateKeys });
+  }
+
+  return execute(() => {
+
+    const signedPublicKey = publicKey.sign(privateKeys);
```
tanx: Since the user needs to pass in the public Key object here anyway, they could just call `publicKey.sign(privateKeys)` directly.

src/key.js
```diff
@@ -623,6 +623,86 @@ Key.prototype.revoke = function() {
...
+    throw new Error('Only public key can be signed');
+  }
+  primaryUser = this.getPrimaryUser();
+  if (!primaryUser) {
+    throw new Error('Could not find primary user');
+  }
+  primaryKey = this.primaryKey;
+  dataToSign = {};
+  dataToSign.userid = primaryUser.user.userId || primaryUser.user.userAttribute;
+  dataToSign.key = primaryKey;
```
tanx: I'm not familiar with the RFC here. Do we only sign the primary key or also the subkeys @toberndo ?

src/key.js
```diff
@@ -491,11 +491,11 @@ function getExpirationTime(keyPacket, selfCertificate) {
 Key.prototype.getPrimaryUser = function() {
   var primUser = [];
   for (var i = 0; i < this.users.length; i++) {
-    if (!this.users[i].userId || !this.users[i].selfCertifications) {
+    if ((!this.users[i].userId && !this.users[i].userAttribute) || !this.users[i].selfCertifications) {
```
toberndo: This changes the semantic of the method. I think getPrimaryUser should only return users with userId. In the case that userId is initial and userAttribute is set the user should be filtered out.

src/key.js
```diff
@@ -623,6 +623,86 @@ Key.prototype.revoke = function() {
...
 
 /**
+ * Signs the public key
+ * @param  {Array<module:key~Key>} privateKey decrypted private keys for signing
+ * @return {module:key~Key} new public key with new certificate signature
+ */
+Key.prototype.sign = function(privateKeys) {
+  var primaryUser, primaryKey, dataToSign, signingKeyPacket, signaturePacket, user, key;
+  if (this.isPrivate()) {
+    throw new Error('Only public key can be signed');
```
toberndo: Not required to allow only public keys. Would also like to sign private ones.

src/key.js
toberndo: You sign User-Ids. So this functionality should be located at User.prototype.

tanx: I don't fully understand. So you only sign user IDs and not the public key itself? For example see my uploaded public key:

https://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0xBFB264F6

This brings up the next question... shouldn't we sign all of the key's user IDs as it can have more than one?

zartdinov: Sorry, I am little busy now. Signature makes pair { primaryUser, primaryKey } trustable
```
dataToSign = {};
dataToSign.userid = primaryUser.user.userId || primaryUser.user.userAttribute;
dataToSign.key = primaryKey;
```

toberndo: Correct, the thing you do on key signing parties is normally to sign User-Ids. There are also more exotic use cases for "direct signatures" on keys, but that should be out of scope here.
So I would propose to have a User.prototype.sign method (one private key as a parameter should be enough, I don't see the use case for an array). We could have helper methods like Key.prototype.signAllUser and Key.prototype.signPrimaryUser.
On the verification side we should go for a User.prototype.verifyOtherCertifications (or better name) and an array of keys as parameter. Here we have the same problem as with Message.prototype.verify: you don't know in advance which keys to hand over. Basically you have to look in Users.otherCertifications to determine the keyids, or you hand over your complete keyring. I once experimented with a method to supply the keys via a callback (https://github.com/mailvelope/openpgpjs/commit/4ffa64ae3e7507245ab9e2fadba3c1d25ccbc721), but have not gone further here.
As a return parameter for verifyOtherCertifications I'm leaning towards the same interface as used in Message.prototype.verify: true, false, null.

src/key.js
zartdinov: Yeah, I will discrd this changes.
Anyway if it no primary user is found returns first valid self signed user

src/key.js
zartdinov: `Key.prototype.signAllUsers` is well-looking option to set some day another user as primary and not collect again all trusted signatures.
About functionality located at User.prototype, we will have to also always pass primaryKey (of public key) to prototype methods, cause they need both to sign and verify.
If I understand correctly userIds cant be signed without related keyid, otherwise someone could put own primaryKey and self-signature to trusted public key.
About signing of other user keys, I think it is not neccesary cause they are bound with primaryKey by binding signatures

toberndo: > Key.prototype.signAllUsers is well-looking option to set some day another user as primary and not collect again all trusted signatures.

Yes, that is one option to create a new signature for all userIds. But it should also be enough to sign only one userId with the primary flag set. In the end, the newest signature with the primary flag sets the useId as "primary".

> About functionality located at User.prototype, we will have to also always pass primaryKey (of public key) to prototype methods, cause they need both to sign and verify.

Unfortunately yes.

> If I understand correctly userIds cant be signed without related keyid, otherwise someone could put own primaryKey and self-signature to trusted public key.

What do you mean by "related keyid"? The keyid of the primary key? I think two primaryKey packets would be a violation of the spec. If you replace the primaryKey and add a valid userId+signature, then you basically have a new key. But if available, subkey binding signatures have to match as well.

zartdinov: What do u think about for the first time realize functionality in `Key.proto`:
```javascript
Key.proto.signPrimaryUser(privateKeys) => newKey,
Key.proto.signAllUsers(privateKeys) => newKey,
Key.proto.verifyPrimaryUser(publicKeys) => [
    {
        keyid: ...
        valid: ...
    },
    ...
],
Key.proto.verifyAllUsers(publicKeys) => [
    {
        userid: ...
        keyid: ...
        valid: ...
    },
    ...
]
```
And later change inner structure, move some part to User.proto, etc..

toberndo: :+1: I think it would be definitely good to have these high level methods implemented via low level methods in User.proto in order to make it easy to achieve the following use cases:

1. Sign specific user xyz
2. Verify all signatures of user xyz

zartdinov:
I think something like that

zartdinov:
Seems that Saucelabs is a little bit unreliable, sometimes service can be overloaded

danenania:
Hey folks, thanks for all your hard work on this library--it's the backbone of an app I'm working on.

I was just wondering what's up with this PR? How current/usable/safe is the code that's been written? Public key signing seems like an important feature for apps with end-to-end encryption that generate and distribute their own keys. Without it, it's easy for a compromised/subpoenaed server to spoof public keys. What would it take to get this production-ready at this point?

bartbutler:
To be honest I'm not sure. I'm behind looking at pull requests. This one looks pretty solid and has tests, so other than the minor conflict it could probably get merged without too much effort.

danenania:
@bartbutler Ok, that's what I was thinking too. I'll try to deal with that conflict as soon as I have some time.

zartdinov:
Hi, I resolved conflict and replayed my commits onto the latest version of master.
But I need to find more time to look on repository changes and refresh my memories.

danenania:
@zartdinov Awesome. You've probably got it covered, but let me know if I can help at all. This is a really great addition.

zartdinov:
Sorry for my inactivity. I think I don't have a right to say something about production use.
But mainly code was taken from key generation part which based on same principles.

This addition is very significant for PGP's "Web of Trust" concept (Authority servers, etc.).
If I remember right generating a new key with same fingerprint takes a just few seconds on GPU.

Looks like PR code really is very separeted from existing lib functionality.
Maintainers, what to do next? I am ready to refactor, rebase after any review or audit. Thanks.

bartbutler:
@zartdinov I'm a little swamped but I'll try to look at it and do a final review in the next week or so.

danenania:
@bartbutler Just wanted to check in on this... are still planning to review it soon?

src/key.js
```diff
@@ -728,6 +786,65 @@ User.prototype.isValidSelfCertificate = function(primaryKey, selfCertificate) {
...
+  dataToSign.key = primaryKey;
+  dataToSign.userid = this.userId || this.userAttribute;
+  user = new User(this.userId || this.userAttribute);
+  user.otherCertifications = [];
+  privateKeys.forEach(privateKey => {
+    if (privateKey.isPublic()) {
+      throw new Error('Need private key for signing');
+    }
+    if (privateKey.primaryKey.getKeyId().equals(primaryKey.getKeyId())) {
+      throw new Error('Not implemented for  self signing');
```
bartbutler: Extra space in error message. Also, what happens in the case of KeyID collision? Do we want to disallow signing? Comparing the full key fingerprint would be better I think.

src/key.js
```diff
@@ -728,6 +786,65 @@ User.prototype.isValidSelfCertificate = function(primaryKey, selfCertificate) {
...
+      throw new Error('Private key is not decrypted.');
+    }
+    signaturePacket = new packet.Signature();
+    // Most OpenPGP implementations use generic certification (0x10)
+    signaturePacket.signatureType = enums.write(enums.signature, enums.signature.cert_generic);
+    signaturePacket.keyFlags = [enums.keyFlags.certify_keys | enums.keyFlags.sign_data];
+    signaturePacket.hashAlgorithm = privateKey.getPreferredHashAlgorithm();
+    signaturePacket.publicKeyAlgorithm = signingKeyPacket.algorithm;
+    signaturePacket.signingKeyId = signingKeyPacket.getKeyId();
+    signaturePacket.sign(signingKeyPacket, dataToSign);
```
bartbutler: I'm a little fuzzy on key signing, but shouldn't we also include a Signer's User ID (RFC 4880 5.2.3.22) sub packet here? Otherwise there's no information about who the signer is except for the KeyID, which is short and not terribly unique. How does GPG do this by default?

src/key.js
```diff
@@ -728,6 +786,65 @@ User.prototype.isValidSelfCertificate = function(primaryKey, selfCertificate) {
...
+ * Verifies all user signatures
+ * @param  {module:packet/secret_key|module:packet/public_key} primaryKey The primary key packet
+ * @param  {Array<module:key~Key>} keys array of keys to verify certificate signatures
+ * @return {Array<({keyid: module:type/keyid, valid: Boolean})>} list of signer's keyid and validity of signature
+ */
+User.prototype.verifyAllSignatures = function(primaryKey, keys) {
+  var dataToVerify = { userid: this.userId || this.userAttribute, key: primaryKey };
+  var certificates = this.selfCertifications.concat(this.otherCertifications || []);
+  return certificates.map(signaturePacket => {
+    var keyPacket = keys.filter(key => key.getSigningKeyPacket(signaturePacket.issuerKeyId))[0] || null;
```
bartbutler: I think it might help if this iterates over key packets which match the search KeyID rather than just using the first one. It would iterate until a valid signature was found or no more key packets. That way it will work on keys with a KeyID collision, as I think multiple key packets with the same KeyID are allowed. The result in the case of multiple _input_ keys with the same KeyID would be ambiguous, but that should be a vanishingly rare case. Thoughts?

bartbutler:
@zartdinov @danenania Thanks for the push :) Just did, looks good, though I have a couple questions.

gentauro:
Will this logic be exposed through you `dist` -> `openpgp.min.js` file?
- https://github.com/openpgpjs/openpgpjs/tree/master/dist

bartbutler:
Yes, see the tests added in the PR for syntax examples.

gentauro:
This is pretty awesome !!!

- [x] Domain: sign2x.org
- [x] HTTPS: Almost https://gitlab.com/gitlab-org/gitlab-ce/issues/28857#note_24578234
- [x] Key signing: Yeah baby

Soon "Sign Sign" (sign2x) will become a reality !!!

sign_sign

Thx guys for your efforts :+1:

src/key.js
zartdinov: Extra space will be deleted.
I wanted to prevent using this function for self signing. Because it creates more simplified signature (without preferred algorithms, key servers). Second reason is that every key-user are already signed by its key on generation stage and I don't know what happens if we will add the extra one.
Yeah, I think we can change it on comparing of fingerprints.

src/key.js
zartdinov: Honestly, this one has very strange RFC definition.
After some researches I decided that it is not widely used.

In my opnion if full public key can cryptographicaly verify signature then it's ok.
Every key is also cryptographicaly bounded with some user and we can easily get this information.
May be it is more correct way to avoid using short ids.
If it was really important we would be must see it in message signinig too.
 
Anyway I can try to look how it is impemented in other libraries.

src/key.js
zartdinov: Yeah, right. I will change it.

src/key.js
bartbutler: Thanks. The difference I see between a message and a key signing is that with a message you have a sender that you can lookup. With key signing, I'd have to rely on short IDs to do the lookup if I don't have some UserId information to help me out. It'll work, but the obvious question of 'who has signed this key' will be more time-consuming to answer and verify. In the end, I'd be happy doing whatever GPG does.

src/key.js
bartbutler: Great.

src/key.js
zartdinov: Ok, about PGP. I didn't look at the sources.
Just tried to sign some test key and check packets list.
I used **gpg --list-packets** and **pgpdump** too.
Looks like GPG doesn't add this packet.

bartbutler: Okay, good enough for me.

bartbutler:
Merged, released as OpenPGPjs 2.4.0. Thanks @zartdinov!

zartdinov:
Great! Thank you all!

danenania:
Woot! Well done folks.

gentauro:
I found out that it's possible to sign a `public key` several times with the same `private key`. I don't know if this gives any sense?
```
    function signKey(pubkey,password){
        var privkey = privKeyring();
        
        var key        = openpgp.key.readArmored(pubkey) .keys[0];
        var privateKey = openpgp.key.readArmored(privkey).keys[0];
        
        privateKey.decrypt(password);

        var sigs = key.verifyPrimaryUser([privateKey]);

        // If already signed with this private key, just return pubkey
        for (var i = 0; i < sigs.length; ++i){
            (function(sig) {
                if(sig.keyid.toHex() ===
                   privateKey.getSigningKeyPacket().getKeyId().toHex())
                {
                    return pubkey;
                }
            })(sigs[i]);
        }

        return key.signPrimaryUser([privateKey]).armor();
    }
```

Therefore I added a local check that if a `key` is already signed, then just return it as it is ... but shouldn't this be built-in to `signPrimaryUser` and `signAllUsers`?

zartdinov:
Yeah, I didn't consider this scenario.
The key that contains absolutly same signatures is not good looking for me.
Also It's not very nice look that you have to do it.  Verification is time-consuming process.
I think it should be build-in.

gentauro:
A friend of mine who works with `security` tells me that it's possible with OpenPGP and GnuPGP to do this as a key signed has a timestamp.

An example: _"If you meet with a friend once a year, by re-signing his/her key, the trust gets stronger"_. 

Nevertheless, I don't think `openpgpjs` have the trust implemented yet right? Therefore you might just leave the code like it is now and a positive side-effect would be that no `verification` (time consuming) is needed :+1:

gentauro:
A bit more feedback. I noticed when I try to sign 10 keys with the same `private key`, that it takes some time.

I tried to use the built-in in `worker`, but as this functionality is added on the instantiated `key` and not as a library taking a `key` as `parameter` the following code will not work:
```js
    var privateKey = openpgp.key.readArmored(privkey).keys[0];
    privateKey.decrypt(password);
    
    var worker = openpgp.getWorker();
    worker.delegate(
        /* This works:
        'sign',
        {
            data: 'Hello, World!',
            privateKeys: privateKeys
        }
        and this, doesn't */
        'signPrimaryUser',
        {
            privateKeys: [ privateKey ]
        }
    ).then(
        function(result) {
            var baz = result;
        }
    ).catch(
        function(err) {
            var qux = err;
        }
    );
```

> **Note:** Error message from `Promise`, `"Unknown Worker Event"`

bartbutler:
We can probably add a high level signKey function or the like to the openpgp object that uses the worker.

gentauro:
> We can probably add a high level signKey function or the like to the openpgp object that uses the worker.

That sounds good. I found out that all related to `sign` (specially if it's `deattached` file) takes long time. It would be ideal to have those heavy calculations in a `worker` thread.

zartdinov:
It would be nice to use this like other main tools.

PR #435, opened 2016-04-06
allow to use worker from blob url

ssendev:
This allows to use the worker when serving from `file://` to allow portable mini apps.

``` html
<script id='openpgp-worker' type='text/javascript-worker'>// openpgp.worker.js</script>
<script id='openpgp' type='text/javascript'>// openpgp.js</script>
<script>
  var worker = new Worker(
    window.URL.createObjectURL(
      new Blob([document.getElementById('openpgp-worker').innerHTML],
               { type: 'text/javascript' })));
  worker.postMessage({ event: 'openpgp-url', 
                       url: window.URL.createObjectURL(
                         new Blob([document.getElementById('openpgp').innerHTML],
                                  { type: 'text/javascript' })) });
  openpgp.initWorker({ worker: worker });
</script>
```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -108,19 +108,22 @@ module.exports = function(grunt) {
           to: 'OpenPGP.js v<%= pkg.version %>'
         }]
       },
-      worker_min: {
-        src: ['dist/openpgp.worker.min.js'],
-        dest: ['dist/openpgp.worker.min.js'],
+      openpgp_min: {
+        src: ['dist/openpgp.js'],
+        dest: ['dist/openpgp.min.js'],
         replacements: [{
-          from: "importScripts('openpgp.js')",
-          to: "importScripts('openpgp.min.js')"
+          from: "'openpgp.worker.js'",
+          to: "'openpgp.worker.min.js'"
+        }, {
+          from: "'openpgp.js'",
+          to: "'openpgp.min.js'"
         }]
       }
     },
     uglify: {
       openpgp: {
         files: {
-          'dist/openpgp.min.js' : [ 'dist/openpgp.js' ],
+          'dist/openpgp.min.js' : [ 'dist/openpgp.min.js' ],
           'dist/openpgp.worker.min.js' : [ 'dist/openpgp.worker.min.js' ]
         }
       },
--- a/src/openpgp.js
+++ b/src/openpgp.js
@@ -54,11 +54,12 @@ let asyncProxy; // instance of the asyncproxy
 /**
  * Set the path for the web worker script and create an instance of the async proxy
  * @param {String} path     relative path to the worker scripts, default: 'openpgp.worker.js'
+ * @param {String} pgpPath  relative path to OpenPGP.js, default: 'openpgp.js'
  * @param {Object} worker   alternative to path parameter: web worker initialized with 'openpgp.worker.js'
  */
-export function initWorker({ path='openpgp.worker.js', worker } = {}) {
+export function initWorker({ path='openpgp.worker.js', pgpPath='openpgp.js', worker } = {}) {
   if (worker || typeof window !== 'undefined' && window.Worker) {
-    asyncProxy = new AsyncProxy({ path, worker, config });
+    asyncProxy = new AsyncProxy({ path, pgpPath, worker, config });
     return true;
   }
 }
--- a/src/worker/async_proxy.js
+++ b/src/worker/async_proxy.js
@@ -28,16 +28,18 @@ const INITIAL_RANDOM_SEED = 50000, // random bytes seeded to worker
  * Initializes a new proxy and loads the web worker
  * @constructor
  * @param {String} path     The path to the worker or 'openpgp.worker.js' by default
+ * @param {String} pgpPath  The path to OpenPGP.js or 'openpgp.js' by default
  * @param {Object} config   config The worker configuration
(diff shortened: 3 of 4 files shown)


tanx:
I don't get it. What's the advantage of specifying the worker path via `.innerHTML` instead of `openpgp.initWorker({ path:'openpgp.worker.js' });`? It would still require fetching two separate script, no?

If there is a way to build a single `openpgp.js` file with the worker built in, I'm all for it. But this PR seems to address only half of the required change. If you could patch the gruntfile and main script accordingly I would gladly review/merge the proposed changes.

*P.S. @evilaliv3 if this works, we could also investigate if the Promise polyfill could also be included using this approach and left out of the main release.*

ssendev:
@tanx With this technique it would indeed be possible to ship only one js file with an included worker. But that is not what this pull request is about. (and it looks like browser support would not suffer http://caniuse.com/#feat=getrandomvalues http://caniuse.com/#feat=bloburls)

What this tries to solve is that when you open a html file from the local file system for security reasons you can't load additional files. in this case openpgp.worker.js. The workaround is to include the required file inline and create a blob url. but since opengpg.worker.js is just a simple wrapper and loads openpgp.js this fails since the worker url then is blob://123random.js and there is no ./openpgp.js which could be loaded. this patch allows to pass the url of opengpg.js to the worker.

fpietrosanti:
super cool

evilaliv3:
@tanx +1 for having in the worker a message in order to push inside the worker additional libraries.

it's an interesting direction of research where the worker could try to import libraries (e.g. openpgp.js.min) only if it the library is not injected from outside.

i'm get the point of @ssendev that in a superminified library is not acceptable that:
1) for generating a worker one has to have the code of a worker in a file; the bypass on this is possible only using Blob while if one uses datauri any browser would blame about the different domain of the worker so exactly and only with the code written by him.
2) the worker ```openpgp.worker.min.js``` makes us of ```openpgp.min.js``` so possibly causing to re-download the openpgp.min.js a second time; this is also already solvable during the build process of the tool that makes use of globaleals; one has simply to embed the ```openpgp.min.js``` as a data uri inside the worker and that would work like a charm (we are already doing this in GlobaLeaks)!

ssendev:
Here is some sugar. Now it is possible to specify the url like this:

```html
<script id='openpgp-worker' type='text/javascript-worker'>// openpgp.worker.js</script>
<script id='openpgp' type='text/javascript'>// openpgp.js</script>
<script>
  openpgp.initWorker({
    path:    window.URL.createObjectURL(
               new Blob([document.getElementById('openpgp-worker').innerHTML],
                        { type: 'text/javascript' })),
    pgpPath: window.URL.createObjectURL(
               new Blob([document.getElementById('openpgp').innerHTML],
                        { type: 'text/javascript' }))
  });
</script>
```

ssendev:
And an example app to demonstrate the usefulness of this pull request. https://github.com/ssendev/pass2go

Just download, open and observe the worker still working although it is served from `file:///home/user/downloads/pass2go.html`

evilaliv3:
delicious PoC!

@tanx i've no particular critics on the pull, maybe you have on the stile but i'm really confident that there is and interesting added value in this pull.

tanx:
> @tanx With this technique it would indeed be possible to ship only one js file with an included worker.

Great. I'd like users to be able to include the library including worker in a single script tag like so:

```html
<script src="openpgp.js"></script>
```

The current suggestion with IDs and DOM queries looks far too complex for the mainline release IMHO.

ssendev:
Then there are two paths to be taken

the easy way: leave everything largely untouched und wrap it a la
```javascript
pgp = 'contents of openpgp.js';
worker = 'contents of worker.js';
eval(pgp);
openpgp.initworker(path: ... worker, pgpPath: ... pgp);
```

the hard way: figure out wich parts should go into the worker and wich parts belong into main
```javascript
// the main parts
worker = 'the worker parts';
if (useWorker)
  openpgp.initWorker(path: ... worker);
else 
  eval(worker);
```

caniuse.com suggests that all browsers supporting getrandomvalues support webworker so it would be possible to require usage of a webworker without raising browser requirements. if that is not desired the worker parts could be written in such a way that they can be evaluated in main which would then patch openpgp to include the worker parts. to avoid the eval when no worker is present an openpgp.noworker.js would be a workaround.


yes the current solution is not straight forward but it enables things that are otherwise not possible and it should be fairly backwards compatible. the only thing i can think of breaking is when someone did this.
```html
<script src='openpgp.js'></script>
<script type='text/javascript-worker'>
  worker = new Worker('worker.js');
  worker.postMessage('whatever'); // this breaks as openpgp.js isn't loaded within the worker
  openpgp.initWorker(worker: worker);
  worker.postMessage('whatever'); // now it is ok since openpgp.js is loaded
</script>
```

but if that is too big a break it would be possible to detect the worker being loaded over blob like in my first commit which would not have this problem but might frustrate users which want to use this for e.g. digested urls like openpgp-12345678.js which would allow better caching

ssendev:
@tanx I made a grunt task to build `openpgp.with-worker.min.js` if you are interested i could rebase or create a new pull request. https://github.com/ssendev/openpgpjs/tree/batteries-included

This uses the "easy" way. To avoid the giant eval someone with a little more knowledge of the internals would have to separate the worker from the main parts.

twiss:
This can also be done by taking the code in the OP and replacing `importScripts('openpgp.js');` with `document.getElementById('openpgp').innerHTML` in the worker source.

Issue #443, opened 2016-04-23
Add Key.encrypt(passphrase) to exported API


NSkelsey:
In the API documentation we find [`key.prototype.decrypt`](https://openpgpjs.org/openpgpjs/doc/key.js.html#line373). However, there is no accompanying `key.prototype.encrypt` for [`Key`](https://openpgpjs.org/openpgpjs/doc/module-key-Key.html) objects.

Either this was a specific design choice or it is genuinely missing – please advise.

The current code I use to preform key.encrypt(passphrase) looks like:
```javascript
// lockKeyRing encrypts the private key material of the key using the passed 
// password. The primary key and all subkeys are encrypted with the password.
// { string -> bool }
function lockKeyRing(password) {
  var w = true;
  w = w && keyRing.privateKey.primaryKey.encrypt(password);
  keyRing.privateKey.subKeys.forEach(function(sk) {
      w = w && sk.subKey.encrypt(password);
   });
  return w;
}
```
Note that the keyRing here is a custom data structure.

P.S. Thanks for the library it is fantastic!

evilaliv3:
Thanks @NSkelsey for the suggestion; I've written a pull request that follows the same pattern of the decrypt API and that is already used by various projects like whiteout-io: https://github.com/whiteout-io/mail/blob/master/src/js/crypto/pgp.js#L204

@tanx / @toberndo what do you think?

tanx:
See https://github.com/openpgpjs/openpgpjs/pull/446

PR #446, opened 2016-04-24
Implement feature #443

evilaliv3:
This pull request implement a Key API to the key with a passphrased following the same encrypt API.

This way combining `decrypt``+ `encrypt` one could easily and securely change the passphrase.

The pull request include the unit tests.

@tanx could you take care of review this and eventually push this out in the next useful release?

### DIFF
--- a/src/key.js
+++ b/src/key.js
@@ -377,6 +377,7 @@ Key.prototype.encrypt = function(passphrase) {
   var keys = this.getAllKeyPackets();
   for (var i = 0; i < keys.length; i++) {
     keys[i].encrypt(passphrase);
+    keys[i].clearPrivateMPIs();
   }
 };
 
--- a/src/key.js
+++ b/src/key.js
@@ -374,16 +374,11 @@ Key.prototype.encrypt = function(passphrase) {
   if (this.isPrivate()) {
     var keys = this.getAllKeyPackets();
     for (var i = 0; i < keys.length; i++) {
-      try {
-        keys[i].encrypt(passphrase);
-      } catch (e) {
-        return false;
-      }
+      keys[i].encrypt(passphrase);
     }
   } else {
     throw new Error("Nothing to encrypt in a public key");
   }
-  return true;
 };
 
 /**
--- a/src/key.js
+++ b/src/key.js
@@ -366,6 +366,27 @@ Key.prototype.getEncryptionKeyPacket = function() {
 };
 
 /**
+ * Encrypts all secret key and subkey packets
+ * @param  {String} passphrase
+ * @return {Boolean} true if all key and subkey packets encrypted successfully
+ */
+Key.prototype.encrypt = function(passphrase) {
+  if (this.isPrivate()) {
+    var keys = this.getAllKeyPackets();
+    for (var i = 0; i < keys.length; i++) {
+      try {
+        keys[i].encrypt(passphrase);
+      } catch (e) {
+        return false;
+      }
+    }
+  } else {
+    throw new Error("Nothing to encrypt in a public key");
+  }
+  return true;
+};
+
+/**
  * Decrypts all secret key and subkey packets
  * @param  {String} passphrase
  * @return {Boolean} true if all key and subkey packets decrypted successfully
--- a/src/key.js
+++ b/src/key.js
@@ -368,7 +368,6 @@ Key.prototype.getEncryptionKeyPacket = function() {
 /**
  * Encrypts all secret key and subkey packets
  * @param  {String} passphrase
- * @return {Boolean} true if all key and subkey packets encrypted successfully
  */
 Key.prototype.encrypt = function(passphrase) {
   if (this.isPrivate()) {
@@ -376,8 +375,6 @@ Key.prototype.encrypt = function(passphrase) {
     for (var i = 0; i < keys.length; i++) {
       keys[i].encrypt(passphrase);
     }
-  } else {
-    throw new Error("Nothing to encrypt in a public key");
   }
 };
 
--- a/src/key.js
+++ b/src/key.js
@@ -370,11 +370,13 @@ Key.prototype.getEncryptionKeyPacket = function() {
  * @param  {String} passphrase
  */
 Key.prototype.encrypt = function(passphrase) {
-  if (this.isPrivate()) {
-    var keys = this.getAllKeyPackets();
-    for (var i = 0; i < keys.length; i++) {
-      keys[i].encrypt(passphrase);
-    }
+  if (!this.isPrivate()) {
+    throw new Error("Nothing to encrypt in a public key");
+  }
+
+  var keys = this.getAllKeyPackets();
+  for (var i = 0; i < keys.length; i++) {
+    keys[i].encrypt(passphrase);
   }
 };
 
--- a/src/packet/secret_key.js
+++ b/src/packet/secret_key.js
@@ -199,7 +199,6 @@ SecretKey.prototype.encrypt = function (passphrase) {
   arr.push(crypto.cfb.normalEncrypt(symmetric, key, cleartext, iv));
 
   this.encrypted = util.concatUint8Array(arr);
-  this.isDecrypted = false;
 };
 
 function produceEncryptionKey(s2k, passphrase, algorithm) {
--- a/src/packet/secret_key.js
+++ b/src/packet/secret_key.js
@@ -199,6 +199,7 @@ SecretKey.prototype.encrypt = function (passphrase) {
   arr.push(crypto.cfb.normalEncrypt(symmetric, key, cleartext, iv));
 
   this.encrypted = util.concatUint8Array(arr);
+  this.isDecrypted = false;
 };
 
 function produceEncryptionKey(s2k, passphrase, algorithm) {
--- a/test/general/key.js
+++ b/test/general/key.js
@@ -716,9 +716,13 @@ var pgp_desktop_priv =
       var armor2 = key.armor();
       expect(armor1).to.equal(armor2);
       expect(key.decrypt('passphrase')).to.be.true;
+      expect(key.primaryKey.isDecrypted).to.be.true;
       expect(key.encrypt('new_passphrase')).to.be.true;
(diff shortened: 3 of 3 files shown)


src/key.js
```diff
@@ -366,6 +366,27 @@ Key.prototype.getEncryptionKeyPacket = function() {
...
+ */
+Key.prototype.encrypt = function(passphrase) {
+  if (this.isPrivate()) {
+    var keys = this.getAllKeyPackets();
+    for (var i = 0; i < keys.length; i++) {
+      try {
+        keys[i].encrypt(passphrase);
+      } catch (e) {
+        return false;
+      }
```
tanx: No need for the boolean return value. If the encrypt work then fine, otherwise just throw up the exception and let the developer figure out what he did wrong. Swallowing exceptions is too dangerous here.

test/general/key.js
```diff
@@ -706,5 +706,24 @@ var pgp_desktop_priv =
...
+    if (openpgp.util.getWebCryptoAll()) { opt.numBits = 2048; } // webkit webcrypto accepts minimum 2048 bit keys
+    openpgp.generateKey(opt).then(function(key) {
+      key = key.key;
+      var armor1 = key.armor();
+      var armor2 = key.armor();
+      expect(armor1).to.equal(armor2);
+      expect(key.decrypt('passphrase')).to.be.true;
+      expect(key.encrypt('new_passphrase')).to.be.true;
+      expect(key.decrypt('passphrase')).to.be.false;
+      expect(key.decrypt('new_passphrase')).to.be.true;
```
tanx: You should be able to test if encryption was successful by checking the `isDecrypted` flag on the key object.

src/key.js
evilaliv3: you are right, i just followed what done in the decrypt function().

maybe it would be good to do the same there?

test/general/key.js
evilaliv3: right suggestion! thanks

src/packet/secret_key.js
```diff
@@ -199,6 +199,7 @@ SecretKey.prototype.encrypt = function (passphrase) {
   arr.push(crypto.cfb.normalEncrypt(symmetric, key, cleartext, iv));
 
   this.encrypted = util.concatUint8Array(arr);
+  this.isDecrypted = false;
 };
```
evilaliv3: @tanx i may be wrong but i've noticed that the encrypt() function is not marking the isDecrypted so that obviously my tests with the additions you suggested were not able to work.

I think i've rightly fixd it in this commit. Could you please recheck?

tanx:
Looks fine to me. Although no idea why the tests are failing.

src/key.js
```diff
@@ -366,6 +366,22 @@ Key.prototype.getEncryptionKeyPacket = function() {
 };
 
 /**
+ * Encrypts all secret key and subkey packets
+ * @param  {String} passphrase
+ * @return {Boolean} true if all key and subkey packets encrypted successfully
```
tanx: No return value anymore

src/key.js
```diff
@@ -366,6 +366,22 @@ Key.prototype.getEncryptionKeyPacket = function() {
...
+ */
+Key.prototype.encrypt = function(passphrase) {
+  if (this.isPrivate()) {
+    var keys = this.getAllKeyPackets();
+    for (var i = 0; i < keys.length; i++) {
+      keys[i].encrypt(passphrase);
+    }
+  } else {
+    throw new Error("Nothing to encrypt in a public key");
+  }
```
tanx: This is just nitpicking. But I usual do validation in the if clause and throw an exception. Then you don't need to indent the actual logic and don't need an else block.

src/packet/secret_key.js
tanx: Looks fine to me.

src/key.js
evilaliv3: right! going to remove this.

src/key.js
evilaliv3: ok for removing the else, but what about the indentation? could you please clarify?

i'm thinking to put simply:
```
  if (this.isPrivate()) {
    var keys = this.getAllKeyPackets();
    for (var i = 0; i < keys.length; i++) {
      keys[i].encrypt(passphrase);
    }
  }
```

evilaliv3:
@tanx unit tests were failing cause i removed the return value for the decrypt function but not aligned the tests. 

now they should work!

src/key.js
tanx: You misunderstand. Keep the error handling but check for !isPrivate() before going to the main code. Then you don't need an else block.

src/packet/secret_key.js
toberndo: @evilaliv3 You set `isDecrypted` to false, but it looks to me like the private MPIs are still in the key after the encrypt. If this is correct, I would rather call `clearPrivateMPIs` after encrypt which also sets `isDecrypted`.

evilaliv3:
alliright! i think we are done with this.

following @toberndo i've used keys[i].clearPrivateMPIs() instead of isDecrypted=false.

in addition i've moved this call on the key.encrpypt instead that in the privatekey.encrypt as i was doing;
the reason is tha by design (@tanx can confirm) all the code is currently implemented for having the key by default unlocked, and having the clearPrivateMPIs() called internally by the privatekey.encrypt will change totally this pattern as the privatekey.encrypt is currenty used in a lot of code;

e.g. simply putting the clearPrivateMPIs() inside secret_key.py:encrypt had me got failure on all this tests:
```
  1) Unit Tests General Packet Writing and encryption of a secret key packet.:
     Error: timeout of 120000ms exceeded. Ensure the done() callback is being called in this test.
  

  2) Unit Tests General Key Preferences of generated key:
     Error: timeout of 120000ms exceeded. Ensure the done() callback is being called in this test.
  

  3) Unit Tests General Key Generated key is not unlocked by default:
     Error: timeout of 120000ms exceeded. Ensure the done() callback is being called in this test.
  

  4) Unit Tests General Key Generate key - single userid:
     Error: Error generating keypair: Cannot read property 'toBigInteger' of undefined
      at onError (dist/openpgp.js:15588:9)
      at lib$es6$promise$$internal$$tryCatch (dist/openpgp.js:3305:16)
      at lib$es6$promise$$internal$$invokeCallback (dist/openpgp.js:3317:17)
      at lib$es6$promise$$internal$$publish (dist/openpgp.js:3288:11)
      at lib$es6$promise$$internal$$publishRejection (dist/openpgp.js:3238:7)
      at lib$es6$promise$asap$$flush [as _onTimeout] (dist/openpgp.js:3063:9)

  5) Unit Tests General Key Generate key - multi userid:
     Error: Error generating keypair: Cannot read property 'toBigInteger' of undefined
      at onError (dist/openpgp.js:15588:9)
      at lib$es6$promise$$internal$$tryCatch (dist/openpgp.js:3305:16)
      at lib$es6$promise$$internal$$invokeCallback (dist/openpgp.js:3317:17)
      at lib$es6$promise$$internal$$publish (dist/openpgp.js:3288:11)
      at lib$es6$promise$$internal$$publishRejection (dist/openpgp.js:3238:7)
      at lib$es6$promise$asap$$flush [as _onTimeout] (dist/openpgp.js:3063:9)

  6) Unit Tests General Key Encrypt key with new passphrase:
     Error: Error generating keypair: Cannot read property 'toBigInteger' of undefined
      at onError (dist/openpgp.js:15588:9)
      at lib$es6$promise$$internal$$tryCatch (dist/openpgp.js:3305:16)
      at lib$es6$promise$$internal$$invokeCallback (dist/openpgp.js:3317:17)
      at lib$es6$promise$$internal$$publish (dist/openpgp.js:3288:11)
      at lib$es6$promise$$internal$$publishRejection (dist/openpgp.js:3238:7)
      at lib$es6$promise$asap$$flush [as _onTimeout] (dist/openpgp.js:3063:9)

  7) Unit Tests General OpenPGP.js public api tests encrypt, decrypt, sign, verify - integration tests GCM mode (native) AES encrypt, decrypt should encrypt and decrypt with two passwords:
     Error: timeout of 120000ms exceeded. Ensure the done() callback is being called in this test.
```

while instead i think that having it in the external API like i'm doing in commit https://github.com/openpgpjs/openpgpjs/pull/446/commits/4e3937f1e25384ef293f8a4c754f45fa0e7408a5 would work better and apply only to real password changes and explicit calls for the user to it.

tanx:
> the reason is tha by design (@tanx can confirm) all the code is currently implemented for having the key by default unlocked

Not sure what you mean here, but generated keys are not unlocked by default:
https://github.com/openpgpjs/openpgpjs/blob/master/src/openpgp.js#L99

Other than that, the PR looks good now from my side 👍 

@toberndo, what do you think?

evilaliv3:
k i was saying that we cannot put the clearPrivateMPIs() in the lowlevel secret_key.encrypt() or it will always lock the key in the current workflow.

I think only the external API i added that should do it.

test/general/key.js
```diff
@@ -706,5 +706,28 @@ var pgp_desktop_priv =
...
+      key.encrypt('new_passphrase');
+      expect(key.primaryKey.isDecrypted).to.be.false;
+      expect(key.decrypt('passphrase')).to.be.false;
+      expect(key.primaryKey.isDecrypted).to.be.false;
+      expect(key.decrypt('new_passphrase')).to.be.true;
+      expect(key.primaryKey.isDecrypted).to.be.true;
+      var armor3 = key.armor();
+      expect(armor3).to.not.equal(armor1);
+      done();
+    }).catch(done);
```
tanx: Btw Mocha now supports returning promises: http://tobyho.com/2015/12/16/mocha-with-promises/

bartbutler:
I feel like not clearing MPIs on key encrypt is not going to be understood by most users and will be considered a bug/security hole. I sort of agree with that interpretation myself--I expect MPIs to be cleared when I encrypt a key.

NSkelsey:
@bartbutler I made that mistake in #443.

Would love to see this merged!

tanx:
@toberndo thanks. Will create a new release shortly

tanx:
Released in v2.3.0

Issue #455, opened 2016-05-14
openpgp overrides Promise in node environment


SimonSchick:
Version: 2.3.0
Node: 4.4.2
Description:
We are using opengpg and I tried to update to the newest version, by doing that our whole integration tests started to fail due to timeouts, I'm still not entirely sure where they time out, due to the complexity of our system, but it's probably in one of our dependencies.

After inspecting the global object and eventually applying `Object.freeze(global)` I figured out openpgp is is apparently overriding/polyfilling promises in a node environment that doesn't require it.
Somewhere in the implementation there seems to be a problem, eventually causing an event loop starvation.

Can you perhaps looking into not polyfilling Promises in node environments that already implement them?

tanx:
Can you please test if the following branch resolves the problem:

https://github.com/openpgpjs/openpgpjs/pull/456

Just git clone and do `npm install && grunt` to build openpgp.js. The just copy the files from `dist/` to your node_modules/openpgpjs/dist/ directory.

SimonSchick:
It's still polyfilling it with your changes.

tanx:
Ok. Thanks for checking. It's probably because our browserify/babel build includes the polyfill and the environment detection is broken under node due to browserify.

@evilaliv3 suggested to leave out polyfills out of the build completely. I guess we will have to go this route even though it will make it harder to use th library on older browsers. Devs will need to include the polyfill in both the main window as well as in the web worker context.

bartbutler:
I feel like leaving out the polyfill is fine for the main window but is going to cause major usability issues for the web worker. I suggest we either figure out why the environment detection is broken (it used to work---we have an older OpenPGP.js build in production with node 0.12 with no issues) or remove the polyfill for the main context only--the the web worker should never been used under node anyway. I'm not certain the web worker _can_ be used under node.

tanx:
@bart which version of OpenPGP.js are you using in production?

I remember also having trouble in my travis build for the new Promise polyfill with PhantomJS v1.x. This is why I kept the version `^1.0.0` of the polyfill for such a long time. That version is ancient but got the job done.

I bumped the es6-promise dependency recently because PhantomJS v2.x could handle the new version. I could revert that change.

bartbutler:
For node we are still using our fork of 1.2.0, basically the v2.0 pull request before you cleaned it up.

tanx:
[...]

PR #456, opened 2016-05-15
Only polyfill ES6 Promises when required

tanx:
I thought the module `es6-promise` already did this check internally. But it seems to cause problems under node.js v4.x.
Resolves openpgpjs/openpgpjs#455


tanx:
Does not solve issue on node.js v4.x

NNemec:
Would it be possible to backport this fix to v1.x ? I seem to have the same problem in my project, but I cannot upgrade to v2 for other reasons.

bartbutler:
Feel free to make a PR and I will merge it. I consider 1.x to be deprecated at this point though.

PR #457, opened 2016-05-16
Fix node promise

tanx:
Resolves openpgpjs/openpgpjs#455 by using our own [fork of the es6-promise polyfill](https://github.com/openpgpjs/es6-promise/compare/master...openpgpjs:fix-node-detection-dist) that does not ovierride V8's native implementation.


bartbutler:
Obsolete I think, closing.

PR #474, opened 2016-06-24
Issue/449

evilaliv3:
This pull request implement the possibility to use the combination of both the web workers + the webcrypto by making possible to enable one or the other capability independently.

the only browser that would not benefit of this is internet explorer that is not able to use the webcrypto api from inside the web worker; for this browser only the webworker will be used making the library less fast but more resilient to "this script is running for to long and i want to fucking kill it; can i?" messages.

the pull implement also the additional test for when both the capabilities are enabled.

@tanx could you please review it and suggest eventual feedbacks?

from a more bigger analysis the improvent impacts not only the key generation but also the encryption/decryption routines where the ascii armor is used as because this functions also are currently runned in the main thread when the webcrypto is used.


src/openpgp.js
```diff
@@ -99,7 +99,7 @@ export function destroyWorker() {
 export function generateKey({ userIds=[], passphrase, numBits=2048, unlocked=false } = {}) {
   const options = formatUserIds({ userIds, passphrase, numBits, unlocked });
 
-  if (!util.getWebCryptoAll() && asyncProxy) { // use web worker if web crypto apis are not supported
+  if (asyncProxy) { // use web worker when available
```
tanx: So this will cause [IE and Edge](https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/13904769-support-web-crypto-api-inside-a-web-worker) to use JS instead of WebCrypto.

I am actually not sure which browsers even support WebCrypto inside a worker thread. See [here](https://lists.w3.org/Archives/Public/public-webcrypto-comments/2013Jul/0012.html).

Have you done benchmarks to validate that this doesn't introduce performance regression?

src/openpgp.js
```diff
@@ -158,7 +158,7 @@ export function decryptKey({ privateKey, passphrase }) {
 export function encrypt({ data, publicKeys, privateKeys, passwords, filename, armor=true }) {
   checkData(data); publicKeys = toArray(publicKeys); privateKeys = toArray(privateKeys); passwords = toArray(passwords);
 
-  if (!nativeAEAD() && asyncProxy) { // use web worker if web crypto apis are not supported
+  if (asyncProxy) { // use web worker when available
```
tanx: This will probably cause a performance regression as well for large files. Especially if transferable objects are not activated.

tanx:
@evilaliv3 I made some comments. I am actually not sure this RP makes things faster:

1. Which browsers actually support WebCrypto inside of Web Workers?
2. Have we done performance comparisons before/after to rule out performance regression, especially for large files e.g. with large files together with AES-GCM?
3. Is the key signing after keygen really so slow that this change is even necessary? Have we done compared before/after?

src/openpgp.js
evilaliv3: not that much, but I looked at the saucelabs results where i see practically no difference.

basically for what we gain in the browser crypto, we loose with the other routines running in the main thread; if you try on a mobile or on IE you always the annoying messages that want to kill the browser and you. one day or an other they may succeed.

src/openpgp.js
evilaliv3: let's enable the transferrable but i repeate my point that i think that resilience and not long running scripts is tha way, but i may be wrong and i'm up to your suggestions.

tanx: > let's enable the transferrable but i repeate my point that i think that resilience and not long running scripts is tha way, but i may be wrong and i'm up to your suggestions.

You can try turning them on by default. The last time I tried it broke the build.

test/unittests.html
```diff
@@ -16,7 +16,7 @@
     <script src="lib/fetch.js"></script>
     <script>
         mocha.setup('bdd');
-        mocha.timeout(240000);
+        mocha.timeout(300000);
```
tanx: These times are the limit for individual test cases. Not total build time. I am aware that the cross browser build on saucelabs is currently not very stable, as some browsers timetimes time out (probably due to memory issues). Increasing this value will only reduce our ability to catch performance regressions :/

src/util.js
```diff
@@ -460,7 +460,12 @@ export default {
       return;
     }
 
-    return typeof window !== 'undefined' && window.crypto && window.crypto.subtle;
+    var self = (typeof window === 'undefined') ? self : window;
```
tanx: I'm not sure if declaring `var self` locally does not override the global `self`. Perhaps call the local variable `_self` just to be sure. Otherwise this looks good.

src/openpgp.js
tanx: > basically for what we gain in the browser crypto, we loose with the other routines running in the main thread;

Have your tried generating a 4096 bit key? That took minutes in JS as far as I can remember. In constrast, the actual signing of keys should be constant time.

Again, I would reiterate that we should to some profiling to understand which parts of the code are the actual bottlenecks, before patching.

Thanks :)

test/unittests.html
evilaliv3: ok if you prefer that, but i think that for what relates the performances regressions we should eventually take care of collecting the stats and already we are somehow doing that with travis/saucelabs.

what do you think?

src/openpgp.js
evilaliv3: you are right, but having a long key generation in a webworker is not an issue for me, while it is having a signing or other stuff blocking the main thread as the application reamins frozen all the time.

src/util.js
evilaliv3: agreed, let's call this _self to be sure

evilaliv3:
@tanx: from my side the patch is ready;

- i've removed the commit raising the timeout limit.
- ive implemented a getGlobal utility where to wrap also 'global' that is defined in Node in place of self;
- i've enabled zero_copy that in my tests seems to work and do not break the build process anymore.

evilaliv3:
I ammend, there is still a fix to be done; on some browsers when one use transferrable objects the object is no more accessible from the main thread;
This means that some tests have to be fixed in order to perform their final check using the original object.

evilaliv3:
here we go @tanx; i've changed some of the unit tests because for each argument passed to the library while using transferrable objects we need to keep a copy for doing the test final checks.

tests performances are comparable, https://travis-ci.org/openpgpjs/openpgpjs/builds/140343628

so that on IE i continue prefer not using the main thread and keep the application working than having a performant applications that prompt to die.

\cc @bartbutler @toberndo

evilaliv3:
one thing that could make us all happy would be to have a variable openpgp.techPreference = 'WebWorker' or 'WebCrypto' where to leave to the user the choiche?

tanx:
Rebased it on the master after merging the updated dependencies ... Waiting for build results.

src/util.js
```diff
@@ -452,10 +452,10 @@ export default {
   },
 
   getGlobal: function() {
-    if (typeof window !== 'undefined') {
-      return window;
-    } else if (typeof self !== 'undefined') {
+    if (typeof self !== 'undefined') {
       return self;
+    } else if (typeof window !== 'undefined') {
+      return window;
```
tanx: I tested this locally. Seems the native web crypto apis were never invoked with your patch since `self.window` is set in the worker. Good thing I benchmarked locally ;)

src/openpgp.js
```diff
@@ -99,7 +99,7 @@ export function destroyWorker() {
 export function generateKey({ userIds=[], passphrase, numBits=2048, unlocked=false } = {}) {
   const options = formatUserIds({ userIds, passphrase, numBits, unlocked });
 
-  if (asyncProxy) { // use web worker when available
+  if (!util.getWebCryptoPrefixed() && asyncProxy) { // use web worker when available, but not legacy browsers
```
tanx: Key gen was unbearably slow in Safari, since the webcrypto apis are not supported in the worker context. This brings back native WebCrypto instead of worker in Safari and IE11.

src/util.js
evilaliv3: ah! on which browser?

src/util.js
```diff
@@ -481,6 +481,30 @@ export default {
...
+      return;
+    }
+
+    var _self = this.getGlobal();
+    if (typeof _self === 'undefined') {
+      return;
+    }
+
+    if (typeof _self.crypto !== 'undefined') {
+      return _self.crypto.webkitSubtle;
```
evilaliv3: is this line correct? i think this will always make return _self.crypto.webkitSubtle whenever self.crypto is defined.

anyway i like this mixed approach just for legacy browsers, 

we should just get sure that on recent EDGE and Safari this does not make call always the old API that maybe is still there.

i would implement a check that the self.crypto is not there before the check for the legacy API.

src/util.js
tanx: On all of them: https://github.com/openpgpjs/openpgpjs/blob/master/src/worker/worker.js#L20

Preferring self over window fixed this.

src/util.js
tanx: You're right. Good catch. I'll fix legacy detection.

evilaliv3: :+1: if you need some help when done i will recheck it locally also both on Firefox, Chrome, and IE.

tanx: Ok. I added a check to make sure the standard api is detected. If it is available `util. getWebCryptoPrefixed()` will return undefined.

https://github.com/openpgpjs/openpgpjs/blob/issue/449/src/util.js#L483-L509

evilaliv3: perfect! this is exactly as i would have changed it.

I'm not that satisfied of the name of the utility, that is more a getNonStandardWebCrypto() but it's only a matter of naming so :+1:

evilaliv3:
@tanx both IE and Safari are failing so i expect something is still not ok with the patch.

I will now retest IE and tell you; maybe is simply a timeout or maybe not.

src/util.js
```diff
@@ -503,7 +503,7 @@ export default {
 
     if (_self.msCrypto) {
       return _self.msCrypto.subtle;
-    } else if (_self.crypto) {
+    } else if (_self.crypto && _self.crypto.webkitSubtle) {
```
evilaliv3: @tanx i've added this && in order to be sure that _self.crypto.webkitSubtle exists before returning it;

this is required supposing that a browser different fomr safary could not offer the standard API but have the _self.crypto object; by the current code that browser was returning blindly webkitSubtle.

Issue #424, opened 2016-03-20
Move HKP client to its own module


tanx:
As part of an effort to make the core library leaner and meaner we should create a new git-repo/npm-module named `openpgp-hkp` that can be used optionally by users of OpenPGP.js.

This would also remove the `fetch` polyfill as a direct dependency of OpenPGP.js.

twiss:
Moved to https://github.com/openpgpjs/hkp-client in 43f0724aeb1d2e86c3b586a3631672856e428459.

PR #479, opened 2016-07-25
functionality to search for keys with hkp

### LINKED ISSUE #477: Feature request: hkp search for keys
It would be nice if one could search for keys. I.e. if one does not know the exact key and needs to display a choice to the user:
e.g.: https://pgp.mit.edu/pks/lookup?search=richard+stallman&op=index

hkp.lookup({query: 'Richard Stallman'}).then(function(array_of_key){
  … ask user which one to use …
})

kellerben:
closes #477


twiss:
The HKP client should probably be moved to a separate package (#424).

Issue #427, opened 2016-03-21
Implement ECC support for Curve P-256, P-384, P-521 (RFC6637)


tanx:
See http://tools.ietf.org/html/rfc6637

This issue tracks the implementation of ECDSA and ECDH for "Curve P-256", "Curve P-384", and "Curve P-521". The advantage of these curves is that they have already been standardized for OpenPGP in RFC 6637 and are supported natively by both [WebCrypto](https://www.chromium.org/blink/webcrypto) and [node.js crypto](https://nodejs.org/api/crypto.html).

fpietrosanti:
I feel that those should be given precedence compared to Ed25519 of #428 because of wider WebCrypto support and together with https://github.com/openpgpjs/openpgpjs/issues/427 could lead to a fully native implementation, that's imho a major benefit.

fpietrosanti:
Any progress or effort estimation regarding this very specific implementation?

tanx:
> Any progress or effort estimation regarding this very specific implementation?

Nope. OTF has not responded to my application to the [Core Infrastructure Fund](https://www.opentech.fund/requests/core-infrastructure-fund).

I'd be up to implement this though if anyone has budget for it.

prusnak:
What budget are we talking about here? We could do a crowdfunding via Bitcoin for this.

tanx:
> What budget are we talking about here?

For ECC support including NIST Curves and Ed25119, I'd estimate 15-20k EUR. I'm usually on the pessimistic side when it comes to estimating cost though. But whatever is left, could be used to fund other features on the [v3.0 roadmap](https://github.com/openpgpjs/openpgpjs/milestones/v3.0).

> We could do a crowdfunding via Bitcoin for this. I am willing to manage it, if you don't have time for administration.

It would be great if you could manage a crowdfunding!

evilaliv3:
Interesting patch @ismaelbej

ive to surely find the time to test a bit this implementation and to evaluate the possible integration.

could you please elaborate on the status of the implementation and on the possible issues if present?
the code is all written added is all authored by you or is copied from other libraries?
i see that the unit tests are not passing; are you able to make them work before we start digging into the review?

thank you so much!

ismaelbej:
Hi @evilaliv3 

[...]

PR #484, opened 2016-08-18
Issue/427

ismaelbej:
This is an implementation of RFC 6637 to partially fix issue #427.

A few remarks of the code:
- Uses [Elliptic](https://github.com/indutny/elliptic/) for curve cryptography, but should be possible to add others backends like webcrypto or node's crypto.
- It uses [AES library](https://github.com/cryptocoinjs/aes)/ to implement key wrapping. I've failed to find a suitable AES decrypt function for RFC 3394 in the current code. But should be easy to replace.

It was tested against GnuPG version v2.1.8 + libgcrypt v1.7.0-beta262, some other details  are in the pull request [README-ECC.md](https://github.com/Jaxx-io/openpgpjs/blob/issue/427/README-ECC.md).


extra/bitcoin_interop.js
```diff
@@ -0,0 +1,66 @@
+var openpgp = require('openpgp');
```
evilaliv3: This whole bitcoin example is nice but i would keep it out of the patch. We will document it in a different way.

Let's evaluate the removal before finishing the integration of the patch

ismaelbej:
Removed 'const' from unit test and dropped bitcoin example.

src/packet/public_key.js
```diff
@@ -100,8 +102,11 @@ PublicKey.prototype.read = function (bytes) {
     var p = 0;
 
     for (var i = 0; i < mpicount && p < bmpi.length; i++) {
-
-      this.mpi[i] = new type_mpi();
+      if (this.algorithm === 'ecdsa' && i === 0) {
```
ismaelbej: Here is the issue related to the assumption that key packets only contain MPI. Also for 'ecdh' I've to add a similar condition.

src/packet/public_key.js
```diff
@@ -100,8 +104,13 @@ PublicKey.prototype.read = function (bytes) {
     var p = 0;
 
     for (var i = 0; i < mpicount && p < bmpi.length; i++) {
-
-      this.mpi[i] = new type_mpi();
+      if ((this.algorithm === 'ecdsa' || this.algorithm === 'ecdh') && i === 0) {
```
ismaelbej: Here is the issue related to the assumption that key packets only contain MPI. In order to support Ed25519 also a similar change have to be made.

src/openpgp.js
```diff
@@ -374,6 +374,12 @@ function checkCleartextMessage(message) {
  * Format user ids for internal use.
  */
 function formatUserIds(options) {
+  if (!options.curve) {
```
ismaelbej: This is a work around because some tests do not expect extra parameters like 'curve' and 'material'.

src/key.js
```diff
@@ -968,13 +973,25 @@ export function generate(options) {
...
+    }
+    var material;
+    if (typeof(options.material) !== "undefined") {
+      material = options.material.subkey;
+    }
+    secretSubkeyPacket.algorithm = enums.read(enums.publicKey, subkeyType);
+    return secretSubkeyPacket.generate(options.numBits, options.curve, material);
   }
 
   function wrapKeyObject() {
```
ismaelbej: Here wrapKeyObject has to be modified to remove hash/encryption algorithms that should not be used with a given curve. For example a p521 key should not be used with sha1, but with sha256. Also prefer_hash_algorithm need to reflect that also.

ismaelbej:
Replying to @sanjanarajan who asked why did I choose an external AES library. In order to support ECDH We've to implement key wrapping algorithm from RFC 3394 which uses AES encrypt and AES decrypt functions.

At the time the repository didn't have AES decrypt only AES encrypt was present (https://github.com/Jaxx-io/openpgpjs/blob/issue/427/src/crypto/cipher/aes.js), so we use the fast and light implementation from https://github.com/cryptocoinjs/aes. Before sending our pull request v2.0 was released with the new addition of asmcrypt-lite I did try to use AES decrypt from there but was unsuccessful. So we send our pull request with the annotation that the additional dependency can be removed if AES decrypt was made available to use.

evilaliv3:
@sanjanarajan i'm noticing that you are try to move forward the ECC implementation.

that is great!  i'm interested to help retesting this and eventually study with you the possible adoption of WebCrypto as backend. are you already evaluating this possible backend alternative in your current implementation?

sanjanarajan:
@evilaliv3 Yes, my current plan is to use a WebCrypto backend for the curves for which it is available, fall back to a Node Crypto backend when available, and use the elliptic backend as the last option. I have also done some refactoring to address the structure/style issues detailed in the previous comments. I am hoping to make a new pull request in the next week, looking forward to your feedback!

bartbutler:
Closing as another version of this PR has just been merged!

PR #490, opened 2016-09-30
Fix message signing with multiple keys #451

mavxg:
Fixes issue #451 where signing with multiple keys results in all the keys being used for OnePassSignature packets but only the last key is used multiple times to actually sign the message.


evilaliv3:
Thanks @mavxg

I would find valuable to implement a test that prove the failure and that your solution is solving the issue.

This way we will be protected from the regression.

What do you think?

You could implement a test case that uses 2 keys and verify that both are present in the signature, if not give a failure.
This way before your patch the test will be proven to fail and and after your patch it will have to succeed.

This way we could proceed integrating the fix.

bartbutler:
Thanks, and sorry for the long delay. This issue seems to have been inadvertently fixed with the 3.0 code refactor. I have added a test to make sure it does not reoccur.

PR #509, opened 2017-01-09
V3 header generation was concat()-ing non-Uint8 arrays

alax:
V3 (old style) header generation throws an error in its current form because it specifies non-Uint8 arrays.


bartbutler:
Thanks!

PR #510, opened 2017-01-11
Only use required files in bower main array

Denfie:
There was also a pull request from iDuuck at 22.02.2016. His idea was right. Just put main files to the "main" at your bower.js file.
If some guys need the minify version, they can go directly to the bower_components folder.
The main section defines only the main needed files not the minified version of the main files.
Please put the changes to your newest version :D
Thanks a lot.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "openpgp",
-  "version": "2.3.5",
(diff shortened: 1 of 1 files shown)


bartbutler:
You are absolutely right, thanks! Fixed in the master branch.

Issue #87, opened 2013-07-08
Compress messages by default


toberndo:
openpgp.write_encrypted_message currently does not compress the message before encryption. This should be changed.

RFC 4880 says:

```
OpenPGP implementations SHOULD compress the message after applying
the signature but before encryption.
If an implementation does not implement compression, its authors should
be aware that most OpenPGP messages in the world are compressed.
...
Furthermore, compression has the added side effect that some types of 
attacks can be thwarted by the fact that slightly altered, compressed
data rarely uncompresses without severe errors.
```

Attacks on non-compressed messages: http://www.cs.umd.edu/~jkatz/papers/pgp-attack.pdf
##

tanx:
Is this issue still relevant?

laktak:
Compression by default would be nice. Also see https://github.com/openpgpjs/openpgpjs/pull/272#issuecomment-75221078

bartbutler:
It would be. Last time I checked it was ridiculously slow, but we can
probably look for new ASM or other fast libraries to use for it just as we
did for AES.

laktak:
> Last time I checked it was ridiculously slow

Did you check after #272?

bartbutler:
I think I did. As another comment in that thread says, it doesn't make much sense to use JS when node and browser-native implementations exist.

I doubt compression is an issue for messages, it just doesn't make much difference in that case. For files I remember it being awful, though it should be revisited.

chesnokovilya:
I think this issue  should be closed by PR  #637

PR #513, opened 2017-01-17
Compression support for the high-level API

barhun:
This PR adds compression support to the high-level  **_encrypt_** function. The feature was what I needed today, and I thought others might find it useful because some people had also been requesting compression for outgoing messages as mentioned in pull request #272.

By default, the _**encrypt**_ function still generates uncompressed messages; so the new API maintains backwards compatibility with this new feature added. However, if it comes better to enable compression with no extra parameters as suggested in issue #87; the _**compress**_ property of the parameter object can be made optional with a default value set to true. I am still not sure whether it was the right way to go but the algorithm to use is taken from the value of _**openpgp.config.compression**_. Do you think this is good or it would make a better way to allow parameter passing to override the configuration value for another compression algorithm?

Note: _**bzip2**_ is yet to be implemented by low-level objects. _**zip**_ and _**zlib**_ appear to be working as expected with decent performance overhead.


barhun:
There appears to be some integration errors that cause browser tests to fail. Could you please check them?

bartbutler:
Thanks, finally got this feature merged in the mainline branch, sorry it took so long!

PR #516, opened 2017-01-26
Generate valid version 3 signatures

alax:
Currently, OpenPGP has code to write version 3 (old-style) signatures, but it doesn’t follow the RFC at all. This pull request fixes that, and generates valid version 3 signatures.


alax:
I'd love to get this merged, so please let me know if there is anything you need from me.

bartbutler:
Nope, just needed to look it through myself. Yeah, the old code was totally wrong. Merging.

PR #522, opened 2017-02-18
Readme - add warning about config.aead_protect potentially breaking compatibility

danenania:
It took me awhile to realize this was what was causing my compatibility issues, so I figured a louder warning could be helpful.


bartbutler:
Thanks!

PR #541, opened 2017-04-06
Simplify code of CRC24 checksum in armor encoding

BAKFR:


bartbutler:
Huh, I wonder why the original version of the code is so bloated. The only thing I can think of is some sort of performance hack, but this really does not look like the old version would be any more performant than yours...

Issue #524, opened 2017-02-28
How to avoid slow build time during development?


piec:
Hi,
First thank you for this cool project.

I'm trying some changes in the source of openpgpjs (src/*) but I find the build times quite unpleasant especially when I try small modifications. The task `grunt browserify:openpgp_debug` takes more than 10s to execute on my not-so-old PC (core i7 860).
```
$ time node_modules/.bin/grunt browserify:openpgp_debug
Running "browserify:openpgp_debug" (browserify) task
>> Bundle dist/openpgp_debug.js created.
Done.
13.38s user 0.31s system 116% cpu 11.796 total
```
Most of the time is spent in the user process which means the CPU is the blocking factor (not the disk).
I'm pretty sure the babel pass is responsible. How do you guys develop? Do you have a trick to make the build faster?

I've had a quick look at browserify/babelify/babel documentations but I found no way to enable a cache or to make the build parallel. A way to cache the output of babel if the .js hasn't changed would be nice since most of the times changes are located in only a few files - right now every file seem to be re-babeled at every build. babel/babel-loader seems to do that with `cacheDirectory` but this is for webpack.

----
Things if tried but don't help:
* browserify --fast doesn't help
* grunt watch
* I also tried invoking browserify manually with the same result so grunt isn't the cause

Cheers

bartbutler:
Hmm, 10s does sound about right to me. To be honest, with appropriate editor plugins/linters I don't find the build time to be an issue, as I don't rebuild that often usually. I see your point though, and would entertain any suggestions you come up with to speed it up.

darkyen:
How about a watch mode with a cache of pre-built files? This is out of the box supported by watchify https://github.com/substack/watchify

piec:
I just did a quick test with watchify and it seems to work great!
(not integrated in the gruntfile right now)
```
var fs = require('fs');
var browserify = require('browserify');
var watchify = require('watchify');

var b = browserify('src/index.js', {
    standalone: 'openpgp',
    cache: {},
    packageCache: {},
    debug: true,
})
.external('crypto')
.external('buffer')
.external('node-localstorage')
.external('node-fetch')
.transform('babelify', {
    presets: ['es2015'],
    ignore: /.*\.min\.js/
})
.plugin('browserify-derequire')
.plugin('watchify')

b.on('update', bundle);
bundle();

function bundle() {
  console.log('bundle ...')
[...]

PR #549, opened 2017-05-05
Add grunt 'concurrent' target, can be used with --watch to watchify

piec:
Related to #524 

I discovered that grunt-browserify can use watchify instead of browserify just by using an option (see: https://github.com/jmreidy/grunt-browserify#watch) so the changes are quite simple

Watchify is blocking so I used grunt-concurrent in order to run 1 watchify instance per target (openpgp, openpgp_debug, worker)

Usage:

    $ grunt concurrent --watch

Leave running
In other shell:

    $ touch src/message.js
    $ touch src/worker/worker.js

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -37,6 +37,10 @@ module.exports = function(grunt) {
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
     browserify: {
+      options: grunt.option('watch') ? {
+            watch: true,
+            keepAlive: true,
+          } : {},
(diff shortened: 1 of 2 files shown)


piec:
Hmm don't understand the failures, are they timeouts?

bartbutler:
Hmmm, not sure. Sorry for the long radio silence, but could you look at the current head and see if this is still needed? I might be, there are still some issues with 'grunt browsertest --watch'

piec:
OK I'll take a look

Issue #554, opened 2017-05-31
Reduce memory usage when using with nodejs


neophob:
Hi

I use openpgp.js to generate keys and enable encrypted communication on a embedded device. I did a heap dump of my application and saw that the compile openpgp node module uses 1.7mb of memory (just to load the module). this is quite beefy and I wonder if there are some easy ways to reduce the module size for openpgp.

One easy way to reduce the needed memory would be, when package.json reference the minimized openpgp file (which is there but not used). 

More info about needed memory (v8 heapdump):
- current version: `(string), 1.5mb (shallow)`
- minimized version: `(string), 0.3mb (shallow)`

neophob:
see PR https://github.com/openpgpjs/openpgpjs/pull/558

bartbutler:
Completely agree with you, but npm doesn't have a good way (or any way) to specify the minified version and if we use 'main' to do it it'll break webpack and other integrations. We might be able to publish a minified version to npm as a separate package, maybe that would work...not sure if there is a better solution.

neophob:
thanks @bartbutler - that make totally sense. I agree, a second deployed package (openpgpjs-min?) would be a solution.

as an alternative, a github only version might also be feasible - but i think then you would need a second repo with might be even worse.

what do you think?

twiss:
Fixed by #1047.

PR #558, opened 2017-07-02
Fix for #554 

neophob:


bartbutler:
I think this will break webpack and other integrations, unfortunately.

Issue #527, opened 2017-03-08
work with multiple binding signatures per subkey


tomholub:
I may do a pull request for this one day but it will take me forever to familiarize myself with the code base and establish a work/test flow.

This key:
http://pgp.surfnet.nl:11371/pks/lookup?op=get&search=0x2930100100003344

Has a bunch of of subkeys. The subkeys have 1-3 binding signatures each. Most of them expired, some not.

```
sub  2048R/D3F47501 2010-09-13            
sig sbind  00003344 2010-09-13 __________ 2012-09-12 []

sub  2048R/A81ED221 2012-09-12            
sig sbind  00003344 2012-09-12 __________ 2014-09-12 []

sub  1024D/20D0AAAC 2010-09-13            
sig sbind  00003344 2010-09-13 __________ 2012-09-12 []

sub  1024D/8B985EF9 2012-09-12            
sig sbind  00003344 2012-09-12 __________ 2014-09-12 []

sub  4096R/D22DB38E 2014-08-27            
sig sbind  00003344 2014-08-27 __________ 2016-08-26 []
sig sbind  00003344 2016-08-24 __________ 2017-02-20 []
sig sbind  00003344 2017-03-06 __________ 2017-09-02 []

sub  4096R/92736F3D 2014-08-27            
sig sbind  00003344 2014-08-27 __________ 2016-08-26 []
sig sbind  00003344 2016-08-24 __________ 2017-02-20 []
sig sbind  00003344 2017-03-06 __________ 2017-09-02 []
```

The non-expired signature is always the last and OpenPGP.js ignores it. GPG is happy with it.

A backward compatible fix would be to add `subKey.bindingSignatures = []` next to `subKey.bindingSignature`. Then fill it up like this:

```js
          case _enums2.default.signature.subkey_binding:
            if (!subKey) {
              _util2.default.print_debug('Dropping subkey binding signature without preceding subkey packet');
              continue;
            }
            subKey.bindingSignatures.push(packetlist[i]);
            subKey.bindingSignature = packetlist[i];
            break;
```

Finally,verify it with something like this:
```js
SubKey.prototype.verify = function (primaryKey) {
  ...
  // check V4 expiration time
  if (this.subKey.version === 4) {
    for(i in this.bindingSignatures) {
      if(this.bindingSignatures[i] ... is valid) {
        return _enums2.default.keyStatus.valid;       
      }
    }
    return _enums2.default.keyStatus.expired;
  }
[...]

PR #565, opened 2017-07-22
work with multiple binding signatures per subkey #527

tomholub:
This one was bothering me for quite some time now. The PGP spec allows for more than one binding signature per subkey, but OpenPGP.js will only pick one of them to process.

There may be 5 expired binding signatures and one valid, and OpenPGP.js will most likely think the subkey is invalid.

Also, if you read that armored key and later armor it, it would be missing a bunch of binding signatures if there was more than one in the original armored key (at which point other software can't use it either - not good).

Some people have keys like that, and when they do, others had to fall back to GPG to encrypt for them.

Here's a fix.


bartbutler:
Looks great. As soon as the tests complete I'll merge it.

tomholub:
I like the turnaround time, nice.

Issue #497, opened 2016-11-11
Overflow in four-octet numbers


toberndo:
Time fields are represented by [unsigned four-octet number](https://tools.ietf.org/html/rfc4880#section-3.5).

There seems to be an overflow for values larger than 30 bit:

```
util.readNumber(util.writeNumber(Math.pow(2, 30), 4))
=> 1073741824
util.readNumber(util.writeNumber(Math.pow(2, 31), 4))
=> -2147483648
```

I came across this when implementing the key expiration time in https://github.com/openpgpjs/openpgpjs/pull/458. With the current implementation of write/readNumber the key expiration time is limited to 2080 (which is surely enough).

toberndo:
BTW, sorry for not having the time to port https://github.com/openpgpjs/openpgpjs/pull/458 to OpenPGP.js v2. I hope to get Mailvelope on OpenPGP.js v2 soon.

tomholub:
@bartbutler we should close this

PR #566, opened 2017-07-22
util.readNumber: stop overflowing until full range of uint32 for 4 byte numbers | #497

tomholub:
This allows key expiration to get properly parsed up to full range of pgp spec.

The original error was because bitwise operations in JS operate on int32, so max number is < than uint32.

Added a test, too.


bartbutler:
Good catch!

PR #567, opened 2017-07-27
Streaming encryption (again) WIP

kellym:
_tldr;_ This is a WIP for streaming encryption but I wanted to open a PR for posterity's sake and for potential discussion.

So, this was previously approached in these two PRs: #321 and #306.  This is a WIP but I wanted to revisit this since there are occasionally people like me who require streaming PGP.

This PR doesn't include the changes from #321, but it does add a couple of new items, namely _actual_ streaming where the body length is not known, armor, and message signing.  

There are a couple of issues I've run into as well. One is a weird quirk of having to supply an extra packet for modification detection to remain valid when signing a message (I used the marker packet since it is essentially ignored). I've also not dealt with DSA signing. I've left most of the `cipher.js` file alone from what @hellais had done, other than some cleanup.

I'd also be fine with creating a separate package for this entirely, as mentioned in #321, but there are a few mods required in this master package to support updating hashes and/or supplying hashes to methods.  There's also some redundancy that could pretty easily be solved.


kellym:
Well, this was working before rebasing to v2. Working on fixing now.

bartbutler:
@kellym Really happy you are working on this. It was going to be the next thing we looked into spending some time on. Looking forward to reviewing it!

kellym:
Well, the commits are a mess :) but tests are passing locally for node 0.12.x, 4, 5, and 6 -- I've got node 0.12.18 running locally with no failing tests (not sure if the failed test in Travis is legit or not). 

It still needs some cleanup and optimization, and I'm sure some of the code could be better reused.

I'd also like to change the API calls, too. Open to ideas.

kellym:
Well, I can't seem to get things sorted out re `buffer` and either browsers or node 0.12.x. It either works in a really old version of node, or it works in the browsers, but it seems like there's an issue with how node 0.12.x deals with requires that's causing some complications.

bartbutler:
We can probably safely drop support for very old node versions at this point I think. I do not feel like we are compelled to support old node the same way we are for older browsers, given that most of the node usage will be server-side or in another situation where it's likely to be upgraded somewhat regularly.

kellym:
Ah, then that changes things :). I'll update accordingly. Thanks for the feedback.

kellym:
I saw some old, what looks like possibly unused, code re compression in this repo.  Since I wanted compression and node supports streaming compression pretty effortlessly, I added it to this PR. 

I don't have any tests for it yet (that I've committed to the repo), but I have a test locally that generates an encrypted, compressed file, and I have used the built-in `gpg` to verify it works. It didn't look like there were any compression/decompression tests already.

bartbutler:
No, I don't think compression got a lot of attention because the existing implementation had some performance problems in the browser. I think adding it is a great idea.

drakkhen:
Thank you for working on this, @kellym !

kellym:
I seem to be going in circles re this Buffer issue. I can't seem to get around browserify to get at the native node `buffer` module.  `require('buffer')` just returns the browser package, and `global.Buffer` has been replaced. 

I tried aliasing the `buffer` package to `_buffer`, which works for the openpgp code, but browserify still attempts to include `buffer` because the Stream classes need it.

Out of good ideas at the moment.

kellym:
That last commit fixed an issue of encoding when piping in node, but for some reason I'm running into another issue of some sort of corruption when the data to encrypt exceeds a certain number of bytes. For 512-byte chunks, it always gives out at 5619 bytes. Seems higher at 1024-byte chunks. Doing some digging.

kellym:
Yeah, seems that the max is `11n - 13`, where `n` is the chunk size. Not sure where that number is coming from yet.

kellym:
It seems that the issue is in the CipherFeedbackStream, which is from the previous PR. I can call it once with any size buffer and it works. If I call it more than once, I'm limited to `11n - 13`. I haven't touched that file much other than to update `Buffer` references.

kellym:
Resolved the issue I was having with the streaming CFB. I'm guessing there's probably a way to make use of node's crypto library to do CFB, although some quick tests failed with it.

kellym:
That last commit adds the ability to use the new `openpgp.encrypt()` syntax with a stream, i.e.
```javascript
// by passing a stream as the data
const pgp = openpgp.encrypt({ data: stream, ... });
pgp.pipe(stdout);

// or by not passing data at all
const pgp = openpgp.encrypt({ /* data is not defined */ ... });
stream.pipe(pgp).pipe(stdout);
```

tomholub:
Thanks for working on this @kellym !

Do you have a list of what needs to be done before this can be tested in a semi-production setting?

kellym:
We're using it in our (dare I say) production environment right now, but it's only working for Node environments. I just can't seem to get past this issue with it using the non-native `Buffer` class because the browserified `buffer` package overrides the global.  I got it semi-working at one point, but then the browserified `stream` package still seemed to be forcing its way in, which ended up using the wrong `Buffer` class.

So, three main points still blocking this:
1. Issues respecting the usage of native node modules
2. No streaming DSA -- only RSA has been implemented (shouldn't be a huge thing to add)
3. Better tests for each individual piece. Right now it just does everything all as one big test.

Regarding # 1 again, you _can_ get all the tests to pass for browsers and Node, but when it's actually implemented in Node, you'll find that things like streaming to HTTP fail because it doesn't recognize the `Buffer` being sent to the HTTP stream.

evilaliv3:
I may be available to support this with some retesting if this may help you move it forward.

tomholub:
We will also be testing it in a pure V8 engine (no browser/no node), if at all possible. After that, more tests in browser environment (early 2018).

kellym:
I think I finally resolved the issues with `buffer` between node and the browser. It feels like a hack, but I found absolutely no other way to get it working properly: https://github.com/openpgpjs/openpgpjs/pull/567/commits/b731b00690111b0175f30e853da4c4d0b3b0fe20#diff-34a6d62af0cf0b784f8444529f3130efR574. I suppose an alternative would be to use grunt to replace some placeholder with `buffer`?

bartbutler:
That's nuts, but works I guess. Maybe using a variable works, depending on how thorough browserify is.

kellym:
Yeah, so everything works locally but seems that Travis doesn't like the `buffer/` reference. I'm not sure why, but I'll look into it.

justinmchase:
What about decrypt into a stream?

eg:
```
const pgp = openpgp.decrypt({ data: stream, ... }).pipe(stdout)
```

kellym:
Decryption isn't yet covered in this PR. Also not yet complete, streaming armor has limited headers.

ntallapa12:
Any timeline on when this is going to be merged?

Vineetsai:
@kellym : Noticed that performance with Stream Encryption is very Slow, I tried with running 1.2g file and take about 15min. 

Also Noticed that src/stream/chunked using block size:  this.blockSize = 1024 max, I though that may be the buffer size for encryption block you are using,  increased to 200MB but no help.

Is there any way to improve on performance?, Any suggession, in my test I'm using fileInputStream with buffer size 200MB.

kellym:
Yep it’s super slow. One of the problems is that it can’t block the event
loop in all its calculations or else there’s not a huge point in streaming.
That seems to add a pretty significant amount of time. An alternative would
possibly be to spin up a child process or use threading but both require
significantly more changes from what I could tell.

Vineetsai:
@kellym  Is there way to safely increase the block size, as noticed it set with blockSize=16
c.blockSize = c.prototype.blockSize = 16;
Is there any specific reason for this to be set such low?

Changing this break whole lot things, even it fail to decrypt private key, which I able to tweak to work. but for some reason after encryption resulted encrypted message is seems corrupted. As noticed that encrypted message size goes up, which seems it write out extra byte by updated blockSize, for blockSize 32, it right out extra 16 byte, which look dumping full block at the end of the message.

Tried decrypting resulted message using gpg tool, which give:
gpg --decrypt plain_enc.txt                                                                                                                                                      
gpg: encrypted with 4096-bit RSA key, ID 50BCADAFF0B8140D, created 2018-05-11
      "vineet <vineet@test.com>"
gpg: [don't know]: invalid packet (ctb=2a)
gpg: WARNING: encrypted message has been manipulated!
gpg: [don't know]: invalid packet (ctb=65)

PR #574, opened 2017-08-19
Fix for issue 573: Don't throw with unknown Armor Headers

pbrunschwig:
Don't throw with unknown Armor Headers but just with invalid headers


bartbutler:
Thanks. Made a couple minor edits, the main problem currently is that a test is failing. Can you fix it?

pbrunschwig:
I could not run the TC, but I think my fix should be OK anyhow

bartbutler:
Released as 2.5.9

PR #514, opened 2017-01-24
implement keyExpirationTime

dpjanes:
This implements keyExpirationTime when generating a new key. 

This is more or less a 2.x port of https://github.com/openpgpjs/openpgpjs/pull/458/files

See also: https://github.com/openpgpjs/openpgpjs/issues/511

PR #576, opened 2017-08-30
Set keyExpirationTime for subkey binding signature

toberndo:
Currently when generating a key the keyExpirationTime is only added to the user id self signature but not to the subkey binding signature. Hence those subkeys are not displayed with an expiration time in PGP clients.
Keys generated with GPG set the expiration time also for subkeys.
This was already included in https://github.com/openpgpjs/openpgpjs/pull/458 but omitted in https://github.com/openpgpjs/openpgpjs/pull/514


bartbutler:
Great, merged.

Issue #581, opened 2017-09-15
Failing tests on master due to expired keys


kellym:
The test that just recently got added in #576 is causing tests on master to fail for me locally and also in #567. All keys are registering as expired.

kellym:
Verified via `gpg` on the command line. I'll try to generate a few more unless someone can get to it faster than I can.
![image](https://user-images.githubusercontent.com/283371/30492292-a1d38f68-9a05-11e7-98ef-701448c0f124.png)

kellym:
The purpose of the test that is failing is to select the valid key from a list of expired and valid keys. Looks like when that PR was merged, there was one valid key (expired 2017-09-02), and now there are none.

bartbutler:
That makes sense. Does the replacement key you generated have an invalid subkey? I'm guessing no, as GPG would probably not allow that?

kellym:
No, not invalid beyond being expired. There may be a way to manipulate a key to invalidate it, but I'm not entirely sure how to manipulate a single subkey without corrupting the entire key block. It seems that the previous test had all valid subkeys as well (I imported them locally, and they were all valid but expired).

bartbutler:
Sorry, I meant expired. Does the replacement key have an expired subkey?

kellym:
Yep, one expired encryption subkey, one valid signing subkey, and one valid encryption subkey (the test fails when only the valid signing subkey is present).

bartbutler:
Sounds good, I'll merge it.

PR #582, opened 2017-09-15
Replace expired subkeys with three other subkeys,

kellym:
One is an expired key for encryption, one is a non-expiring key for
only signatures, and one is a non-expiring key for encryption. Fixes https://github.com/openpgpjs/openpgpjs/issues/581.


kellym:
For reference, these are the new keys as compared to the ones shown in that issue: 
image

PR #593, opened 2017-10-11
Add a console error when numBits param too low

laurinenas:


bartbutler:
See comment on issue please.

PR #594, opened 2017-10-18
* [bug] fixed openpgpjs/openpgpjs#451 :message signing with multiple keys

snowyu:
fixed issue #451. 

Also see PR #490 and unit test added.


bartbutler:
Thanks, and sorry for the long delay. This issue seems to have been inadvertently fixed with the 3.0 code refactor, and a test has been added to make sure it does not reoccur.

Issue #590, opened 2017-10-04
You shouldn't rethrow a different error


justinmchase:
I'm currently experiencing a null reference exception in this library but I'm having a hard time debugging it because the error that is thrown has its callstack nerfed.
```
Error: Error decrypting message: Cannot read property 'isDecrypted' of null
    at onError (.../node_modules/openpgp/dist/openpgp.js:16111:9)
    at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

[This](https://github.com/openpgpjs/openpgpjs/blob/master/src/openpgp.js#L550) line strikes me as an anti-pattern. It should just be:
```js
error.data = message;
throw error;
```

Or if you wanted to get more clever you could create a custom Error type that preserves the callstack of the inner exception and modifies message like you have it here. [Here](https://stackoverflow.com/a/17891099/12958) is an example of doing that. Which may look like this:
```js
function PgpError(message, innerError) {
    var temp = Error.apply(this, arguments);
    temp.name = this.name = 'PgpError';
    this.message = temp.message;
    this.innerError = innerError;
    var stack = temp.stack + (innerError ? '\n\n' + innerError.stack : '')
    if(Object.defineProperty) {
        // getter for more optimizy goodness
        /*this.stack = */Object.defineProperty(this, 'stack', { 
            get: function() {
                return stack
            },
            configurable: true // so you can change it if you want
        })
    } else {
        this.stack = stack
    }
}
//inherit prototype using ECMAScript 5 (IE 9+)
PgpError.prototype = Object.create(Error.prototype, {
    constructor: {
        value: MyError,
        writable: true,
        configurable: true
    }
});
```

After altering my code in my node_modules folder to throw the original error, this is the actual callstack I'm getting:
```
TypeError: Cannot read property 'isDecrypted' of null
    at Message.decryptSessionKey (.../node_modules/openpgp/dist/openpgp.js:15003:26)
    at .../node_modules/openpgp/dist/openpgp.js:14949:38
    at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

Which is [this line](https://github.com/openpgpjs/openpgpjs/blob/master/src/message.js#L157) unminifed:
```js
var privateKeyPacket = privateKey.getKeyPacket(encryptionKeyIds);
if (!privateKeyPacket.isDecrypted) {
  throw new Error('Private key is not decrypted.');
}
```

`privateKeyPacket` is null. I don't know how that is possible yet.

[...]

PR #604, opened 2017-11-24
Fix/590 preserve error stack

justinmchase:
Adds innerError to re-thrown Error for slightly better debugging experience.

Based on this issue:
https://github.com/openpgpjs/openpgpjs/issues/590

I found that there wasn't a need to concat the stacks if you just embedded the error onto a property on the thrown error. It prints out find and it also preserves any other fields that may have been on that error beyond just the stack and message. It looks like this when printed out:
capture

### DIFF
--- a/src/openpgp.js
+++ b/src/openpgp.js
@@ -552,7 +552,6 @@ function onError(message, error) {
   if (config.debug) { console.error(error.stack); }
   // rethrow new high level error for api users
   const newError = new Error(message + ': ' + error.message);
-  newError.stack += '\n' + error.stack;
   newError.innerError = error;
   throw newError;
 }
--- a/src/openpgp.js
+++ b/src/openpgp.js
@@ -551,7 +551,10 @@ function onError(message, error) {
   // log the stack trace
   if (config.debug) { console.error(error.stack); }
   // rethrow new high level error for api users
-  throw new Error(message + ': ' + error.message);
+  const newError = new Error(message + ': ' + error.message);
+  newError.stack += '\n' + error.stack;
+  newError.innerError = error;
+  throw newError;
 }
 
 /**
--- a/src/openpgp.js
+++ b/src/openpgp.js
@@ -553,6 +553,7 @@ function onError(message, error) {
   // rethrow new high level error for api users
   const newError = new Error(message + ': ' + error.message);
   newError.innerError = error;
+  newError.stack += '\n' + error.stack;
   throw newError;
 }
 
--- a/test/general/openpgp.js
+++ b/test/general/openpgp.js
@@ -1202,6 +1202,19 @@ describe('OpenPGP.js public api tests', function() {
       });
 
       describe('Errors', function() {
+        it('Errors stack should contain the stack of innerError', function(done) {
+          openpgp.encrypt({
+            data: new Uint8Array([0x01, 0x01, 0x01]),
+            passwords: null
+          })
+          .then(function() {
+            done(new Error('Error expected.'));
+          })
+          .catch(function(error) {
+            expect(error.stack).to.match(/\nError: No keys or passwords/);
+            done();
+          });
+        });
         it('Errors should contain innerError', function(done) {
           openpgp.encrypt({
             data: new Uint8Array([0x01, 0x01, 0x01]),
--- a/test/general/openpgp.js
+++ b/test/general/openpgp.js
@@ -1207,7 +1207,9 @@ describe('OpenPGP.js public api tests', function() {
             data: new Uint8Array([0x01, 0x01, 0x01]),
             passwords: null
           })
-          .then(() => done(new Error('Error expected.')))
+          .then(function () {
+            done(new Error('Error expected.'));
+          })
           .catch(function(error){
             expect(error.innerError).to.exist;
             done();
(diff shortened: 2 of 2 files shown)


tomholub:
I'd be interested to see what it looks like when the error is left completely unhandled. Is the original stack also visible in the console?

tomholub:
It also doesn't seem to like the new `() => ...` syntax in tests for some reason https://travis-ci.org/openpgpjs/openpgpjs/jobs/306947382#L554

justinmchase:
Oh ok, I'll do a regular function. I must be using a newer version of node and I thought it was down compiling it.

Without catching the promise it looks like this:
capture

No stack at all.

justinmchase:
Also, should I include the files in the dist folder as part of the PR? Or do you build that later as part of a release?

bartbutler:
Thanks! Please remove the dist files, yes.

justinmchase:
Travis is failing due to not being able to connect to Sauce Labs I think, is that something my PR is causing or is it some other kind of issue?

```
Tested http://127.0.0.1:3000/test/unittests.html
Platform: OS X 10.10,chrome,46.0
Passed: false
Url https://saucelabs.com/jobs/edff89908ef749e9a5766db8d8840a1e
>> All tests completed with status false
>> 
Warning: Task "saucelabs-mocha:all" failed. Use --force to continue.
Aborted due to warnings.
```

I'm not sure how to resolve.

tomholub:
Bart has a way to re-run the tests. It does seem like it's failing due to something out of your control.

bartbutler:
Yes, I'll check it out. Thanks!

tomholub:
@justinmchase check this out in a browser (I've tested Chrome). The original stack will not be printed when the error is thrown. I'm in favor of appending the original stack trace unless I missed something. See:
```js
var e = new Error('user facing message');
e.innerError = new Error('original error');
throw e;
VM273:1 Uncaught Error: user facing message
    at <anonymous>:1:9
(anonymous) @ VM188:1
```

tomholub:
Eg same test with stack appending (Chrome):

```js
var e = new Error('user facing message');
e.stack += '\n' + (new Error('original error').stack);
throw e;
VM311:1 Uncaught Error: user facing message
    at <anonymous>:1:9
Error: original error
    at <anonymous>:1:20
```

justinmchase:
@tomholub That's interesting, the browser recognizes its an error and does formatting that omits custom fields.

I think I could do both, concat stack and add the innerError object? The reasoning there is just that sometimes other properties get added to an object or the error has a different Type than Error and you may want to preserve that. It seems to not hurt anything in the browser, but I agree that it is nicer to concat the stacks for the browser.

tomholub:
Looks good to me, though I think it's not necessary to log the stack above with console.log anymore. This yellow line: https://github.com/openpgpjs/openpgpjs/pull/604/commits/a5a913419b88c9b3449507f42e1f90c8af9ece1f#diff-a8c60b3c1a1f245471b2188f214b13d9L552

Wait for @bartbutler to confirm, maybe it had some other purpose as well.

bartbutler:
We can remove the console log but it doesn't have to be part of this PR. Looks good, merging.

bartbutler:
Turns out the saucelabs issue was a combination of something screwy with the tests (which we are looking at now) and the tests failing with the web worker (these tests do not run in node). They fail because we can't actually pass a complete Error object through the structured clone algorithm back to the main context. So for now we are settling for appending the stack.

https://github.com/openpgpjs/openpgpjs/commit/49695ff50cfc9743415ced71609f03ba136db774

bartbutler:
The appending of the stack was problematic because the stack format doesn't appear to be at all standard over different browsers. Also, the message is not included in the stack in all browsers, which makes stack-based tests really difficult. I simplified things and just passed the original stack through instead.

https://github.com/openpgpjs/openpgpjs/commit/7a9c3b0ddd1201815af79271f90abf9e8dccbb77

Let me know if you see any problems with this solution.

tomholub:
I was under impression that this would produce stacks only within the openpgp library scope, which would leave the API user clueless as to which lines in their own code caused this to happen in the first place. 

I need to do some tests though, I'm not 100% sure this is the case.

tomholub:
Checked your code in 7a9c3b0 - you're actually re-throwing the original Error, just with modified message. Looks good to me!

bartbutler:
This is true, but the original PR did not do that either. We could concat in another place to make that happen but it might screw up debugging tools. We can also do something like:

`error.originalStack = stack;`

So you'd have stack and originalStack properties. Does that work?

tomholub:
Giving it a second look, I think it will all work nicely when you re-throw it. I'll test it though to double check.

bartbutler:
I don't think it'll work through the asyncProxy without concatenation:

https://github.com/openpgpjs/openpgpjs/blob/master/src/worker/async_proxy.js#L69

I think I need to change that replacement to concatenation to get the main context stack.

tomholub:
Any resources on how to run tests that include the web worker?

bartbutler:
I've been sort of faking it by fooling the saucelabs tests into running locally, but it's hacky. I think that's one of the things that needs some work.

tomholub:
Let me make a separate issue for that, we can discuss it there

Issue #602, opened 2017-11-22
better user error when supplying encrypted key to reformatKey


tomholub:
I was doing something like this
```js
  openpgp.generateKey({userIds: [{ name:'Jon Smith', email:'jon@example.com' }], numBits: 1024, passphrase: 'super long and hard to guess secret'}).then(generated => {
    openpgp.reformatKey({privateKey: generated.key, passphrase: 'super long and hard to guess secret', userIds: [{name: 'New Smith', email: 'new@example.com'}]}).then(new => {
      console.log(new);
    });
  });
```
And getting **Uncaught (in promise) Error: Error reformatting keypair: Cannot read property 'toBigInteger' of undefined**

Turns out the signing method didn't have access to all of the MPIs because the key was still encrypted.

I suggest that we check this early, eg in `reformat` or `wrapKeyObject` and throw user-friendly error: **Cannot sign: Key not decrypted** or something of that sort.

bartbutler:
Yes, lots of common errors could benefit from being more friendly. Feel free to take a crack at it.

PR #606, opened 2017-11-25
reformatKey: Key not decrypted error + test | #602

tomholub:
reformatKey: Key not decrypted error + test | #602


bartbutler:
Thanks!

PR #611, opened 2017-12-02
Option to pass in custom session key in top-level encrypt function

sanjanarajan:


src/message.js
```diff
@@ -213,21 +213,31 @@ Message.prototype.getText = function() {
...
       throw new Error('No keys or passwords');
     }
 
-    let sessionKey = crypto.generateSessionKey(enums.read(enums.symmetric, symAlgo));
-    msg = encryptSessionKey(sessionKey, enums.read(enums.symmetric, symAlgo), keys, passwords);
+    if (sessionKey) {
+      if (!util.isUint8Array(sessionKey.data) || !util.isString(sessionKey.algorithm)) {
+        throw new Error('Invalid session key for encryption.');
+      }
+      symAlgo = sessionKey.algorithm;
```
bartbutler: Shouldn't this be:

`symAlgo = enums.read(enums.symmetric, sessionKey.algorithm);`

src/message.js
```diff
@@ -238,12 +248,18 @@ Message.prototype.encrypt = function(keys, passwords) {
...
 
   }).then(() => {
     msg.packets.push(symEncryptedPacket);
     symEncryptedPacket.packets = new packet.List(); // remove packets after encryption
-    return msg;
+    return {
+      message: msg,
+      sessionKey: {
+        data: sessionKey,
+        algorithm: symAlgo
```
bartbutler: This is now an enum. I think you should change it back to being a string, personally.

src/openpgp.js
```diff
@@ -177,6 +177,7 @@ export function decryptKey({ privateKey, passphrase }) {
  * @param  {Key|Array<Key>} publicKeys        (optional) array of keys or single key, used to encrypt the message
  * @param  {Key|Array<Key>} privateKeys       (optional) private keys for signing. If omitted message will not be signed
  * @param  {String|Array<String>} passwords   (optional) array of passwords or a single password to encrypt the message
+ * @param  {Object} sessionKey                (optional) session key in the form: { data:Uint8Array, algorithm:String }
  * @param  {String} filename                  (optional) a filename for the literal data packet
  * @param  {Boolean} armor                    (optional) if the return values should be ascii armored or the message/signature objects
  * @param  {Boolean} detached                 (optional) if the signature should be detached (if true, signature will be added to returned object)
```
bartbutler: add returnSessionKey here?

sanjanarajan: done

PR #612, opened 2017-12-03
Remove 'user strict' and use 'let' and 'const'

BafS:
- Use let and const ES6 features. Some parts need to be refactored so I let some `var` (for eg. in packetlist.js).
- 'use strict' is unnecessary inside modules because module code is always strict mode code. Ref: https://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code

If this PR is merged I can continue to modernise OpenPGPjs :)


bartbutler:
Hi @BafS,

Sorry for the additional effort, but can you rebase on the current master? The huge ECC branch was just merged and broke your PR :(

sanjanarajan:
Closing this because @BafS is going to start this from scratch on a new branch :)

Issue #616, opened 2017-12-31
skip unsupported algos in pubkey


tomholub:
The following public key has an ECC algo as one of the subkeys, but also has a standard RSA subkey available. 

This will be solved as a side-product of #580 #428 #427 

Still, ideally, OpenPGP.js should skip unsupported algos and use the ones it knows how to use.

```
-----BEGIN PGP PUBLIC KEY BLOCK-----

mQINBFmF3poBEACjmnBUOd79vOc6sywdw7d85ItYQDeSh5WhTguXkt84+8C+oLXY
bWelVmyjtZ+vOsZSXb44tE2zNCWouNK/wcE16JmmIV3VVn9I/EpMTK/iFx3e/SGk
6C56uopgsQFlDjbYRZ6bPc+ImsEdfaj3DyYrWJO3dCEBEEOv4/u7UZ/ulBJOEW4p
CJjgk7esmT9LmG0AvSSUZhZQqqMo5wvrfz07uDbJ6MFjPXgBIThtJPjKSpTXCaaC
s2GJZVkKagR1rexytIgsEU3OYHAl+qYfHqcFqa1xbrdEjC8RDFIU0vJc3aGpRufG
EVlnDw/LfWN6qX4NE9IC/QyKFXEQOybP9UEl9ti/dE6FUSHT+g/M4zCdSL1RFMkz
91/UnjgrZAboMoKgBZqYLpawjYIyVMYTdIerDQYiymCdD+Ifkgc3FwgTsfSFnFJX
wLL2hqJGoZSz6jHKKiDDKathLQqWEvAbCo6FP5VsqILi2Tu+6IWOeLp07YXzDiDN
R7/yqDGOCavYmwN3mt5TwdNeOd22LFWHLBcsDc5gIB2EkJoMpemG1TJi60FX9P2W
I8K7ICBmFzwnvH9i2xBln+4+YDHyod7clC1I0wE2BbO6ttNaUR/6Xczm8gjx0zzK
svOZZDTXjXtO8ss41j7Rg0TS3C4+QWMcmHCJCiu8tD4GS5du/V5xw8TJ8QARAQAB
tBIweEFDRTU2OTcxRDU1NURBQTWJAlcEEwEIAEECGwEFCQGfZkMFCwkIBwIGFQgJ
CgsCBBYCAwECHgECF4AWIQR5EMRfifyMwsvSSrCs5Wlx1VXapQUCWYYAQgIZAQAK
[...]

Issue #580, opened 2017-09-14
ECC supported ?


ThierryIT:
Hi,
When is the possibility to use ECC/PGP will be available with Openpgpjs ?
I am using Mailvelope who is depending on OpenPGPjs.
Thx

bartbutler:
Hi Thierry,

This is being actively developed and we hope released within the next couple months.

tanx:
![image](https://user-images.githubusercontent.com/1374174/31190625-e7b5d2c6-a93b-11e7-8bb9-631b85ffc30e.png)

rahiparikh:
@bartbutler That's awesome! Looking forward to it. I came across https://safecurves.cr.yp.to/ and thought it might help you guys when making decisions on curves.

bartbutler:
Thanks @rahiparikh!

Issue #428, opened 2016-03-21
Implement ECC support for Curve Ed25519 (IETF draft)


tanx:
See https://tools.ietf.org/html/draft-koch-eddsa-for-openpgp-00

This issue tracks the implementation of EdDSA for Curve Ed25519. Unlike the NIST curves, the IETF has not standardized this as an RFC. Also, there is no native support in the WebCrypto api yet. In spite of this [GnuPG v2.1 uses Curve25519 as the default](https://www.gnupg.org/faq/whats-new-in-2.1.html#ecc). So we should add support to stay compatible and could use a JS implementation until native support exists in both WebCrypto and node.js.

Available JS implementations:
https://github.com/joebandenburg/libaxolotl-crypto-curve25519
https://github.com/indutny/elliptic
https://github.com/google/end-to-end/tree/master/src/javascript/crypto/e2e/ecc

tanx:
elliptic seems to be the most popular implementation, looks well tested, and it's being actively maintained by @indutny:

https://github.com/indutny/elliptic

Regarding performance, there are some [benchmarks](https://github.com/indutny/elliptic/blob/master/README.md#benchmarks). Although it does not seem to use typed arrays, where as this asm.js compiled library does:

https://github.com/joebandenburg/libaxolotl-crypto-curve25519

In any case good quality code, tests, and active maintenance are probably more important.

rmhrisk:
On browser can use WebCrypto, I made a server side WebCrypto also, it is done in TypeScript which compiles down to ES5 with Babel - https://github.com/PeculiarVentures/node-webcrypto-ossl

tanx:
Unfortunately we can use WebCrypto only for NIST curves right now. See https://github.com/openpgpjs/openpgpjs/issues/427.

I like the idea of using a single api for node and the browser, but as mentioned on Twitter I'd prefer to use a wrapper for the node crypto api than a custom OpenSSL build. Node.js has a huge community relying on it's release process with timely security updates.

indutny:
As a maintainer of elliptic, I would like to say thank you guys for considering it. Please let me know if I could be any help to you.

rmhrisk:
it is not a custom OpenSSL build it is a binding to openssl. That said I understand the desire to use something with a staffed security process around it.

rmhrisk:
If you wanted to do a JS polyfill for webcrypto based on @indutny's great work I would base it on - https://github.com/vibornoff/webcrypto-shim

[...]

PR #618, opened 2018-01-04
Adds ECC Support for Ed25519/Curve25519 and NIST Curves

### LINKED ISSUE #26: Integrate fast ECC support
See http://eprint.iacr.org/2011/654

mahrud:
Happy new year to all!

This pull request finishes compatibility with elliptic curve cryptography keys. Specifically, the supported [curves](https://wiki.gnupg.org/ECC) are P-256, P-384, P-521, SECP256K1, and Ed25519/Curve25519. Closes #26 #427 #428 #580 #616 .

When available, key generation and signing/verifying with P-256, P-384, and P-521 curves are relegated to webCrypto or nodeCrypto. SECP256K1 can also be relegated to nodeCrypto, but currently, it is NOT.

As a bonus, it also adds a way to perform browser testing and debugging locally, using the command `grunt browsertest`.

Mahrud


tanx:
Happy new year indeed :)

tanx:
@mahrud thank you for this contribution. Very impressive work! I've only briefly skimmed over the PR... but ESlint, unit tests, and travis build is green. I'm speechless.

The only thing I saw at first sight... please remove the built files from `dist/` and rebase. Those assets will be generated by the npm maintainer on the next release.

I hope @bartbutler and the ProtonMail team have time to review this soon. I'm unfortunately swamped right now.

bartbutler:
Yep, @sanjanarajan and I will start the review today/this weekend. @mahrud is ProtonMail's very talented winter intern :)

tanx:
@bartbutler cool. Well you make sure you don't let go of him :) @mahrud welcome to the small and humble email crypto community!

tomholub:
Thanks for the contribution @mahrud !

bartbutler:
Yeah, in the end it was a group effort by @ismaelbej, the original branch author, @sanjanarajan, and @mahrud which made this happen. Looking forward to getting it merged!

mahrud:
@tanx, @tomholub thanks! As Bart mentioned, most of the work had already been done in the ecc_nist branch. I only shuffled sign/verify-related functions around to use async/await-style Promises and added Curve25519/Ed25519 support, more on that later.

> The only thing I saw at first sight... please remove the built files from dist/ and rebase.
Done.

A couple of notes I remembered:
- I'm currently using a patched version of the @indutny's [`elliptic`](https://github.com/openpgpjs/elliptic) library for Curve25519/Ed25519. There is at least one issue (https://github.com/indutny/elliptic/issues/142) that I couldn't solve, but I don't know if it's a problem for PGP or not.
- Also, key wrapping and unwrapping ECC keys, which is done in `src/crypto/aes_kw.js`, seem to be awfully slow.
- From Travis [tests](https://saucelabs.com/beta/tests/c7922df44f49486d9e19eb1ed4810eeb/metadata), it looks like older browsers aren't just failing a few tests, but for some reason they're not running the tests at all.

tanx:
Well congrats to the whole team! This feature request has been open since 2011.

src/crypto/public_key/elliptic/curves.js
```diff
@@ -27,107 +27,210 @@
...
+    // TODO add more here
+  };
+}
 
 const curves = {
   p256: {
     oid: util.bin2str([0x2A, 0x86, 0x48, 0xCE, 0x3D, 0x03, 0x01, 0x07]),
-    curveName: 'P-256',
-    hashName: 'SHA-256',
+    keyType: enums.publicKey.ecdsa,
```
sanjanarajan: I don't really understand how these key types work. Why does this curve have type ecdsa if it can also be used for ecdh? It seems like ecdsa and ecdh are treated the same way in the curve generation function anyway.

src/crypto/public_key/elliptic/curves.js
```diff
@@ -27,107 +27,210 @@
...
+
+
+//////////////////////////
+//                      //
+//   Helper functions   //
+//                      //
+//////////////////////////
+
+
+async function webGenKeyPair(name, algorithm) {
```
sanjanarajan: As far as I can tell, the only value for algorithm that you ever pass in is ECDSA, and I believe that webcrypto's generateKey function produces identical results for name = ECDH and name = ECDSA as long as you make the key extractable. So is this ECDH handling code necessary?

src/crypto/public_key/elliptic/curves.js
mahrud: It's not to differentiate between ECDSA and ECDH, but ECDSA and EdDSA. I named it keyType because it matches the higher level keyType in the key packet (as opposed to subKeyType, which would be used to differentiate between ECDH and some other DH algorithm)

src/crypto/public_key/elliptic/curves.js
mahrud: I assume you're referring to `algorithm`, since `name` just gets the curve name. It's not necessary. I left it there mostly as a reminder of possible values for the `algorithm` field in `generateKey`, which are `deriveKey/deriveBits` or `sign/verify`. I couldn't find a nice and clean way of using `deriveKey` in `ecdsa.js`, but if you can we should use it, and in that case, this `algorithm` input could be used.

sanjanarajan: Then for simplicity's sake I think we should get rid of the algorithm parameter and the separate ECDH handling altogether rather than keeping this dead code around (which will probably confuse someone else at some point too).

src/crypto/public_key/elliptic/curves.js
sanjanarajan: So you basically use "ecdsa" to mean "ecdsa or ecdh" and "ecdh" to mean "only ecdh"? Instead could we just simplify this to have key types EC and Ed? It doesn't look like you ever use the ecdsa vs ecdh distinction, even though I see that curve25519 has the ecdh key type.

src/crypto/public_key/elliptic/ecdsa.js
```diff
@@ -25,25 +25,52 @@
...
+import base64 from '../../../encoding/base64.js';
+
+const webCrypto = util.getWebCrypto();
+const webCurves = curves.webCurves;
+const nodeCrypto = util.getNodeCrypto();
+const nodeCurves = curves.nodeCurves;
+
+var ECDSASignature = ASN1.define('ECDSASignature', function() {
+  this.seq().obj(
+    this.key('r').int(),  // FIXME int or BN?
```
sanjanarajan: Can you explain the FIXME here?

src/encoding/base64.js
```diff
@@ -18,15 +18,17 @@
 'use strict';
 
 var b64s = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/';
+var b64u = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_';
 
 /**
  * Convert binary array to radix-64
  * @param {Uint8Array} t Uint8Array to convert
```
sanjanarajan: Could you add documentation here for the other input parameters?

src/encoding/base64.js
```diff
@@ -87,19 +91,20 @@ function s2r(t, o) {
  * @returns {Uint8Array} binary array version of input string
```
sanjanarajan: Need documentation for input parameter u

src/crypto/public_key/elliptic/ecdsa.js
```diff
@@ -25,25 +25,52 @@
...
  * @param  {Uint8Array}  m          Message to sign
-* @param  {BigInteger}  d          Private key used to sign
+ * @param  {BigInteger}  d          Private key used to sign
  * @return {{r: BigInteger, s: BigInteger}}  Signature of the message
  */
-function sign(oid, hash_algo, m, d) {
+async function sign(oid, hash_algo, m, d) {
+  var signature;
   const curve = curves.get(oid);
+  hash_algo = hash_algo ? hash_algo : curve.hash;
```
sanjanarajan: So, when signing a message, it turns out that hash_algo just comes directly from the config's preferred hash algorithm (always SHA256). So it may make more sense here to default to the curve's recommended hash.

src/crypto/public_key/elliptic/ecdsa.js
```diff
@@ -53,15 +80,172 @@ function sign(oid, hash_algo, m, d) {
...
  * @param  {Uint8Array}  m          Message to verify
- * @param  {BigInteger}  Q         Public key used to verify the message
+ * @param  {BigInteger}  Q          Public key used to verify the message
+ * @return {Boolean}
  */
-function verify(oid, hash_algo, signature, m, Q) {
+async function verify(oid, hash_algo, signature, m, Q) {
+  var result;
   const curve = curves.get(oid);
+  hash_algo = hash_algo ? hash_algo : curve.hash;  // FIXME is this according to the RFC?
```
sanjanarajan: See line 62 comment

src/key.js
```diff
@@ -1320,6 +1351,7 @@ export function getPreferredSymAlgo(keys) {
     });
   });
   var prefAlgo = {prio: 0, algo: config.encryption_cipher};
+  // FIXME
```
sanjanarajan: What needs to be fixed?

src/message.js
```diff
@@ -126,56 +124,64 @@ Message.prototype.decrypt = function(privateKey, sessionKey, password) {
...
-
-  if (password) {
-    var symEncryptedSessionKeyPacketlist = this.packets.filterByTag(enums.packet.symEncryptedSessionKey);
-    var symLength = symEncryptedSessionKeyPacketlist.length;
-    for (var i = 0; i < symLength; i++) {
-      keyPacket = symEncryptedSessionKeyPacketlist[i];
-      try {
-        keyPacket.decrypt(password);
-        break;
+  var keyPacket, results, error;
```
sanjanarajan: What's going on with this error variable? It doesn't look like it ever gets used

src/message.js
```diff
@@ -126,56 +124,64 @@ Message.prototype.decrypt = function(privateKey, sessionKey, password) {
...
+    if (password) {
+      var symESKeyPacketlist = this.packets.filterByTag(enums.packet.symEncryptedSessionKey);
+      if (!symESKeyPacketlist) {
+        throw new Error('No symmetrically encrypted session key packet found.');
       }
-      catch(err) {
-        if (i === (symLength - 1)) {
-          throw err;
+      // FIXME need a circuit breaker here
+      results = await Promise.all(symESKeyPacketlist.map(async function(packet) {
```
sanjanarajan: I think the "circuit breaker" that you want can just be achieved by using a for loop here instead of a promise all. In general I think this can be applied to other parts of your code (sacrificing the parallelization for fast-failing and code simplicity).

src/crypto/public_key/elliptic/eddsa.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param  {enums.hash}  hash_algo  Hash algorithm used in the signature
+ * @param  {{r: BigInteger, s: BigInteger}}  signature  Signature to verify
+ * @param  {Uint8Array}  m          Message to verify
+ * @param  {BigInteger}  Q          Public key used to verify the message
+ * @return {Boolean}
+ */
+async function verify(oid, hash_algo, signature, m, Q) {
+  var result;
+  const curve = curves.get(oid);
+  hash_algo = hash_algo ? hash_algo : curve.hash;  // FIXME is this according to the RFC?
```
sanjanarajan: See comment on ecdsa.js

src/crypto/public_key/elliptic/eddsa.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param  {String}      oid        Elliptic curve for the key
+ * @param  {enums.hash}  hash_algo  Hash algorithm used to sign
+ * @param  {Uint8Array}  m          Message to sign
+ * @param  {BigInteger}  d          Private key used to sign
+ * @return {{r: BigInteger, s: BigInteger}}  Signature of the message
+ */
+async function sign(oid, hash_algo, m, d) {
+  var signature;
+  const curve = curves.get(oid);
+  hash_algo = hash_algo ? hash_algo : curve.hash;
```
sanjanarajan: See comment on ecdsa.js

src/key.js
```diff
@@ -567,44 +569,47 @@ Key.prototype.update = function(key) {
...
-      if (srcUser.userId && (srcUser.userId.userid === that.users[i].userId.userid) ||
-          srcUser.userAttribute && (srcUser.userAttribute.equals(that.users[i].userAttribute))) {
-        that.users[i].update(srcUser, that.primaryKey);
+    await Promise.all(that.users.map(async function(dstUser) {
+      if ((srcUser.userId && (srcUser.userId.userid === dstUser.userId.userid)) ||
+          (srcUser.userAttribute && (srcUser.userAttribute.equals(dstUser.userAttribute)))) {
+        await dstUser.update(srcUser, that.primaryKey);
         found = true;
-        break;
+//        break;
```
sanjanarajan: If you are going the Promise.all route here, you should remove this commented out break. Otherwise, you have the option of doing this sequentially rather than with a Promise.all.

src/worker/worker.js
```diff
@@ -15,11 +15,11 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
-/* globals self: true */
+/* eslint-disable no-restricted-globals */
 
 self.window = {}; // to make UMD bundles work
 
-importScripts('openpgp.js');
+importScripts('openpgp_browser.js'); // FIXME
```
sanjanarajan: What is this?

src/crypto/public_key/elliptic/curves.js
```diff
@@ -27,107 +27,210 @@
...
   p256: {
     oid: util.bin2str([0x2A, 0x86, 0x48, 0xCE, 0x3D, 0x03, 0x01, 0x07]),
-    curveName: 'P-256',
-    hashName: 'SHA-256',
+    keyType: enums.publicKey.ecdsa,
     hash: enums.hash.sha256,
     cipher: enums.symmetric.aes128,
-    nist: true
+    node: nodeCurves.secp256r1,
+    web: webCurves.secp256r1,
```
sanjanarajan: What is going on with these node and web values (for all of the NIST curves)? None of them seem to correspond to the names that node/web crypto use for the curves or the values in webCurves/nodeCurves.

src/crypto/public_key/elliptic/ecdsa.js
```diff
@@ -25,25 +25,52 @@
...
+  hash_algo = hash_algo ? hash_algo : curve.hash;
   const key = curve.keyFromPrivate(d.toByteArray());
-  const signature = key.sign(m, hash_algo);
+  if (webCrypto && config.use_native && curve.web) {
+    signature = await webSign(curve, hash_algo, m, key.keyPair);
+  } else if (nodeCrypto && config.use_native && curve.node) {
+    signature = await nodeSign(curve, hash_algo, m, key.keyPair);
+  } else {
+    signature = await key.sign(m, hash_algo);
+  }
```
ismaelbej: One of the intentions while splitting KeyPair and Curve was to later use the Facade pattern. So you do not have to use lots of `if`s, the `curve.get()` and `curve.keyFromPrivate()` will instantiate the appropiate objects NodeCurve, WebcryptoCurve, PlainJSCurve, NodeKeyPair, WebCryptKeyPair, etc.

src/crypto/public_key/elliptic/ecdsa.js
mahrud: At the time I needed to confirm that the `asn1.js` package handes big integers correctly, which it does, so we can remove this.

src/message.js
mahrud: ESLint would not allow this: https://eslint.org/docs/rules/no-await-in-loop. We can silence this, of course, but I do think it's worth it overall and it doesn't make the code that much more complicated. Also, I believe a "circuit breaker" for promises is on the way.

src/worker/worker.js
mahrud: See line 90 of `Gruntfile.js`. In particular, certain external parts (compare line 53 with 98) are required for running on Node but would break testing on browsers altogether, which is why I had to separate the builds. `test/unittests.html` loads `openpgp_browser.js` instead.

src/crypto/public_key/elliptic/curves.js
mahrud: Ahh, this is definitely a bug. I'll fix it in the next commit. Thanks!

src/key.js
mahrud: Will do. I stared at it for 10 minutes to remember why I didn't just remove it, but can't think of anything (and I would have left a FIXME somewhere if there was anything to consider anyway).

src/crypto/public_key/elliptic/curves.js
mahrud: Will do.

src/crypto/public_key/elliptic/ecdsa.js
mahrud: I don't understand this comment; this already defaults to curve's hash when `hash_algo` isn't defined. Are you saying we should disregard `hash_algo` and always use the curve's default hash? I'd assume this would break things.

src/key.js
mahrud: The only thing that comes to mind is this: https://eslint.org/docs/rules/guard-for-in. I'll remove the FIXME.

src/crypto/public_key/elliptic/ecdsa.js
sanjanarajan: So, if you trace back to the upper-level functions, the hash algorithm is set here: https://github.com/openpgpjs/openpgpjs/blob/master/src/message.js#L371 - always to config.prefer_hash_algorithm which is set by default to SHA256. This made sense when we were just signing with RSA keys, but since different curves have different recommended hash algorithms I think we probably should use those instead.

src/message.js
sanjanarajan: I had a chat with Bart and some of our Javascript frontend guys, and we came to the conclusion that a for loop would be cleaner and actually more of a standard in these cases (when you would break out of the loop after success/failures). We can just add /* eslint no-await-in-loop: "off" */ here to silence these particular cases rather than changing our eslint config altogether.

src/crypto/public_key/elliptic/ecdsa.js
sanjanarajan: Although now that I think about it, we should be getting this value from the key if available rather than relying on the config.

src/encoding/base64.js
```diff
@@ -18,15 +18,17 @@
...
+var b64u = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_';
 
 /**
  * Convert binary array to radix-64
  * @param {Uint8Array} t Uint8Array to convert
  * @returns {string} radix-64 version of input string
  * @static
  */
-function s2r(t, o) {
+function s2r(t, o, u) {
```
sanjanarajan: If o is not being used, let's get rid of it

src/encoding/base64.js
```diff
@@ -18,15 +18,17 @@
...
   // TODO check btoa alternative
+  var b64 = (u === "base64url") ? b64u : b64s;
```
sanjanarajan: Can we make u a boolean instead of a string?

src/crypto/public_key/elliptic/ecdsa.js
sanjanarajan: @mahrud we can stuff webSign and nodeSign into the key.Sign function to achieve this

mahrud: Ahh, now I understand. Ok, I'll incorporate this by moving the fallback pattern into `key.js`.

src/key.js
```diff
@@ -1132,20 +1149,27 @@ export function generate(options) {
   return Promise.resolve().then(() => {
 
     if (options.curve) {
-      options.keyType = options.keyType || enums.publicKey.ecdsa;
+      if (options.curve === 'ed25519' || options.curve === 'curve25519') {
```
sanjanarajan: Let's validate the value of curve here (using enums?)

src/key.js
```diff
@@ -1163,14 +1187,18 @@ export function generate(options) {
 
   function generateSecretKey() {
     secretKeyPacket = new packet.SecretKey();
+    secretKeyPacket.packets = null;
     secretKeyPacket.algorithm = enums.read(enums.publicKey, options.keyType);
-    return secretKeyPacket.generate(options.numBits, options.curve);
+    var curve = options.curve === 'curve25519' ? 'ed25519' : options.curve;
```
sanjanarajan: enums

src/key.js
```diff
@@ -1163,14 +1187,18 @@ export function generate(options) {
...
+    return secretKeyPacket.generate(options.numBits, curve);
   }
 
   function generateSecretSubkey() {
     secretSubkeyPacket = new packet.SecretSubkey();
+    secretKeyPacket.packets = null;
     secretSubkeyPacket.algorithm = enums.read(enums.publicKey, options.subkeyType);
-    return secretSubkeyPacket.generate(options.numBits, options.curve);
+    var curve = options.curve === 'ed25519' ? 'curve25519' : options.curve;
```
sanjanarajan: enums

src/crypto/public_key/elliptic/curves.js
mahrud: Would it be acceptable to set the keyType of curve25519 to also be ecdsa? Technically speaking, it can be used for ECDSA. Just asking because this is probably the easiest fix.

sanjanarajan: Yeah, that works

mahrud:
@waza123 perhaps you would like to mention what platform you're using?

PR #621, opened 2018-01-18
Add flag to enable/disable allowing message signature verification with expired keys

sanjanarajan:
The default is set to true, because this can be useful for verifying message signatures that were created when the key was not expired.

Also removed a function getValidSelfCertificate that was not used anywhere and is anyway made redundant by the certificate-choosing code in getPrimaryUser.


mahrud:
I understand the rationale that one should never sign with an invalid key, but I think verifying the validity of the key should be done later in `sign`, not in `getSigningKeyPacket`. It's probably okay to skip expired keys in `getSigningKeyPacket`, but verifying the certificates whenever we're grabbing the signing key is really inefficient.

Also, there are 77 instances of this in the tests folder that would need to be changed.

PR #629, opened 2018-02-01
move from http to https where possible.

dkg:
There are dozens of links in the OpenPGP.js codebase that are http but
could be replaced with https links.  I've converted as many of them as
i could find.


tomholub:
Thanks, have you checked that all modified links are still working after the change?

dkg:
yes, i've checked.  you'll note that some http links remain -- those
were the ones that didn't work with https. (though some of them don't
even work with http either any more, sadly)

PR #635, opened 2018-02-09
Allow decryption with multiple keys/passwords

sanjanarajan:
This is helpful when using the wildcard Key ID option in public key encrypted session keys (https://tools.ietf.org/html/rfc4880#section-5.1) which rather than indicating the key ID that the session key was encrypted with, has a wild card value of '0' in which case all available private keys should be tried.

Some changes to upper level functions:
1) privateKey -> privateKeys in decrypt and decryptSessionKeys. Now takes either a Key object or an array of them as input.

2) password -> passwords in decrypt and decryptSessionKeys. Now takes either a single string or an array of them as input.

3) sessionKey -> sessionKeys in decrypt. Now takes either a single session key object or an array of them as input.

4) useWildcard flag in encrypt and encryptSessionKey, defaults to false. Will add the wild card key ID of '0' to all public key encrypted session keys rather than the actual key ID.

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -91,13 +91,13 @@ Message.prototype.getSigningKeyIds = function() {
 
 /**
  * Decrypt the message. Either a private key, a session key, or a password must be specified.
- * @param  {Key} privateKey      (optional) private key with decrypted secret data
- * @param  {Object} sessionKey   (optional) session key in the form: { data:Uint8Array, algorithm:String }
- * @param  {String} password     (optional) password used to decrypt
+ * @param  {Array<Key>} privateKeys     (optional) private key with decrypted secret data
+ * @param  {Array<String>} passwords    (optional) password used to decrypt
+ * @param  {Object} sessionKey          (optional) session key in the form: { data:Uint8Array, algorithm:String }
  * @return {Message}             new message with decrypted content
  */
-Message.prototype.decrypt = async function(privateKey, sessionKey, password) {
-  let keyObjs = sessionKey || await this.decryptSessionKeys(privateKey, password);
+Message.prototype.decrypt = async function(privateKeys, passwords, sessionKey) {
+  let keyObjs = sessionKey || await this.decryptSessionKeys(privateKeys, passwords);
   if (!util.isArray(keyObjs)) {
     keyObjs = [keyObjs];
   }
@@ -140,15 +140,15 @@ Message.prototype.decrypt = async function(privateKey, sessionKey, password) {
 };
 
 /**
- * Decrypt an encrypted session key either with a private key or a password.
- * @param  {Key} privateKey    (optional) private key with decrypted secret data
- * @param  {String} password   (optional) password used to decrypt
+ * Decrypt an encrypted session key either with private keys or passwords.
+ * @param  {Array<Key>} privateKeys    (optional) private key with decrypted secret data
+ * @param  {Array<String>} passwords   (optional) password used to decrypt
  * @return {Array<{ data:Uint8Array, algorithm:String }>} array of object with potential sessionKey, algorithm pairs
  */
-Message.prototype.decryptSessionKeys = function(privateKey, password) {
+Message.prototype.decryptSessionKeys = function(privateKeys, passwords) {
   var keyPackets = [];
   return Promise.resolve().then(async () => {
-    if (password) {
+    if (passwords) {
       var symESKeyPacketlist = this.packets.filterByTag(enums.packet.symEncryptedSessionKey);
       if (!symESKeyPacketlist) {
         throw new Error('No symmetrically encrypted session key packet found.');
@@ -160,7 +160,7 @@ Message.prototype.decryptSessionKeys = function(privateKey, password) {
         } catch (err) {}
       }));
 
-    } else if (privateKey) {
+    } else if (privateKeys) {
       var pkESKeyPacketlist = this.packets.filterByTag(enums.packet.publicKeyEncryptedSessionKey);
       if (!pkESKeyPacketlist) {
         throw new Error('No public key encrypted session key packet found.');
--- a/src/message.js
+++ b/src/message.js
@@ -245,10 +245,10 @@ Message.prototype.getText = function() {
  * @param  {Array<Key>} keys           (optional) public key(s) for message encryption
  * @param  {Array<String>} passwords   (optional) password(s) for message encryption
  * @param  {Object} sessionKey         (optional) session key in the form: { data:Uint8Array, algorithm:String }
- * @param  {Boolean} useWildcard       (optional) use a key ID of 0 instead of the public key IDs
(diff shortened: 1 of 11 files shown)


src/message.js
```diff
@@ -91,13 +92,13 @@ Message.prototype.getSigningKeyIds = function() {
 
 /**
  * Decrypt the message. Either a private key, a session key, or a password must be specified.
- * @param  {Key} privateKey      (optional) private key with decrypted secret data
- * @param  {Object} sessionKey   (optional) session key in the form: { data:Uint8Array, algorithm:String }
- * @param  {String} password     (optional) password used to decrypt
+ * @param  {Array<Key>} privateKeys     (optional) private key with decrypted secret data
```
bartbutler: description, 'private keys'

src/message.js
```diff
@@ -91,13 +92,13 @@ Message.prototype.getSigningKeyIds = function() {
...
+ * @param  {Array<String>} passwords    (optional) password used to decrypt
```
bartbutler: description, 'passwords'

src/message.js
```diff
@@ -140,41 +141,56 @@ Message.prototype.decrypt = async function(privateKey, sessionKey, password) {
 };
 
 /**
- * Decrypt an encrypted session key either with a private key or a password.
- * @param  {Key} privateKey    (optional) private key with decrypted secret data
- * @param  {String} password   (optional) password used to decrypt
+ * Decrypt an encrypted session key either with private keys or passwords.
+ * @param  {Array<Key>} privateKeys    (optional) private key with decrypted secret data
+ * @param  {Array<String>} passwords   (optional) password used to decrypt
```
bartbutler: See above for description comments

src/message.js
```diff
@@ -140,41 +141,56 @@ Message.prototype.decrypt = async function(privateKey, sessionKey, password) {
...
       var symESKeyPacketlist = this.packets.filterByTag(enums.packet.symEncryptedSessionKey);
       if (!symESKeyPacketlist) {
         throw new Error('No symmetrically encrypted session key packet found.');
       }
       await Promise.all(symESKeyPacketlist.map(async function(packet) {
-        try {
-          await packet.decrypt(password);
-          keyPackets.push(packet);
-        } catch (err) {}
+        for (var i = 0; i < passwords.length; i++) {
```
bartbutler: Could do a 'await Promise.all(passwords.map(async function' thing like above and get rid of the eslint exception

src/message.js
```diff
@@ -140,41 +141,56 @@ Message.prototype.decrypt = async function(privateKey, sessionKey, password) {
...
-            keyPackets.push(packet);
-          } catch (err) {}
+        for (var i = 0; i < privateKeys.length; i++){
+          var privateKeyPackets;
+          if (packet.publicKeyId.isWildcard()) {
+            privateKeyPackets = privateKeys[i].getAllKeyPackets();
+          } else {
+            privateKeyPackets = [privateKeys[i].getKeyPacket([packet.publicKeyId])];
+          }
+          for (var j = 0; j < privateKeyPackets.length; j++) {
```
bartbutler: Same re: Promise.all thing above

src/message.js
```diff
@@ -295,9 +312,10 @@ Message.prototype.encrypt = function(keys, passwords, sessionKey) {
  * @param  {String} symAlgo            session key algorithm
  * @param  {Array<Key>} publicKeys     (optional) public key(s) for message encryption
  * @param  {Array<String>} passwords   (optional) for message encryption
+ * @param  {Boolean} useWildcard       (optional) if a key ID wildcard of 0 should be used instead of the public key IDs
```
bartbutler: confusing syntax in description, maybe 'use a key ID of 0 instead of the public key IDs'

src/message.js
```diff
@@ -295,9 +312,10 @@ Message.prototype.encrypt = function(keys, passwords, sessionKey) {
...
  * @return {Message}                   new message with encrypted content
  */
-export function encryptSessionKey(sessionKey, symAlgo, publicKeys, passwords) {
+export function encryptSessionKey(sessionKey, symAlgo, publicKeys, passwords, useWildcard) {
```
bartbutler: useWildcard default false?

src/message.js
```diff
@@ -309,7 +327,11 @@ export function encryptSessionKey(sessionKey, symAlgo, publicKeys, passwords) {
           throw new Error('Could not find valid key packet for encryption in key ' + key.primaryKey.getKeyId().toHex());
         }
         var pkESKeyPacket = new packet.PublicKeyEncryptedSessionKey();
-        pkESKeyPacket.publicKeyId = encryptionKeyPacket.getKeyId();
+        if (!useWildcard) {
```
bartbutler: Ternary would be more concise here

src/openpgp.js
```diff
@@ -245,24 +246,24 @@ export function encrypt({ data, publicKeys, privateKeys, passwords, sessionKey,
...
  * @param  {Message} message             the message object with the encrypted data
- * @param  {Key} privateKey              (optional) private key with decrypted secret key data or session key
- * @param  {Key|Array<Key>} publicKeys   (optional) array of public keys or single key, to verify signatures
- * @param  {Object} sessionKey           (optional) session key in the form: { data:Uint8Array, algorithm:String }
- * @param  {String} password             (optional) single password to decrypt the message
- * @param  {String} format               (optional) return data format either as 'utf8' or 'binary'
- * @param  {Signature} signature         (optional) detached signature for verification
+ * @param  {Key|Array<Key>} privateKeys     (optional) private key with decrypted secret key data or session key
+ * @param  {String|Array<String>} passwords (optional) single password to decrypt the message
+ * @param  {Object} sessionKey              (optional) session key in the form: { data:Uint8Array, algorithm:String }
```
bartbutler: Plural in descriptions

src/message.js
```diff
@@ -140,41 +141,56 @@ Message.prototype.decrypt = async function(privateKey, sessionKey, password) {
...
-      }
       await Promise.all(pkESKeyPacketlist.map(async function(packet) {
-        if (packet.publicKeyId.equals(privateKeyPacket.getKeyId())) {
-          try {
-            await packet.decrypt(privateKeyPacket);
-            keyPackets.push(packet);
-          } catch (err) {}
+        for (var i = 0; i < privateKeys.length; i++){
+          var privateKeyPackets;
+          if (packet.publicKeyId.isWildcard()) {
```
bartbutler: What about making getKeyPacket return AllKeyPackets if the public key ID is zero?  Then you can eliminate the if/else here.

Another thing to think about is to call this.getEncryptionKeyIds, then loop over privateKeys calling getKeyPacket(), then construct an object with properties keyed by keyID merging the results, and only then looping over privateKeyPackets. That would eliminate the nested loops you have here and probably be more efficient. Let me know what you think.

PR #637, opened 2018-02-13
Feat/compression

mmso:
This is a reboot of https://github.com/openpgpjs/openpgpjs/pull/513

I've added support for the **encrypt** function to compress messages. You can override the behavior by setting the `compression` parameter in the global config object or the options object passed to the **encrypt** function.

By default, it uses the value from the config which is set to `uncompressed`.

**NOTE:** It looks like a lot of changes for the test. But they are mostly white space indentation changes from using the `withCompression` test helper together with a modification of the arguments to encrypt + verification of the results.

Furthermore, instead of exposing a parameter on the decrypted object for the tests (like here https://github.com/openpgpjs/openpgpjs/pull/513/files#diff-eafe30f58aedef3cc702bff70e8917f4R15365), I use spies to check that the compress/decompress functions were called and with the correct parameters. Spying on functions does not work when it's run in the webworker so I had to disable those checks for the webworker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,35 @@ openpgp.decrypt(options).then(function(plaintext) {
 });
 ```
 
+#### Encrypt with compression
+
+By default, `encrypt` will not use any compression. It's possible to override that behavior in two ways:
+
+Either set the `compression` parameter in the options object when calling `encrypt`.
+
+```js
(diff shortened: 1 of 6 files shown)


bartbutler:
Thanks!

PR #647, opened 2018-02-16
Add a date parameter to the sign/verify/encrypt/decrypt functions

KAYLukas:
This allows:
- creating signatures with creation times different from the current clock time;
- verifying messages using a passed in time;
- loosely couple time from openpgp functions, so it doesn't have a static dependency allowing greater flexibility in testing.

In addition to adding the parameter I:
- removed the `verify_expired_keys` flag. Instead one can pass in a `null` date to disable expiration checking on keys
- started checking if the creation time of keys is later than the current time to detect invalid keys. This is disabled by passing in a `null` date.
- changed all `new Date()` calls to `util.currentDate()` which differs from `new Date()` in that it has second precision in contrast to millisecond precision. The reason for this is that openpgp uses normal unix timestamps (see [here](https://tools.ietf.org/html/rfc4880#section-3.5)) that doesn't store the exact milliseconds, so writing packages and reading the output afterwards currently change the creation date by dropping the millisecond precision.

### DIFF
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -24,7 +24,6 @@
  * @module cleartext
  */
 
-import util from './util.js';
 import config from './config';
 import armor from './encoding/armor';
 import enums from './enums';
@@ -72,7 +71,7 @@ CleartextMessage.prototype.getSigningKeyIds = function() {
  * @param  {Date} date                       (optional) The creation time of the signature that should be created
  * @return {module:message~CleartextMessage} new cleartext message with signed content
  */
-CleartextMessage.prototype.sign = async function(privateKeys, signature = null, date=new Date()) {
+CleartextMessage.prototype.sign = async function(privateKeys, signature=null, date=new Date()) {
   return new CleartextMessage(this.text, await this.signDetached(privateKeys, signature, date));
 };
 
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -69,25 +69,25 @@ CleartextMessage.prototype.getSigningKeyIds = function() {
  * Sign the cleartext message
  * @param  {Array<module:key~Key>} privateKeys private keys with decrypted secret key data for signing
  * @param  {Signature} signature             (optional) any existing detached signature
- * @param  {Date} creationDate     The creation time of the signature that should be created
+ * @param  {Date} date     The creation time of the signature that should be created
  * @return {module:message~CleartextMessage} new cleartext message with signed content
  */
-CleartextMessage.prototype.sign = async function(privateKeys, signature = null, creationDate = new Date()) {
-  return new CleartextMessage(this.text, await this.signDetached(privateKeys, signature, creationDate));
+CleartextMessage.prototype.sign = async function(privateKeys, signature = null, date = new Date()) {
+  return new CleartextMessage(this.text, await this.signDetached(privateKeys, signature, date));
 };
 
 /**
  * Sign the cleartext message
  * @param  {Array<module:key~Key>} privateKeys private keys with decrypted secret key data for signing
  * @param  {Signature} signature             (optional) any existing detached signature
- * @param  {Date} creationDate     The creation time of the signature that should be created
+ * @param  {Date} date     The creation time of the signature that should be created
  * @return {module:signature~Signature}      new detached signature of message content
  */
-CleartextMessage.prototype.signDetached = async function(privateKeys, signature = null, creationDate = new Date()) {
+CleartextMessage.prototype.signDetached = async function(privateKeys, signature = null, date = new Date()) {
   const literalDataPacket = new packet.Literal();
   literalDataPacket.setText(this.text);
 
-  return new Signature(await createSignaturePackets(literalDataPacket, privateKeys, signature, creationDate));
+  return new Signature(await createSignaturePackets(literalDataPacket, privateKeys, signature, date));
 };
 
 /**
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -24,13 +24,13 @@
  * @module cleartext
  */
 
+import util from './util.js';
 import config from './config';
 import armor from './encoding/armor';
 import enums from './enums';
 import packet from './packet';
 import { Signature } from './signature';
 import { createVerificationObjects, createSignaturePackets } from './message';
-import { getPreferredHashAlgo } from './key';
 
 /**
  * @class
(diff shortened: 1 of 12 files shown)


src/key.js
```diff
@@ -377,7 +376,7 @@ Key.prototype.getEncryptionKeyPacket = function(keyId) {
   // if no valid subkey for encryption, evaluate primary key
   const primaryUser = this.getPrimaryUser();
```
bartbutler: Use date here?

src/cleartext.js
```diff
@@ -24,13 +24,13 @@
  * @module cleartext
  */
 
+import util from './util.js';
```
sanjanarajan: Doesn't look like this is used anywhere?

src/key.js
```diff
@@ -976,15 +971,15 @@ SubKey.prototype.isValidEncryptionKey = async function(primaryKey) {
...
  * Returns true if the subkey can be used for signing of data
  * @param  {module:packet/secret_key|module:packet/public_key}  primaryKey The primary key packet
- * @param  {Boolean} allowExpired allows signature verification with expired keys
+ * @param  {Date} date use the given date for verification instead of the current time
  * @return {Boolean}
  */
-SubKey.prototype.isValidSigningKey = async function(primaryKey, allowExpired=false) {
-  if (await this.verify(primaryKey, allowExpired) !== enums.keyStatus.valid) {
+SubKey.prototype.isValidSigningKey = async function(primaryKey, date=new Date()) {
+  if (await this.verify(primaryKey) !== enums.keyStatus.valid) {
```
sanjanarajan: Should pass in date here?

src/key.js
```diff
@@ -1295,7 +1290,7 @@ async function wrapKeyObject(secretKeyPacket, secretSubkeyPacket, options) {
     const dataToSign = {};
     dataToSign.userid = userIdPacket;
     dataToSign.key = secretKeyPacket;
-    const signaturePacket = new packet.Signature();
+    const signaturePacket = new packet.Signature(new Date(1000));
```
sanjanarajan: Why 1000?

src/key.js
```diff
@@ -959,14 +953,15 @@ SubKey.prototype.toPacketlist = function() {
 /**
  * Returns true if the subkey can be used for encryption
  * @param  {module:packet/secret_key|module:packet/public_key}  primaryKey The primary key packet
+ * @param  {Date} date use the given date for verification instead of the current time
  * @return {Boolean}
  */
-SubKey.prototype.isValidEncryptionKey = async function(primaryKey) {
+SubKey.prototype.isValidEncryptionKey = async function(primaryKey, date=new Date()) {
```
sanjanarajan: Should pass in date to this.verify?

src/message.js
```diff
@@ -443,26 +446,30 @@ Message.prototype.compress = function(compression) {
...
  * Create a detached signature for the message (the literal data packet of the message)
- * @param  {Array<module:key~Key>}           privateKey private keys with decrypted secret key data for signing
+ * @param  {Array<module:key~Key>}           privateKeys private keys with decrypted secret key data for signing
  * @param  {Signature} signature             (optional) any existing detached signature
+ * @param  {Date} date                       (optional) override the creation time of the signature
  * @return {module:signature~Signature}      new detached signature of message content
  */
-Message.prototype.signDetached = async function(privateKeys=[], signature=null) {
+Message.prototype.signDetached = async function(privateKeys=[], signature=null, date=new Date()) {
+  const packetlist = new packet.List();
```
sanjanarajan: Doesn't look like this variable is used

bartbutler:
All browser tests fail (web worker?). You may have to check that the Date class serializes and deserializes correctly with the structured clone algorithm and put in a special case for it if it does not. Also, use util.Uint8Array2str on binary data before checking equality or it fails, at least in browsers.

You can use 'grunt browsertest' locally to run the tests in a browser.

screen shot 2018-02-17 at 8 45 27 am

src/key.js
```diff
@@ -1008,15 +1002,15 @@ SubKey.prototype.verify = async function(primaryKey, allowExpired=false) {
     return enums.keyStatus.revoked;
   }
   // check V3 expiration time
-  if (!allowExpired && this.subKey.version === 3 && this.subKey.expirationTimeV3 !== 0 &&
-      Date.now() > (this.subKey.created.getTime() + this.subKey.expirationTimeV3*24*3600*1000)) {
+  if (date !== null && this.subKey.version === 3 && this.subKey.expirationTimeV3 !== 0 &&
+      util.normalizeDate(date) > (this.subKey.created.getTime() + this.subKey.expirationTimeV3*24*3600*1000)) {
```
bartbutler: Check creation time?

src/key.js
```diff
@@ -1026,8 +1020,8 @@ SubKey.prototype.verify = async function(primaryKey, allowExpired=false) {
     }
     // check V4 expiration time
     if (that.subKey.version === 4) {
-      if (!allowExpired && bindingSignature.keyNeverExpires === false &&
-         Date.now() > (that.subKey.created.getTime() + bindingSignature.keyExpirationTime*1000)) {
+      if (date !== null && bindingSignature.keyNeverExpires === false &&
+          util.normalizeDate(date) > (that.subKey.created.getTime() + bindingSignature.keyExpirationTime*1000)) {
```
bartbutler: Check creation time?

src/message.js
```diff
@@ -358,11 +360,12 @@ export function encryptSessionKey(sessionKey, symAlgo, publicKeys, passwords, wi
 
 /**
  * Sign the message (the literal data packet of the message)
- * @param  {Array<module:key~Key>}        privateKey private keys with decrypted secret key data for signing
+ * @param  {Array<module:key~Key>} privateKeys private keys with decrypted secret key data for signing
```
bartbutler: Indentation

src/key.js
```diff
@@ -471,13 +464,13 @@ Key.prototype.verifyPrimaryKey = async function(allowExpired=false) {
...
-  const primaryUser = this.getPrimaryUser(allowExpired);
+  const primaryUser = this.getPrimaryUser(date);
   if (!primaryUser) {
     return enums.keyStatus.invalid;
   }
   // check V4 expiration time
-  if (!allowExpired && this.primaryKey.version === 4 && primaryUser.selfCertificate.keyNeverExpires === false &&
-    Date.now() > (this.primaryKey.created.getTime() + primaryUser.selfCertificate.keyExpirationTime*1000)) {
+  if (date !== null && this.primaryKey.version === 4 && primaryUser.selfCertificate.keyNeverExpires === false &&
+    util.normalizeDate(date) > (this.primaryKey.created.getTime() + primaryUser.selfCertificate.keyExpirationTime*1000)) {
```
bartbutler: Check creation time?

src/key.js
```diff
@@ -460,8 +453,8 @@ Key.prototype.verifyPrimaryKey = async function(allowExpired=false) {
     return enums.keyStatus.revoked;
   }
   // check V3 expiration time
-  if (!allowExpired && this.primaryKey.version === 3 && this.primaryKey.expirationTimeV3 !== 0 &&
-    Date.now() > (this.primaryKey.created.getTime() + this.primaryKey.expirationTimeV3*24*3600*1000)) {
+  if (date !== null && this.primaryKey.version === 3 && this.primaryKey.expirationTimeV3 !== 0 &&
+    util.normalizeDate(date) > (this.primaryKey.created.getTime() + this.primaryKey.expirationTimeV3*24*3600*1000)) {
```
bartbutler: Check creation time?

PR #654, opened 2018-02-25
Modernizes MPI handling and public key algorithms; jsbn.js is dead, long live bn.js!

### LINKED ISSUE #67: Inconsistent bit length on key generation
Calling:

```
openpgp.generate_key_pair(1, 1024, 'a@bc.de', '12345')
```

Produces keys of either 1024 or 1023 bit size. Same applies to 2048 and 4096: in around 50% of the cases we get only 2047 or 4095 bit keys.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1236047-inconsistent-bit-length-on-key-generation?utm_campaign=plugin&utm_content=tracker%2F290945&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F290945&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #507: Error: Error decrypting message: Cannot read property 'toBigInteger' of undefined
Can someone point me in the right direction?

```
Error: Error decrypting message: Cannot read property 'toBigInteger' of undefined
```

This happens to some (about 0.5-1%) users, with a private key generated by openpgp.js. Regardless of the message content, they will always get this error when trying to decrypt. 

It seems like the key was somehow wrongly created.

For now, when creating keys, I will probably have to do a cycle of

1. create a new key
2. encrypt message for it
3. try to decrypt it
4. catch  `Cannot read property 'toBigInteger' of undefined` -> go to step 1

Is there any way to debug this, without access to the private keys of these users?

### LINKED ISSUE #631: Unify BN libraries
Eliminate jsbn in favor of one of the modern BN libraries, probably BN.js or asmCrypto's bn library. Need to do appropriate benchmarks first:

https://peterolson.github.io/BigInteger.js/benchmark/

mahrud:
Tom Wu's 2005-2009 `jsbn.js` is entirely replaced by Fedor @Indutny's [bn.js](https://github.com/indutny/bn.js/), which considerably outperforms the former library in relevant [benchmarks](https://mahrud.github.io/benchmark/).
Because of this, any part of the code that handled multi-precision integers had to be updated. This includes all public key related algorithms. In particular, new probabilistic random prime generation algorithms are added to assist with RSA key generation in javascript.

Closes #631. Probably fixes #507 and fixes #67 as well. Also fixes sporadic bug in Ed25519 signing caused by jsbn.js truncation.


src/crypto/public_key/elliptic/ecdsa.js
```diff
@@ -18,52 +18,43 @@
...
  * Sign a message using the provided key
- * @param  {String}      oid        Elliptic curve for the key
- * @param  {enums.hash}  hash_algo  Hash algorithm used to sign
- * @param  {Uint8Array}  m          Message to sign
- * @param  {BigInteger}  d          Private key used to sign
- * @return {{r: BigInteger, s: BigInteger}}  Signature of the message
+ * @param  {module:type/oid} oid        Elliptic curve object identifier
+ * @param  {enums.hash}      hash_algo  Hash algorithm used to sign
+ * @param  {Uint8Array}      m          Message to sign
+ * @param  {Uint8Array}      d          Private key used to sign the message
```
bartbutler: Why is this Uint8Array and it is BN for EdDSA?

src/crypto/public_key/elliptic/ecdsa.js
```diff
@@ -18,52 +18,43 @@
...
- * @param  {String}      oid        Elliptic curve for the key
- * @param  {enums.hash}  hash_algo  Hash algorithm used in the signature
- * @param  {{r: BigInteger, s: BigInteger}}  signature  Signature to verify
- * @param  {Uint8Array}  m          Message to verify
- * @param  {BigInteger}  Q          Public key used to verify the message
+ * @param  {module:type/oid} oid        Elliptic curve object identifier
+ * @param  {enums.hash}      hash_algo  Hash algorithm used in the signature
+ * @param  {{r: BN, s: BN}}  signature  Signature to verify
+ * @param  {Uint8Array}      m          Message to verify
+ * @param  {Uint8Array}      Q          Public key used to verify the message
```
bartbutler: Why is this Uint8Array and it is BN for EdDSA?

src/crypto/public_key/rsa.js
```diff
@@ -18,254 +18,205 @@
...
+          publicExponent: E.toArrayLike(Uint8Array), // take three bytes (max 65537) for exponent
           hash: {
             name: 'SHA-1' // not required for actual RSA keys, but for crypto api 'sign' and 'verify'
           }
         };
-
-        keys = webCrypto.generateKey(keyGenOpt, true, ['sign', 'verify']);
-        if (typeof keys.then !== 'function') { // IE11 KeyOperation
-          keys = util.promisifyIE11Op(keys, 'Error generating RSA key pair.');
-        }
```
bartbutler: This is not needed?

src/crypto/public_key/rsa.js
```diff
@@ -18,254 +18,205 @@
...
-    // JS code
-    //
-
-    return new Promise(function(resolve) {
-      const key = new KeyObject();
-      const rng = new SecureRandom();
-      const qs = B >> 1;
-      key.e = parseInt(E, 16);
-      key.ee = new BigInteger(E, 16);
+    while (true) {
```
bartbutler: Why is this in a while loop?

src/crypto/random.js
```diff
@@ -95,41 +97,29 @@ export default {
...
+    const modulus = max.sub(min);
+    const length = modulus.byteLength();
 
-  getRandomBigIntegerInRange: function(min, max) {
-    if (max.compareTo(min) <= 0) {
-      throw new Error('Illegal parameter value: max <= min');
-    }
+    // Using a while loop is necessary to avoid bias
+    do {
+      r = new BN(this.getRandomBytes(length));
```
bartbutler: Can't you do a bit shift here to reduce the worst-case behavior of this loop?

src/crypto/public_key/elliptic/ecdsa.js
mahrud: There are a few tests in `test/crypto/elliptic.js` that pass in arrays or typed arrays directly to this function. I can just as easily change those.

src/crypto/public_key/rsa.js
mahrud: Which part? The `promisifyIE11Op` is not needed because `await` can be used for both promises or regular return values.

src/crypto/public_key/elliptic/ecdsa.js
bartbutler: Consistency is a virtue--I would change it if it's easy.

src/crypto/public_key/rsa.js
bartbutler: Yeah, but in IE's case the return value is not a regular return value, it's a KeyOperation object which is it's own async thing. I would think you would still need this...

src/crypto/public_key/rsa.js
mahrud: Ah, good catch. The previous implementation would restart if certain checks failed, but those checks are all in `randomProbablePrime` now.

@sanjanarajan do you know if there is any case where two primes would not be good to use together? Something about common factors of `p-1` and `q-1` perhaps?

src/crypto/random.js
mahrud: I think it'll introduce bias to the random values, unless there's a tricky way that changes the probability of all values uniformly.

bartbutler: I don't see why it would introduce any bias. It would basically be using the number of bits for the length instead of the number of bytes, but otherwise there is no difference to your current method except the worst-case behavior gets a lot better.

src/crypto/public_key/rsa.js
mahrud: Ah, sorry I missed that. I'll revert this change in a new commit.

src/crypto/random.js
mahrud: There is a difference, and the reason is that the number of bit shifts is not determined only by the most significant bit. Suppose we want a random number 0 <= x < 0110. Now take three random values 0011, 0110 and 0111. All of them map to 0011 after bit shifts, whereas only 0000 can be mapped to 0000. Hence 0011 is three times as likely as 0000.

bartbutler: I don't think that's correct. So in your example, you have 3 random bits, and want to convert them to two random bits. So great, ignore one of the bits. That is exactly what we are doing here. If I bit shift 0011 to the right I get 0001, not 0011, and 0001 bit-shifted also becomes 0000. So that's two each for 11 and 00, exactly how it should be. There is also 2 for 01 and 10,  because fundamentally all we are doing it ignoring bits.

mahrud: And what happens to 1111 and 1110? After bit shift we get 0111, which is out of bounds. Shifting again would give two more possibilities for 0011.

src/crypto/public_key/rsa.js
```diff
@@ -176,7 +192,8 @@ export default {
 
       // export the generated keys as JsonWebKey (JWK)
       // https://tools.ietf.org/html/draft-ietf-jose-json-web-key-33
-      let jwk = await webCrypto.exportKey('jwk', keyPair.privateKey);
+      let jwk = webCrypto.exportKey('jwk', keyPair.privateKey);
+      jwk = promisifyIE11Op(jwk, 'Error exporting RSA key pair.');
```
bartbutler: Don't need it here, this is old Safari only. Revert.

src/crypto/public_key/rsa.js
```diff
@@ -158,7 +173,8 @@ export default {
             name: 'SHA-1' // not required for actual RSA keys, but for crypto api 'sign' and 'verify'
           }
         };
-        keyPair = await webCrypto.generateKey(keyGenOpt, true, ['sign', 'verify']);
+        keyPair = webCrypto.generateKey(keyGenOpt, true, ['sign', 'verify']);
+        keyPair = promisifyIE11Op(keyPair, 'Error generating RSA key pair.');
```
bartbutler: Missing an 'await' here.

src/crypto/public_key/rsa.js
```diff
@@ -18,254 +18,220 @@
...
-  this.keyObject = KeyObject;
-}
+    const phi = p.subn(1).mul(q.subn(1));
+    return {
+      n: p.mul(q),
+      e: E,
+      d: E.invm(phi),
+      q: q,
+      p: p,
+      // dq: d.mod(q.subn(1)),
```
sanjanarajan: What was/is the purpose of these? We should probably either return these values for both the fallback and webcrypto or neither. But they don't seem necessary to me.

src/crypto/public_key/dsa.js
```diff
@@ -18,111 +18,106 @@
...
-    let k;
-    let s1;
-    let s2;
     while (true) {
-      k = random.getRandomBigIntegerInRange(BigInteger.ONE, q.subtract(BigInteger.ONE));
-      s1 = (g.modPow(k, p)).mod(q);
-      s2 = (k.modInverse(q).multiply(hash.add(x.multiply(s1)))).mod(q);
-      if (s1 !== 0 && s2 !== 0) {
-        break;
+      // TODO confirm this range
```
sanjanarajan: According to this: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf - k should be in [1, q-1] inclusive (go to Appendix B).

src/crypto/public_key/elgamal.js
```diff
@@ -18,39 +18,42 @@
...
+   * m, p, g, y are all BN
+   * returns { c1: BN, c2: BN }
+   */
+  encrypt: function(m, p, g, y) {
+    const redp = new BN.red(p);
+    const mred = m.toRed(redp);
+    const gred = g.toRed(redp);
+    const yred = y.toRed(redp);
+    // TODO confirm this range
+    const k = random.getRandomBN(two, p.subn(1)); // returns in [2, p-2]
```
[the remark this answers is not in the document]
sanjanarajan: See section 11.5 here: https://crypto.stanford.edu/~dabo/cryptobook/BonehShoup_0_4.pdf

sanjanarajan:
I also noticed some TODOs in other places in the code - like avoiding race conditions in verifyPrimaryUser. Are those going to be part of this PR?

src/crypto/random.js
```diff
@@ -95,41 +97,33 @@ export default {
...
-    if (bits % 8 > 0) {
-      randomBits = String.fromCharCode(((2 ** (bits % 8)) - 1) &
-        randomBits.charCodeAt(0)) +
-        randomBits.substring(1);
-    }
-    const mpi = new type_mpi(randomBits);
-    return mpi.toBigInteger();
-  },
+    let r;
+    const modulus = max.sub(min);
```
sanjanarajan: Why "modulus"?

bartbutler: Yeah, 'diff' or 'difference' would be better.

src/crypto/public_key/rsa.js
mahrud: Do you mean dq and dp? They're for faster RSA encryption and verification, but OpenPGP doesn't support them. I'll leave it there for the sake of reference.

Issue #641, opened 2018-02-14
performance - option to use N worker threads in initWorker


tomholub:
I've done some informal performance testing (involving our own code + OpenPGP.js) to see the difference between using a single worker thread vs 8 threads with a round-robin.

The test creates 20 or 50 iframes with that load and decrypt a short message, with 1 or 8 workers. This mimics opening an email thread. Results in seconds to decrypt all (lower is better):

```
20 msgs
1 worker	9.31	9.20	9.17
8 workers	7.09	6.42	6.35

50 msgs
1 worker	33.51	24.12	24.13
8 workers	15.25	15.44	15.41
```

I think it would be worth to allow N worker threads setting for initWorker.

For the tests, I have been following this pattern:
```js
function initWorker() {
  if (worker || typeof window !== 'undefined' && window.Worker) {
    asyncProxy = [
      new _async_proxy2.default({ path: path, worker: worker, config: _config2.default }),
      new _async_proxy2.default({ path: path, worker: worker, config: _config2.default }),
      ... 
    ];
    return true;
  }
}

// round robin worker select
function getWorker() {
    if(asyncProxy) {
        asyncProxyLast = (asyncProxyLast + 1) % asyncProxy.length;
        return asyncProxy[asyncProxyLast];
    }
  return asyncProxy;
}

// always use getWorker() instead of direct reference when using, eg:
  if (asyncProxy) {
    return getWorker().delegate('reformatKey', options);
  }
```

This is just to give you a rough idea, it would probably be better to use `asyncProxies = []` and test for zero length + other cleanups.

tomholub:
To clarify, the 1 or 8 workers were shared across all iframes: decryption was happening in a background script. That's why having a single worker becomes a bottleneck, and having an option to init N workers improves performance in a noticeable way.

bartbutler:
Agreed, this is a good idea. I wonder if multiple workers utilize multiple cores or if all workers run in the same thread. The answer is probably browser-dependent.

tomholub:
To my knowledge, they are supposed to utilize separate threads. Web worker
docs say that each web worker spawns an actual separate process which may
run in another core. At the same time, I did not see a 4x improvement in
speed (got 4 cores), so also unsure. There was a noticeable improvement
though.

bartbutler:
Separate process may still have been sharing cores. In any case, this is a good idea, we would welcome a PR, otherwise I'll try to get to it when I can.

tomholub:
[...]

PR #661, opened 2018-03-06
Eliminate running out of random bytes in workers, allow N workers

### LINKED ISSUE #546: Error decrypting message: Random number buffer depleted
Hi,

I am trying to speed up openpgp.js (latest commit dc04af4) using web workers and the native AES-GCM (adding the setup `openpgp.config.aead_protect=true`), as mentioned in the documentation, with the expectation of a significant speed increase.

I decrypt many messages in a loop and I get the following error messages after ~200 decryptions:
`Error decrypting message: Random number buffer depleted`

When I don't use web workers, I don't get these error messages.

However, before the errors, it seems that using web workers is faster than without them. Without web workers I don't see the expected 30x speedup of using native AES-GCM.

Any idea?

I use Safari 10.1 on macOs Sierra 10.12.4.

Thanks in advance!

 - Ed

bartbutler:
Should close #546 and #641


src/crypto/pkcs1.js
```diff
@@ -69,25 +71,24 @@ export default {
      * create a EME-PKCS1-v1_5 padding (See {@link https://tools.ietf.org/html/rfc4880#section-13.1.1|RFC 4880 13.1.1})
      * @param {String} M message to be encoded
      * @param {Integer} k the length in octets of the key modulus
-     * @return {String} EME-PKCS1 padded message
+     * @return {Promise<String>} EME-PKCS1 padded message
```
mahrud: Oooh, should we be specifying when a function returns a promise? I didn't do that anywhere!

bartbutler: I think so, but I don't think it's a huge deal. I probably missed it a bunch of times in this PR--maybe we should grep for 'async'.

PR #672, opened 2018-03-15
Adds Brainpool Curves + tests + docs

mahrud:
Title says it all. The curves brainpoolP256r1, brainpoolP384r1, and brainpoolP512r1, as defined in [RFC5639](https://tools.ietf.org/html/rfc5639), are added in our own [fork](https://github.com/openpgpjs/elliptic/commit/e187e706e11fa51bcd20e46e5119054be4e2a4a6) of elliptic and added to openpgp.js in this PR.

#### Caveat:
- ~No compatibility test with GnuPG yet.~ thanks @sanjanarajan!
- Node.js native cryptography is disabled. The fix is to change the ANS1 algorithm identifier in `key.js` depending on the key type.


tanx:
@mahrud I've done an initial code review and looked and code quality and unit tests. Great work and thanks for stepping up on such short notice. Very much appreciated.

@toberndo we should still cover brainpool as well as other recent additions in the security audit.

Other than that it's good to merge from my side.

PR #679, opened 2018-03-22
When signature type is text, canonicalize line endings

sanjanarajan:


src/message.js
```diff
@@ -564,9 +564,14 @@ export async function createVerificationObjects(signatureList, literalDataList,
       }
     }));
 
+    const literalDataPacket = literalDataList[0];
```
bartbutler: A comment here explaining why this is done might be nice.

131:
I dunno why, but i think this breaks my setup. I'll try to give more details

Issue #675, opened 2018-03-19
Native Node.js support for X25519 and Brainpool curves


mahrud:
Node.js has bindings with the local OpenSSL for elliptic curve computations. In particular, `nodeCrypto.getCurves()` returns whatever curves the local OpenSSL supports, which can be found using this command:
```
    $ openssl ecparam -list_curves
```

Currently the curve names of all curves supported by OpenPGP.js are defined in `curves.js`:
https://github.com/openpgpjs/openpgpjs/blob/f88c1bc8a43e2694b01b7441f46bb4874033e8b6/src/crypto/public_key/elliptic/curves.js#L47-L57

However, in `key.js`, the ASN1 encoding functions are only defined for ECDSA:
https://github.com/openpgpjs/openpgpjs/blob/afc9164e7db688e1a9907c0340ff6ca0035e6df3/src/crypto/public_key/elliptic/key.js#L234-L250
Because of that, I've manually disabled native Node.js computation on ED25519, X25519, and the Brainpool curves; for instance:
https://github.com/openpgpjs/openpgpjs/blob/f88c1bc8a43e2694b01b7441f46bb4874033e8b6/src/crypto/public_key/elliptic/curves.js#L98
To enable native Node.js support, we need to:
- Add ASN1 definitions for necessary new algorithms as above. This also involves fixing this line:
https://github.com/openpgpjs/openpgpjs/blob/afc9164e7db688e1a9907c0340ff6ca0035e6df3/src/crypto/public_key/elliptic/key.js#L210
- Modify `nodeSign` and `nodeVerify` in `key.js` to call the appropriate ASN1 encoder/decoder.

PR #680, opened 2018-03-26
Add native Brainpool support on Node.js

twiss:
Fixes part of #675. Doesn't add support for ED25519 and X25519 yet because the version of OpenSSL included in the latest Node.js (1.0.2m) doesn't support them yet.


src/crypto/public_key/elliptic/curves.js
```diff
@@ -98,32 +98,32 @@ const curves = {
     node: false // nodeCurves.ed25519 TODO
   },
   curve25519: {
-    oid: [0x06, 0x08, 0x2B, 0x06, 0x01, 0x04, 0x01, 0x97, 0x55, 0x01, 0x05, 0x01],
+    oid: [0x06, 0x0A, 0x2B, 0x06, 0x01, 0x04, 0x01, 0x97, 0x55, 0x01, 0x05, 0x01],
```
bartbutler: Was the curve25519 OID really wrong?

bartbutler: Huh, no kidding. Cool.

mahrud:
At least on Mac and Linux, Node.js itself doesn't include OpenSSL. It hooks into the installed version of it. If you're on Linux, the default OpenSSL version probably only has non-patented curves, but if you compile OpenSSL from source you *should* be able to see them.

src/crypto/public_key/elliptic/curves.js
mahrud: I can't count :100:

twiss:
Hm — I actually tried compiling OpenSSL from source and didn't see them. I probably did something wrong, I'll try again.

mahrud:
Ah I see. Thanks for looking into this! I bet it's an issue with endian-ness.
I wonder how gpg2 supports X25519 now :confused:

twiss:
The OpenSSL C API does support it, although it apparently depends on which interface you use. So GPG probably uses that, and so could Node — but I don't know how they interface with OpenSSL so I'm not sure how much work it would be.

PR #685, opened 2018-04-17
Use crypto.getRandomValues in Workers

twiss:
Modern browsers support getRandomValues in Workers. Webkit added support in August 2016, so Safari 9 doesn't support it (Safari 10 probably does — not tested). Once we drop support for older browsers the code for getting randomness from the main thread can be removed.

### DIFF
--- a/src/worker/worker.js
+++ b/src/worker/worker.js
@@ -34,9 +34,7 @@ importScripts('openpgp.js');
 var openpgp = window.openpgp;
 
 var randomQueue = [];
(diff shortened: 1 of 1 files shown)


sanjanarajan:
Discussed with @bartbutler and the conclusion was that we don't want to drop support for these older browsers yet. But if the new solution is more performant, we could selectively use it for the browsers that do support it and fall back to the original solution for the older browsers.

bartbutler:
Looks good to me.

PR #687, opened 2018-04-18
Only use the most recent self-signature

### LINKED ISSUE #266: Support for multiple subkey binding signatures
If a subkey has mulitple binding signatures (happens e.g. if you extend the expiration time of a key with GPG) then OpenPGP.js will only consider the last signature of the subkey. As a consequence the subkey could get the status invalid although there exists a valid binding signature.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5887221-support-for-multiple-subkey-binding-signatures?utm_campaign=plugin&utm_content=tracker%2F290945&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F290945&utm_medium=issues&utm_source=github).
</bountysource-plugin>

twiss:
Fixes #266.


src/key.js
```diff
@@ -513,32 +528,31 @@ Key.prototype.getValidUsers = async function(date=new Date(), allowExpired=false
...
-      }
-      if (!allowExpired && cert.isExpired(date)) {
-        continue;
-      }
-      lastPrimaryUserID = cert.isPrimaryUserID;
-      lastCreated = cert.created;
-      validUsers.push({ index: i, user: user, selfCertification: cert });
+    const cert = getLatestSignature(user.selfCertifications);
+    // skip if certificate is not the most recent
+    if ((cert.isPrimaryUserID && cert.isPrimaryUserID < lastPrimaryUserID) ||
```
sanjanarajan: I know this code is copied from before, but this check is a vestige of our previous implementation for getPrimaryUser, and I am realizing it is no longer really correct if this function is meant to return *all* valid users. This is especially important as it relates to checking expiration time.

So what I propose that we do is remove the lastCreated and lastPrimaryUserID variables, remove this check, and then make this for-loop a Promise.all.

twiss: Yeah, that code is incorrect. But the only calling function that uses the whole list of users is `Key#getExpirationTime`, which in turn is never used (although it's part of the public API, of course). `verifyPrimaryKey`, `getEncryptionKeyPacket` etc instead only check the expiration time based on the primary user. So.. are we sure that checking every user is important?

sanjanarajan: I think it's best to just be correct and return the expected result here for both this function and Key.prototype.getExpirationTime, because at least getExpirationTime is pretty frequently used by people who use the library. Plus it's debatable whether we are saving any time doing it this way vs. making it a Promise.all.

PR #691, opened 2018-04-25
Implement RFC4880bis-04

### LINKED ISSUE #627: Add AEAD-enabled V5 keys and incorporate changes from RFC4880bis-04
Warner Koch published the latest draft of the ~[RFC4880bis-03](https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-03) on Dec 29, 2017~ [RFC4880bis-04](https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-04) on Jan 25, 2018. AEAD specifications for V5 keys are added, including an AEAD encrypted data packet. We only support AEAD as a global setting, but even that only supports AES-GCM, whereas this draft mandates support for AES-EAX. This includes changes in the S2K function as well. Also added are two new public key options: AEDH and AEDSA.

There are some smaller changes too. See the full diff [here](https://tools.ietf.org/rfcdiff?url2=draft-ietf-openpgp-rfc4880bis-03.txt) and [here](https://tools.ietf.org/rfcdiff?url2=draft-ietf-openpgp-rfc4880bis-04.txt).

twiss:
Fixes #627. Previous discussion: https://github.com/twiss/openpgpjs/pull/1.

Implementation Notes
=============================================================================================================================

  

  

Native AES-EAX
--------------

  

Native Web Crypto and Node Crypto do not support AES-EAX, but they do support AES-CTR, which AES-EAX is built upon. Therefore, I've implemented EAX in a way that is agnostic to the underlying CTR implementation, and automatically selects native CTR whenever available.

  

Similarly, EAX also makes use of CMAC, which in turn makes use of CBC, which is supported by Web and Node Crypto, so I've made use of that as well.

  

EAX mode itself (as opposed to AES-EAX) does not require AES and is flexible enough to be used with any block cipher, but RFC4880bis-04 only mandates support for AES-128. We support AES-{128, 192, 256} (although AES-192 is not supported by Web Crypto in Chrome).

  

A note on implementing a draft specification
--------------------------------------------

  

RFC4880bis-04 is a work in progress and may change at any time, and any implementation of it should be considered as such as well. The OpenPGP.js documentation already did a good job of making users aware that enabling AEAD may break compatibility with other implementations, but I have now also added a warning that future versions of OpenPGP.js may break compatibility with it, as well.

  

Backwards compatibility
-----------------------

  

Unfortunately, RFC4880bis-04 is already incompatible with our current implementation of the previous draft: for example, the AEAD Encrypted Data Packet has gained fields for the cipher algorithm, AEAD algorithm, and chunk size.

  

A look at [GitHub code search reveals](https://github.com/search?q=openpgp.config.aead_protect+%3D+true&type=Code&utf8=%E2%9C%93) that `aead_protect` is actually used quite a lot, even in apps that store encrypted PGP messages on disk or present them to the user (luckily no large projects seem to do this). This is probably because [our first example of setting up OpenPGP.js](https://github.com/openpgpjs/openpgpjs#set-up) includes it (which is modified now).

  

Therefore, we have decided to maintain support for the previous version, and only enable support for the new version when setting `openpgp.config.aead_protect_version = 4`.

AEAD Support Feature Flags
--------------------------

  

The new draft adds feature flags for AEAD support in public keys. We add those flags when support for the new draft is enabled. When encrypting using a public key, we only use AEAD when the target key claims support for it. (This only applies for the new draft — the old draft didn't support feature flags so we don't look for them.)

  

GCM in the new draft
------
Because GCM is faster than EAX and OCB using Web Crypto (although EAX doesn't lag too far behind, it's not 100% native as GCM is, and so still a bit slower), we have implemented it in the new draft as well, as a "private/experimental algorithm" (with algorithm ID 100). Since GCM is not defined by the spec, we had to decide on a few implementation details ourselves, most notably the computation of the nonce from the IV and chunk counter. A future version of the standard may define GCM mode differently, hopefully under a different ID (e.g. ID 3) so that we can maintain backwards compatibility.

You can enable GCM by setting `openpgp.config.aead_mode = openpgp.enums.aead.gcm`.

Preferred AEAD Algorithms
-------------------------

  

The new draft also adds a packet containing the AEAD algorithms that the user of the public key prefers. When encrypting using a public key, those preferences have precedence over the `aead_mode` setting. When generating a public key, we claim preference for EAX over OCB, since EAX is for a large part implemented using native crypto.

  

The `aead_mode` setting currently does not influence the generated preferences, causing it to have no effect when solely using public-key cryptography. This restriction also applies to the other algorithm options (`prefer_hash_algorithm`, `encryption_cipher` and `compression`) and could be fixed together with those in the future.

Version 5 Keys
------
The draft also specifies V5 Public/Private Key packets and V5 Symmetric-Key Encrypted Session Key packets. The main difference with their V4 counterparts is that they encrypt their contents with an AEAD algorithm. We generate these V5 packets when support for this draft version is enabled.

V5 Public/Private Keys also have some other minor improvements. As per the spec:

> The version 5 format is similar to the version 4 format except for the addition of a count for the key material.  This count helps parsing secret key packets (which are an extension of the public key packet format) in the case of an unknown algoritm.  In addition, fingerprints of version 5 keys are calculated differently from version 4 keys.


src/packet/literal.js
```diff
@@ -37,21 +37,20 @@ function Literal(date=new Date()) {
...
  */
-Literal.prototype.setText = function(text) {
-  // normalize EOL to \r\n
-  text = text.replace(/\r\n/g, "\n").replace(/\r/g, "\n").replace(/[ \t]+\n/g, "\n").replace(/\n/g, "\r\n");
-  this.format = 'utf8';
-  // encode UTF8
-  this.data = util.str_to_Uint8Array(util.encode_utf8(text));
+Literal.prototype.setText = function(text, format='utf8') {
+  this.format = format;
+  this.text = text;
```
bartbutler: Zero out this.data here as well? Also should probably set this.text to null in setBytes if not done already.

src/packet/sym_encrypted_session_key.js
```diff
@@ -145,14 +175,24 @@ SymEncryptedSessionKey.prototype.encrypt = async function(passphrase) {
...
+    const mode = crypto[this.aeadAlgorithm];
+    this.iv = await crypto.random.getRandomBytes(mode.ivLength); // generate new random IV
+    const adata = new Uint8Array([0xC0 | this.tag, this.version, enums.write(enums.symmetric, this.sessionKeyEncryptionAlgorithm), enums.write(enums.aead, this.aeadAlgorithm)]);
+    const modeInstance = await mode(algo, key);
+    this.encrypted = await modeInstance.encrypt(this.sessionKey, this.iv, adata);
+  } else {
+    const algo_enum = new Uint8Array([enums.write(enums.symmetric, this.sessionKeyAlgorithm)]);
+
+    const private_key = util.concatUint8Array([algo_enum, this.sessionKey]);
+
```
bartbutler: Remove some newlines?

src/packet/literal.js
twiss: Good idea 👍 done

README.md
```diff
@@ -54,17 +54,32 @@ OpenPGP.js [![Build Status](https://travis-ci.org/openpgpjs/openpgpjs.svg?branch
...
-    | curve25519      | ECDH       | N/A       | Yes      | No (TODO)  | No        |
-    | ed25519         | N/A        | EdDSA     | Yes      | No (TODO)  | No        |
-    | brainpoolP256r1 | ECDH       | ECDSA     | Yes      | No (TODO)  | No        |
-    | brainpoolP384r1 | ECDH       | ECDSA     | Yes      | No (TODO)  | No        |
-    | brainpoolP512r1 | ECDH       | ECDSA     | Yes      | No (TODO)  | No        |
+    | brainpoolP256r1 | ECDH       | ECDSA     | Yes      | Yes*       | No        |
+    | brainpoolP384r1 | ECDH       | ECDSA     | Yes      | Yes*       | No        |
+    | brainpoolP512r1 | ECDH       | ECDSA     | Yes      | Yes*       | No        |
+    | curve25519      | ECDH       | N/A       | Yes      | No         | No        |
+    | ed25519         | N/A        | EdDSA     | Yes      | No         | No        |
```
mahrud: Just out of curiosity, is there a specific reason why X25519 with Node Crypto doesn't work?

src/crypto/public_key/elliptic/curves.js
```diff
@@ -182,14 +182,14 @@ Curve.prototype.keyFromPublic = function (pub) {
 
 Curve.prototype.genKeyPair = async function () {
   let keyPair;
-  if (webCrypto && this.web) {
```
mahrud: Was this a big fix? If so and webCrypto and nodeCrypto variables aren't used anymore you should delete them from the top of the file too.

README.md
twiss: Yeah, see https://github.com/openpgpjs/openpgpjs/pull/680#issuecomment-376439537

src/crypto/public_key/elliptic/key.js
```diff
@@ -45,7 +45,7 @@ function KeyPair(curve, options) {
 }
 
 KeyPair.prototype.sign = async function (message, hash_algo) {
-  if (webCrypto && this.curve.web) {
+  if (this.curve.web && util.getWebCrypto()) {
```
mahrud: Same here. If webCrypto and nodeCrypto aren't used they can be deleted above.

src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -46,7 +46,7 @@ function buildEcdhParam(public_algo, oid, cipher_algo, hash_algo, fingerprint) {
     new Uint8Array([public_algo]),
     kdf_params.write(),
     util.str_to_Uint8Array("Anonymous Sender    "),
-    fingerprint
+    fingerprint.subarray(0, 20)
```
mahrud: Wait, is this supposed to be the first 20 bytes of the fingerprint or the first 10 bytes? If the input is a hex string here then this takes the first 10 bytes. Right?

twiss: `fingerprint` is an `Uint8Array` here.

PR #705, opened 2018-05-11
Fix IE11 support

mmso:
With IE11 it would throw a `DataCloneError` when using `postMessage` to the service worker. This was because it was sending an array as transferable which IE11 does not support.

https://developer.mozilla.org/en-US/docs/Web/API/Transferable
[1] Internet Explorer 10 only accepts a single Transferable object as a parameter, but not an array.

~~Also, I had to change the actual function call. IE11 still threw the error if I returned `undefined` in `getTransferables`.~~

2) After I fixed that I noticed that an error was thrown in `PublicKeyEncryptedSessionKey.prototype.decrypt` that `Promise` was undefined. So to fix that I always require the Promise polyfills.

### DIFF
--- a/src/polyfills.js
+++ b/src/polyfills.js
@@ -15,9 +15,10 @@ if (typeof Array.prototype.find === 'undefined') {
 if (typeof Array.from === 'undefined') {
   require('core-js/fn/array/from');
 }
-if (typeof Promise === 'undefined') {
-  require('core-js/fn/promise');
-}
+
+// No if-statement on Promise because of IE11. Otherwise Promise is undefined in the service worker.
+require('core-js/fn/promise');
+
 if (typeof Uint8Array.from === 'undefined') {
   require('core-js/fn/typed/uint8-array');
 }
--- a/src/util.js
+++ b/src/util.js
@@ -30,8 +30,9 @@ import config from './config';
 import util from './util'; // re-import module to access util functions
 import b64 from './encoding/base64';
 
-export default {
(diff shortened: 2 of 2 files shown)


bartbutler:
Some of the polyfills aren't core-js (whatwg-fetch) and in general I think it's better to let the parent application choose which polyfills to use and only supply them in the case that the parent application doesn't. Are the polyfill-related changes necessary? I would think Promise would get polyfilled automatically with the existing code.

mmso:
I agree with you on that. I removed the if-statements because the core-js polyfills and whatwg-fetch polyfill already check in their own libraries if they exist or not.

I've removed my changes to the polyfill file. However, the Promise problem remains in IE11.

I've tracked it down even further to only be coming from this line: https://github.com/openpgpjs/openpgpjs/blob/master/src/worker/worker.js#L49

Promise is defined in IE11, but not in the WebWorker context. I have no idea why and I am not sure why the core-js/promise polyfill doesn't get evaluated after the `importScripts` has executed.

I see two solutions, either remove the dependency on Promise on randomBuffer init for the worker, or force include the core-js/promise polyfill (that apparently seems to make it work too). What do you think?

mmso:
@bartbutler I added that and it fixes it. However that means that `openpgp.worker.js` needs to be babelified for `require` to work, and the `core-js/promise` polyfills seem to be quite large so the minified file size ends up being `24K` instead of `1.6K`. Which is probably not what we want right?

 I opt for removing this if-statement: https://github.com/openpgpjs/openpgpjs/blob/master/src/polyfills.js#L18

It doesn't change the file size and it doesn't change any behavior in openpgp.js either because the core-js polyfills checks if Promise is defined natively or not before injecting itself.

bartbutler:
OK, we can do that.

PR #708, opened 2018-05-17
Run tests with randomly generated strings

MaximilianKrambach:
I'm working on introducing random data into the tests, in order to ease fuzzy tests.

As a first step, I replaced all 'Hello World' and similar with some pseudorandom utf-8 generation.

The downside is that the string generation itself takes a bit of time, thus many of these tests get slightly worse in performance.


sanjanarajan:
Would you mind using camelCase for all of your variable names? Otherwise, looks good to me!

PR #709, opened 2018-05-18
WIP: Streaming

twiss:
Yet another streaming PR. Compared to #567, this one is based on OpenPGP V3 and aims to be a bit more integrated into the codebase.

This PR implements streaming encryption and decryption using two approaches:

- Streaming CFB

  When encrypting a ReadableStream, encrypted data is written to the
  returned stream in chunks of 512 bytes.

  When decrypting a ReadableStream, by default, decryption will start
  immediately, but data will only be written to the returned stream
  after the integrity tag at the end of the packet has been validated.

  This is because using unauthenticated data is very prone to security
  issues, as e.g. the recent https://efail.de/ vulnerability shows. For
  example, when displaying unauthenticated HTML, an attacker could read
  the decrypted plaintext by inserting `<img src="https://efail.com/` at
  the beginning of the ciphertext. Similarly, when passing
  unauthenticated data to any other code or system (e.g. a video codec)
  an attacker might be able to exploit bugs in that system.

  Also, when using unauthenticated data, it is critical that the sender
  of the data not get access to the decrypted version. For example, if
  you present an unauthenticated email to a user, and they reply to it,
  the sender could read a third party's email.

  If you know what you're doing and can guarantee that none of the above
  will happen, you can enable reading unauthenticated data on the stream
  by setting `openpgp.config.allow_unauthenticated_stream = true`.

- Streaming AEAD
  
  When the new draft is enabled (`openpgp.config.aead_protect = true`),
  we simulate streaming encryption using chunks. The chunk size is `2 **
  (openpgp.config.aead_chunk_size_byte + 6)` bytes. The default
  chunk_size_byte is 12, so the default chunk_size is 256KiB. If you
  want to receive data on the stream more often, you should lower
  `openpgp.config.aead_chunk_size_byte`.

  Since every chunk has its own integrity tag, you will always receive
  chunks as soon as they are available when decrypting, and it is safe
  to operate on them as long as you can deal with unexpected truncation
  (i.e. unexpected end of the stream).

## API changes

- `message.fromText()`, `fromBinary()`, `readArmored()` and `read()` now accept
  ReadableStreams.

- `openpgp.encrypt()` and `openpgp.sign()` now take a `message` parameter instead of
  `data`, `dataType` and `filename`. Use `openpgp.message.fromText()`, `fromBinary()` or `new openpgp.cleartext.CleartextMessage()` to create a message. Note that if you previously used the `date` parameter, you should now additionally pass it to `fromText`/`fromBinary` as well.

- High-level `openpgp.*` functions now have an `streaming` parameter, to
  control whether the return value is a ReadableStream. It defaults to
  true iff the message you pass was created from a stream.

- all `readArmored()` functions are now asynchronous. For example, instead of writing
  ```js
  let publicKey = openpgp.key.readArmored(publicKeyArmored).keys[0];
  ```
  write:
  ```js
  let publicKey = (await openpgp.key.readArmored(publicKeyArmored)).keys[0];
  ```

- after calling `let keyring = new openpgp.Keyring()`, you now have to
  call `await keyring.load()` to read the keys from LocalStorage.

### Writing stream data

To accommodate writing (packet) data containing streams, there's a new
`util.concat()` function with the following signature:

- Array[Uint8Array] -> Uint8Array
- Array[Uint8Array | ReadableStream] -> ReadableStream[Uint8Array]
- Array[String] -> String
- Array[String | ReadableStream] -> ReadableStream[String]

For packets that support partial lengths, `packet.write()` may now
return a ReadableStream which `packetlist.write()` will encode.

The reason `util.concat()` doesn't return a ReadableStream if you don't
pass any is to keep the overhead low when not using streams.

Alternatively, there's `stream.concat()` which always returns a
ReadableStream, even if you only pass Uint8Arrays or Strings.

### Reading stream data

There's a new API you can use to read (packet) data from either a stream or a
Uint8Array:

```js
const reader = stream.getReader(streamOrArray);
while (true) {
  const { done, value } = await reader.read(); // { done: false, value: Uint8Array } or { done: true, value: undefined }
  if (done) break;
  // Do something with value
}
```

There are also helper functions for reading a specific number of bytes, or a single line, etc:

```js
stream.parse(streamOrArray, reader => {
  const byte = await reader.readByte(); // Single byte or undefined
  const bytes = await reader.readBytes(n); // Uint8Array of up to n bytes, or undefined
  const line = await reader.readLine(); // Returns String up to and including the first \n, or undefined. This function is specifically for a ReadableStream[String].
  // There's also peekBytes() and unshift(), which you can use to look ahead in the stream.

  const stream = reader.remainder(); // New stream containing the remainder of the original stream. Only available when using a Reader from stream.parse()
});
```

Most of the functions above are also available when getting a reader using `stream.getReader()` instead of `stream.parse()`.

All of the functions above also work when reading a `ReadableStream[String]` instead of a `ReadableStream[Uint8Array]`, and will return Strings instead of Uint8Arrays in that case.

There are also a few functions not for reading the stream, but for
manipulating the stream for another function to read:

```js
stream.slice(streamOrArray, begin, end); // Returns a stream pointing to part of the original stream, or a Uint8Array
stream.clone(streamOrArray); // Returns a copy of the stream so that two functions can read it. Note: this does *not* clone a Uint8Array, since this function is only meant for reading the same data twice.
stream.passiveClone(streamOrArray); // Also returns a copy of the stream, but doesn't return data immediately when you read from it, only returns data when you read from the original stream. This is meant for respecting backpressure.
```

### Transforming data

To return a stream which contains the data of another stream, but
transformed in some way (e.g. en/decrypted, (de)compressed, hashed, etc)
there's a helper function:

```js
const encrypt = new Encrypt();
const transformed = stream.transform(streamOrArray, function process(value) {
  return encrypt.process(value);
}, function finish() {
  return encrypt.finish();
});
```

Returning data from both functions is optional. Both `process` and
`finish` are optional, and default to not returning any data.

Similarly to `util.concat()`, `stream.transform()` returns a Uint8Array
if you pass one. This has one caveat: for transforming streams, you can
pass asynchronous functions as `process` and `finish`. When transforming
a Uint8Array that's not possible, since the function has to return a
Uint8Array immediately. If you need the asynchronous behavior, convert
the array to a stream first (using `stream.toStream(array)`).

### DIFF
--- a/src/crypto/public_key/elliptic/ecdh.js
+++ b/src/crypto/public_key/elliptic/ecdh.js
@@ -67,7 +67,7 @@ function kdf(hash_algo, X, length, param) {
  * @param  {module:type/mpi}        m            Value derived from session key (RFC 6637)
  * @param  {Uint8Array}             Q            Recipient public key
  * @param  {String}                 fingerprint  Recipient fingerprint
- * @returns {{V: BN, C: BN}}                     Returns ephemeral key and encoded session key
+ * @returns {Promise<{V: BN, C: BN}>}            Returns ephemeral key and encoded session key
  * @async
  */
 async function encrypt(oid, cipher_algo, hash_algo, m, Q, fingerprint) {
@@ -95,7 +95,7 @@ async function encrypt(oid, cipher_algo, hash_algo, m, Q, fingerprint) {
  * @param  {Uint8Array}             C            Encrypted and wrapped value derived from session key
  * @param  {Uint8Array}             d            Recipient private key
  * @param  {String}                 fingerprint  Recipient fingerprint
- * @returns {Uint8Array}                         Value derived from session
+ * @returns {Promise<Uint8Array>}                Value derived from session
  * @async
  */
 async function decrypt(oid, cipher_algo, hash_algo, V, C, d, fingerprint) {
--- a/src/encoding/armor.js
+++ b/src/encoding/armor.js
@@ -196,7 +196,7 @@ function verifyHeaders(headers) {
  * DeArmor an OpenPGP armored message; verify the checksum and return
  * the encoded bytes
  * @param {String} text OpenPGP armored message
- * @returns {Object} An object with attribute "text" containing the message text,
+ * @returns {Promise<Object>} An object with attribute "text" containing the message text,
  * an attribute "data" containing a stream of bytes and "type" for the ASCII armor type
  * @async
  * @static
--- a/src/key.js
+++ b/src/key.js
@@ -37,6 +37,11 @@ import util from './util';
  * @classdesc Class that represents an OpenPGP key. Must contain a primary key.
  * Can contain additional subkeys, signatures, user ids, user attributes.
  * @param  {module:packet.List} packetlist The packets that form this key
+ * @borrows module:packet.PublicKey#getKeyId as Key#getKeyId
+ * @borrows module:packet.PublicKey#getFingerprint as Key#getFingerprint
+ * @borrows module:packet.PublicKey#getAlgorithmInfo as Key#getAlgorithmInfo
+ * @borrows module:packet.PublicKey#getCreationTime as Key#getCreationTime
+ * @borrows module:packet.PublicKey#isDecrypted as Key#isDecrypted
  */
 export function Key(packetlist) {
   if (!(this instanceof Key)) {
@@ -152,10 +157,10 @@ Key.prototype.toPacketlist = function() {
 };
 
 /**
- * Returns packetlist containing all public or private subkey packets matching keyId;
- * If keyId is not present, returns all subkey packets.
+ * Returns an array containing all public or private subkeys matching keyId;
+ * If keyId is not present, returns all subkeys.
  * @param  {type/keyid} keyId
- * @returns {Array<SubKey>}
+ * @returns {Array<module:key~SubKey>}
  */
 Key.prototype.getSubkeys = function(keyId=null) {
   const subKeys = [];
@@ -168,10 +173,10 @@ Key.prototype.getSubkeys = function(keyId=null) {
 };
 
 /**
- * Returns a packetlist containing all public or private key packets matching keyId.
- * If keyId is not present, returns all key packets starting with the primary key.
+ * Returns an array containing all public or private keys matching keyId.
+ * If keyId is not present, returns all keys starting with the primary key.
  * @param  {type/keyid} keyId
- * @returns {Array<Key|SubKey>}
+ * @returns {Array<module:key.Key|module:key~SubKey>}
  */
 Key.prototype.getKeys = function(keyId=null) {
   const keys = [];
@@ -182,7 +187,7 @@ Key.prototype.getKeys = function(keyId=null) {
 };
 
 /**
- * Returns key IDs of all key packets
+ * Returns key IDs of all keys
  * @returns {Array<module:type/keyid>}
  */
 Key.prototype.getKeyIds = function() {
@@ -287,7 +292,7 @@ function isValidSigningKeyPacket(keyPacket, signature, date=new Date()) {
  * @param  {module:type/keyid} keyId, optional
  * @param  {Date} date use the given date for verification instead of the current time
  * @param  {Object} userId, optional user ID
- * @returns {Promise<Key|SubKey|null>} key or null if no signing key has been found
+ * @returns {Promise<module:key.Key|module:key~SubKey|null>} key or null if no signing key has been found
  * @async
  */
 Key.prototype.getSigningKey = async function (keyId=null, date=new Date(), userId={}) {
@@ -330,7 +335,7 @@ function isValidEncryptionKeyPacket(keyPacket, signature, date=new Date()) {
  * @param  {module:type/keyid} keyId, optional
  * @param  {Date}              date, optional
  * @param  {String}            userId, optional
- * @returns {Promise<Key|SubKey|null>} key or null if no encryption key has been found
+ * @returns {Promise<module:key.Key|module:key~SubKey|null>} key or null if no encryption key has been found
  * @async
  */
 Key.prototype.getEncryptionKey = async function(keyId, date=new Date(), userId={}) {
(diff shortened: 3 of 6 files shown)


cedral:
I cannot figure out how to get a readable stream that returns the pgp encrypted data.
var msg = openpgp.message.fromBinary(new Response(file.file).body, file.name);
```javascript
    var msg = openpgp.message.fromBinary(new Response(file.file).body, file.name);   
    message_stream = await openpgp.encrypt({
        message: msg
      , publicKeys: pub_keys
      , armor: false
      , asStream: true
    })
```
despite what your documentation says message_stream is not a stream it is a Message object comprised of 2 packets. I seem to be on the right track because the second packet's body is a ReadableStream. How do I convert the message object to a stream so I can read from it and cause the actual encryption to occur? The idea is that I will do something like...
 ```javascript
message_stream.read(bytes).then(result => {
            var bytedata = new Blob([result.value]);
            chunk.readFinished(bytedata); //this is a callback to flow.js to say that a chunk of the file has been read.
        });
```

**Comment by cedral** [comment]
The only thing that looks remotely intentional on your part is 
```javascript
message_reader = Result.message.packets
```
That provides a read function but it crashes. Am I missing something or is it just horribly broken.

cedral:
I figured out how to get the header.
```javascript
openpgp.stream.getReader(Result.message.packets[0].write())
```
this gets me a reader from which the header can be read.
but if I try the same thing with packets[1] the read function returns a value of 1 byte and done false.

cedral:
That's the one byte on the front. The write function returns 1 byte concatenated with a stream. I must need to something to populate that stream.

twiss:
Yeah, the documentation kinda assumes that you don't set `armor: false`. To get a stream of bytes of all packets, you can do:

```js
const reader = openpgp.stream.getReader(Result.message.packets.write());
```

Also, you need to call `reader.read()` repeatedly. For example, inside an async function:

```js
while (true) {
  const { done, value } = await reader.read();
  if (done) break;
  var bytedata = new Blob([value]);
  chunk.readFinished(bytedata);
}
```

cedral:
works like a charm. Thanks for getting back to me so quick. you really pulled my bacon out of the fire.

hbarylskyi:
Hi! Thank you for your work, works nicely for me! 

Our code needs to use openpgp.message.read function from message.js, declared on line 698:

```
async function read(input, fromStream) {
  const packetlist = new packet.List();
  await packetlist.read(input);
  const message = new Message(packetlist);
  message.fromStream = fromStream;
  return message;
}
```

In your stream branch, this function is not exported, while it is exported in openpgpjs's master branch. Because of that openpgp.message.read is undefined. Could you also export it, so this function is still available? Like this:

```
export async function read(input, fromStream) {
  const packetlist = new packet.List();
  await packetlist.read(input);
  const message = new Message(packetlist);
  message.fromStream = fromStream;
  return message;
}
```

twiss:
Whoops, thanks for catching that! Should be fixed now.

hbarylskyi:
I have encountered a problem when decrypting a pgp message via stream. Decryption works, but openpgp.decrypt does not resolve until the whole encrypted data stream is read. I think this is not intended behavior, right? openpgp.decrypt should resolve with a ReadableStream object which can be processed at the same time as encrypted data is being read.

My code looks like this:

```
      const encryptedStream = new ReadableStream({...});

      const message = await openpgp.message.readArmored(encryptedStream);

      // at this point all chunks from encryptedStream will be read one after another
      const decryptedStream = await openpgp.decrypt({
        message,
        passwords: [password]
      });

      // does not get here until whole encryptedStream is read

      // handle decryptedStream...
```

Also, I was unable to create a build of your stream branch at the latest commit (8a34c6b). When I run `npm test`, I'm getting an error:

> Error: Cannot find module 'streams' from '/Users/hbarylskyi/ssd/openpgpjs/src'

I think this was introduced after 175df56

twiss:
What's supposed to happen is that the function returns early, but by default no data is written to the output stream until the whole input stream is read. For the reason why, please see the explanation under "Streaming CFB" in the original post. Could you confirm that's what's happening?

The OP also contains instructions on how to fix it. In short: if you don't need interoperability with other PGP software, enable AEAD, otherwise, if it's secure for your application, enable reading unauthenticated data.

> Error: Cannot find module 'streams' from '/Users/hbarylskyi/ssd/openpgpjs/src'

Should be fixed now :)

tomholub:
Thanks for maintaining this PR! And the tests. Really looking forward to get a chance to test this.

hbarylskyi:
@twiss Thanks again. Yes, that's what was happening to me. I've switched openpgp.config.allow_unauthenticated_stream to true and everything worked as expected, stream was returned immediately. 

I've also tried setting openpgp.config.aead_protect to true, but that did not help.

twiss:
For AEAD, the default chunk size is kinda high. If you set `openpgp.config.aead_chunk_size_byte = 0` data should arrive faster.

> The chunk size is `2 ** (openpgp.config.aead_chunk_size_byte + 6)` bytes. The default chunk_size_byte is 12, so the default chunk_size is 256KiB. If you want to receive data on the stream more often, you should lower `openpgp.config.aead_chunk_size_byte`.

hbarylskyi:
Of course, I did not think of that. Will give it a try, thanks!

justinmchase:
@hbarylskyi Can you give an updated example of what you're doing? I also need to decrypt a stream without loading the entire document into memory first.

Thank you all for working on this PR, I have been doing something hacky but the document I am working with has finally exceeded my memory size and so I really need to use a stream and this is just in time.

What is the status of this PR being stable and released?

sanjanarajan:
@justinmchase we are hoping to merge and release this today! Just finalizing documentation at the moment.

tomholub:
That sounds great. Note that the new Promises on `readArmored` are breaking changes, that would normally warrant a major version bump - v4, if OpenPGP.js aims to do proper semantic versioning.

Did `openpgp.key.read` and similar stay synchronous?

bartbutler:
@tomholub It's going to be a major version bump, v4.

PR #714, opened 2018-05-25
Add support for Web Key Directory lookups

wiktor-k:
[Web Key Directory is a draft](https://datatracker.ietf.org/doc/draft-koch-openpgp-webkey-service/) for a new key discovery scheme (implemented e.g. in new gnupg). It is similar to HKP but instead of querying the keyserver it queries the domain from the UID's e-mail address. The difference is that WKS returns binary data and HKP ASCII-armored.

For example, when looking for key for `Joe.Doe@Example.ORG` the key would be fetched from `https://example.org/.well-known/openpgpkey/hu/iy9q119eutrkn8s1mk4r39qejnbu3n5q`.

Several services (like kernel.org) already have WKS support, for example Linus' key is at: https://kernel.org/.well-known/openpgpkey/hu/pf113mfnx1f3eb1yiwhsipa91xfc7o4x. Some third party services like posteo.de [support WKS too](https://posteo.de/.well-known/openpgpkey/). WKS is also [supported by Enigmail](https://www.gnupg.org/blog/20170803-web-key-in-enigmail.html).

More details here (in German): https://www.golem.de/news/web-key-service-openpgp-schluessel-ueber-https-verteilen-1609-123194.html

This PR adds support for lookups using Web Key Directory in a similar fashion to already existing HKP lookups keeping the same API as HKP lookup (with differences in `options` object and returned key format).

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "compressjs": "github:openpgpjs/compressjs",
     "elliptic": "github:openpgpjs/elliptic",
     "hash.js": "^1.1.3",
-    "node-fetch": "^1.7.3",
+    "node-fetch": "^2.1.2",
     "node-localstorage": "~1.3.0",
     "pako": "^1.0.6",
     "rusha": "^0.8.12"
--- a/src/util.js
+++ b/src/util.js
@@ -620,5 +620,41 @@ export default {
    */
   removeTrailingSpaces: function(text) {
     return text.replace(/[ \t]+$/mg, "");
+  },
+
+  /**
+   * Encode input buffer using Z-Base32 encoding.
+   * See: https://tools.ietf.org/html/rfc6189#section-5.1.6
+   *
+   * @param {Uint8Array} data The binary data to encode
+   * @returns {String} Binary data encoded using Z-Base32
+   */
+  encodeZBase32: function(data) {
+    if (data.length === 0) {
+      return "";
+    }
+    const ALPHABET = "ybndrfg8ejkmcpqxot1uwisza345h769";
+    const SHIFT = 5;
+    const MASK = 31;
+    let buffer = data[0];
+    let index = 1;
+    let bitsLeft = 8;
+    let result = '';
+    while (bitsLeft > 0 || index < data.length) {
+      if (bitsLeft < SHIFT) {
+        if (index < data.length) {
+          buffer <<= 8;
+          buffer |= data[index++] & 0xff;
+          bitsLeft += 8;
+        } else {
+          const pad = SHIFT - bitsLeft;
+          buffer <<= pad;
+          bitsLeft += pad;
+        }
+      }
+      bitsLeft -= SHIFT;
+      result += ALPHABET[MASK & (buffer >> bitsLeft)];
+    }
+    return result;
   }
 };
--- a/src/wkd.js
+++ b/src/wkd.js
@@ -1,0 +1,77 @@
+// OpenPGP.js - An OpenPGP implementation in javascript
+// Copyright (C) 2018 Wiktor Kwapisiewicz
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 3.0 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+/**
+ * @fileoverview This class implements a client for the Web Key Directory (wkd) protocol
+ * in order to lookup keys on designated servers.
+ * See: https://datatracker.ietf.org/doc/draft-koch-openpgp-webkey-service/
+ * @module wkd
+ */
+
+import util from './util';
+import crypto from './crypto';
+import * as keyMod from './key';
+
+/**
+ * Initialize the WKD client
+ * @constructor
+ */
+function WKD() {
+  this._fetch = typeof window !== 'undefined' ? window.fetch : require('node-fetch');
+}
+
+/**
+ * Search for a public key using Web Key Directory protocol.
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
     "compressjs": "github:openpgpjs/compressjs",
     "elliptic": "github:openpgpjs/elliptic",
     "hash.js": "^1.1.3",
-    "node-fetch": "^1.7.3",
+    "node-fetch": "^2.1.2",
```
wiktor-k: The change here was required to have `arrayBuffer` function on fetch response objects. It is available only for node-fetch version >2.

src/util.js
```diff
@@ -620,5 +620,41 @@ export default {
...
+  },
+
+  /**
+   * Encode input buffer using Z-Base32 encoding.
+   * See: https://tools.ietf.org/html/rfc6189#section-5.1.6
+   *
+   * @param {Uint8Array} data The binary data to encode
+   * @returns {String} Binary data encoded using Z-Base32
+   */
+  encode_zbase32: function(data) {
```
wiktor-k: Is snake case OK here? Some functions in this file use camelCase some snake_case.

src/wkd.js
```diff
@@ -0,0 +1,69 @@
...
+  this._fetch = typeof window !== 'undefined' ? window.fetch : require('node-fetch');
+}
+
+/**
+ * Search for a public key using Web Key Directory protocol.
+ * @param  {String}   options.email   User's email.
+ * @returns {Promise<Uint8Array>}          The public key.
+ * @async
+ */
+WKD.prototype.lookup = function(options) {
```
wiktor-k: WKD does not have any parameters so `lookup` could be a static function, I left it like that so it's similar to HKP...

sanjanarajan:
Thank you for your contribution! 

Some comments:

1) Some of the older code uses snake case variables, but we are moving towards camel case. So ideally the new util functions should be modified to use camel case as well.

2) Your test cases fail on all browsers because the URLs you are fetching from have no 'Access-Control-Allow-Origin' headers set. Could you use some test cases that have 'Access-Control-Allow-Origin' set to *?

3) Why not return a public key object rather than the raw binary data?

4) Are you sure all of those changes to npm-shrinkwrap are necessary? I modified my own copy of package.json with the updated node-fetch version, then ran npm-shrinkwrap, and found far fewer changes.

tomholub:
In general, it may still be useful to return or otherwise have access to raw data. Say OpenPGP.js cannot understand the key due to compatibility issue, we may still want to store it for later debugging. Maybe return an object with both.

src/util.js
twiss: I know this function isn't that big, but perhaps it would be an idea to move it to a dedicated `encoding/zbase32.js` file? Since only one file needs it I don't think it needs to be in util

src/util.js
```diff
@@ -620,5 +620,41 @@ export default {
...
+    }
+    const ALPHABET = "ybndrfg8ejkmcpqxot1uwisza345h769";
+    const SHIFT = 5;
+    const MASK = 31;
+    let buffer = data[0];
+    let index = 1;
+    let bitsLeft = 8;
+    let result = '';
+    while (bitsLeft > 0 || index < data.length) {
+        if (bitsLeft < SHIFT) {
```
twiss: Tiny nitpick: two space indentation

sanjanarajan:
@tomholub what do you think of returning just a public key object in case of success but including the raw data in the error object in case of failure?

tomholub:
I'm not sure, I would imagine that someone may be interested in the raw data even on success scenario.

wiktor-k:
Hi all, thank you excellent feedback!

Thank you for your contribution!

>    Some of the older code uses snake case variables, but we are moving towards camel case. So ideally the new util functions should be modified to use camel case as well.

Will do!

>    Your test cases fail on all browsers because the URLs you are fetching from have no 'Access-Control-Allow-Origin' headers set. Could you use some test cases that have 'Access-Control-Allow-Origin' set to *?

Yes, that's a good idea!

>    Why not return a public key object rather than the raw binary data?

This is just to preserve the same interface as HKP, see here for example:

https://github.com/openpgpjs/openpgpjs#lookup-public-key-on-hkp-server

Personally I would return a public key object as I assume most use-cases the client would want to use the key immediately. Maybe add an option that modifies return type? e.g. `rawKey: true` would return bytes but `rawKey: false` or nothing would by default return public key object?

>    Are you sure all of those changes to npm-shrinkwrap are necessary? I modified my own copy of package.json with the updated node-fetch version, then ran npm-shrinkwrap, and found far fewer changes.

Nope I'm not sure, that's what my npm did, I will try to reduce them to just the things necessary. Thanks for checking this out!

> I know this function isn't that big, but perhaps it would be an idea to move it to a dedicated encoding/zbase32.js file? Since only one file needs it I don't think it needs to be in util

I'm pretty sure zbase32 will be used only in this case but I also thought about making it a separate file, I don't mind either way...

tomholub:
I like the option that modifies the return type, I think that will make it
flexible enough.

wiktor-k:
Okay, phew, I think I've addressed all points here.

There is new options parameter: `rawBytes` - when set to `true` it will return `UInt8Array`, otherwise public key, as parsed by `key.read()`.

I've left ZBase32 in utils as it is small, if you still think it's necessary to extract just tell me.

> Your test cases fail on all browsers because the URLs you are fetching from have no 'Access-Control-Allow-Origin' headers set. Could you use some test cases that have 'Access-Control-Allow-Origin' set to *?

I've set the headers as can be seen via curl:

```
$ curl -I https://metacode.biz/.well-known/openpgpkey/hu/4hg7tescnttreaouu4z1izeuuyibwww1
HTTP/2 200
date: Tue, 29 May 2018 11:00:40 GMT
content-type: application/pgp-keys
content-length: 1244
access-control-allow-origin: *
accept-ranges: bytes
```

But the tests still fail :(

I tried following saucelabs links and looking for detailed logs there but couldn't find them.

The tests also doesn't print anything useful on localhost:

```
$ grunt saucelabs-mocha:all
Running "saucelabs-mocha:all" (saucelabs-mocha) task
=> Starting Tunnel to Sauce Labs
>>
```

@sanjanarajan could you help me with debugging this issue further? Thank you very much for your time!

sanjanarajan:
@wiktor-k - now the only test that is failing is the one called "by email address should not find a key" because the link does not go to a page with the Access-Control-Allow-Origin headers set to *

image
To run the browser tests yourself, use the command `grunt browsertest` and navigate to http://localhost:3000/test/unittests.html in the browser of your choice :)

PR #715, opened 2018-05-31
Use public view of elliptic curve repository.

dkg:
Using ssh to access the elliptic curve repository requires a client
that has ssh access to github, since github doesn't grant anonymous
ssh access.

Using the public view should resolve this problem.

### DIFF
--- a/dist/openpgp.js
+++ b/dist/openpgp.js
@@ -16947,7 +16947,7 @@ module.exports={
   "name": "elliptic",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


twiss:
This PR changes the (compiled) dist file; which gets regenerated on every build - I fail to see how that should help. I imagine the fix for this should involve changing
https://github.com/openpgpjs/openpgpjs/blob/038d8466fecb4299d21b68bd280ba419acede2f1/package.json#L81
but as there's no mention of ssh there I'm not sure how. Another PR fixing it there is welcome, though.

PR #723, opened 2018-06-27
Fix reading signersUserId packet in signatures

wiktor-k:
This bug caused all signersUserIds strings to be prefixed with `"null"`
string. Changed to use only the last value for this packet type.

Previous implementation probably assumed that there will be more than one
signersUserId packet but I haven't been able to generate such signature
using gpg (only last user id was embedded). Moreover signature
serialization function `write_all_sub_packets` writes only one value of
this packet as a UTF-8 string.


sanjanarajan:
good catch :)

PR #729, opened 2018-07-09
Add key revocation option

MaximilianKrambach:
Fixes #465 

Added a method for Keys to be revoked. It is still not fully merge-ready, as I don't yet fully understand how to handle errors in openpgpjs; I'd appreciate some pointers.

For subkey revocations I have a similar method prepared, but I don't fully grasp get what constitutes a 'subkey'. I'll be reading it up.

PR #730, opened 2018-07-11
revoke key and subkey methods added

### LINKED ISSUE #465: revoking a public key
Is this possible? I see in openpgp.js

> // TODO
> Key.prototype.revoke = function () {};

so I guess not, 
would be great though if we could create a revocation certificate so a public key could be revoked.

snowyu:
Fixes #465, #729


sanjanarajan:
Closing this as #465 was fixed in #734

PR #732, opened 2018-07-12
Add the generateSubkey and addSubkey to a key

snowyu:
* add the Key.prototype.generateSubkey method to generate subkey and add it to the key.
* add the Key.prototype.addSubkey method to add subkey packet to the key.


snowyu:
Please review it again.

src/key.js
```diff
@@ -771,6 +771,118 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+ * @returns {Promise<module:key.SubKey>} return the subkey if successful.
+ * @async
+ */
+Key.prototype.generateSubkey = async function(options){
+  options = sanitizeKeyOption(options, this.primaryKey.algorithm);
+  const secretSubKeyPacket = await generateSecretSubkey(options);
+  const result = await this.addSubkey(secretSubKeyPacket, options);
+  return result;
+
+  function sanitizeKeyOption(options, algo) {
```
sanjanarajan: This is basically a duplicate of https://github.com/openpgpjs/openpgpjs/blob/master/src/key.js#L1228. Rather than duplicating this code, I suggest that you pull out the `sanitizeKeyOptions` function from key generation and use it here as well, with any necessary modifications.

snowyu: The `sanitizeKeyOptions` of  the subkey generation should try to guess the curve or rsa from the primaryKey. But the main key generation would not.

sanjanarajan: Understood. But this means that those sanitization functions should probably be pulled out of their parent functions and decomposed so that you can use parts of them in both cases.

snowyu:
Ok, Reuse it now. But It is not the best performance.

snowyu:
Find another bug in the test of x25519.js.

src/key.js
```diff
@@ -771,6 +771,84 @@ Key.prototype.verifyAllUsers = async function(keys) {
   return results;
 };
 
+/**
+ * generate a new subkey to a Key
+ * @param  {Object} options   (optional) options for creating subkey, e.g. {sign: true, passphrase: '123', algorithm:..., curve:...}
```
sanjanarajan: Could you please enumerate all of the options that are possible in this object in the documentation?

src/key.js
```diff
@@ -771,6 +771,84 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+ * @returns {Promise<module:key.SubKey>} return the subkey if successful.
+ * @async
+ */
+Key.prototype.generateSubkey = async function(options){
+  const defaultOptions = {};
+  if (this.primaryKey.algorithm.indexOf('rsa') === 0) {
+    defaultOptions.numBits = 2048;
+  }
+  else {
+    defaultOptions.curve = enums.curve.ed25519;
```
sanjanarajan: Wouldn't it make more sense to default to the curve used by the primary key rather than always ed25519? And shouldn't the number of bits in the RSA case also default to the number of bits used by the primary key?

src/key.js
```diff
@@ -771,6 +771,84 @@ Key.prototype.verifyAllUsers = async function(keys) {
   return results;
 };
 
+/**
+ * generate a new subkey to a Key
```
sanjanarajan: Should be "Adds a new subkey to a key"

src/key.js
```diff
@@ -771,6 +771,84 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+ * @param  {Object} options   (optional) options for creating subkey, e.g. {sign: true, passphrase: '123', algorithm:..., curve:...}
+ *
+ * @returns {Promise<module:key.SubKey>} return the subkey if successful.
+ * @async
+ */
+Key.prototype.generateSubkey = async function(options){
```
sanjanarajan: I don't think this function name makes it clear that the subkey is not only going to be generated, but also added to the key. So I would suggest combining this function with the one under it, renaming this to "addSubkey" and allowing an optional parameter of a subkey packet. If the packet is present, add the packet. Otherwise, generate it with the options.

src/key.js
```diff
@@ -771,6 +771,84 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+ * @param  {Object} options   (optional) options for subkey, e.g. {sign: true, passphrase: '123'}
+ *
+ * @returns {Promise<module:key.SubKey>} return the subkey if successful.
+ * @async
+ */
+Key.prototype.addSubkey = async function(subkeyPacket, options){
+  if (!this.isPrivate()) {
+    throw new Error("Nothing to addSubkey in a public key");
+  }
+  if (!options) options = {};
```
sanjanarajan: Can use a default parameter in the function arguments instead

src/key.js
```diff
@@ -771,6 +771,84 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+    dataToSign, this, secretKeyPacket,
+    subkeySignatureOpts, options.date, options.userId
+  );
+  if (options.passphrase) {
+    subkeyPacket.clearPrivateParams();
+  }
+  const subKey = new SubKey(subkeyPacket);
+  subKey.bindingSignatures.push(subkeySignaturePacket);
+  this.subKeys.push(subKey);
+  if (needRelocked) {
```
sanjanarajan: This is assuming that the secret key and subkey use the same encryption password. While this is true in most cases, it is not true as a general rule (see the way key generation handles/allows for multiple passphrases for each subkey). 
(This is also relevant to your line 815)

src/key.js
```diff
@@ -771,6 +771,84 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+ * add a subkey packet to a Key
+ * @param {module:packet.PublicSubkey|module:packet.SecretSubkey} the subkey packet to add
+ * @param  {Object} options   (optional) options for subkey, e.g. {sign: true, passphrase: '123'}
+ *
+ * @returns {Promise<module:key.SubKey>} return the subkey if successful.
+ * @async
+ */
+Key.prototype.addSubkey = async function(subkeyPacket, options){
+  if (!this.isPrivate()) {
+    throw new Error("Nothing to addSubkey in a public key");
```
sanjanarajan: "Cannot add a subkey to a public key" would make more sense here

src/key.js
```diff
@@ -771,6 +771,84 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+  options = sanitizeKeyOptions(options, defaultOptions);
+  const secretSubKeyPacket = await generateSecretSubkey(options);
+  const result = await this.addSubkey(secretSubKeyPacket, options);
+  return result;
+};
+
+/**
+ * add a subkey packet to a Key
+ * @param {module:packet.PublicSubkey|module:packet.SecretSubkey} the subkey packet to add
+ * @param  {Object} options   (optional) options for subkey, e.g. {sign: true, passphrase: '123'}
```
sanjanarajan: Again, please enumerate options

src/key.js
snowyu: no problem.

src/key.js
snowyu: I haven't found any methods or properties to get it directly yet.  I will read the code more carefully.  I know RSA well, But I have no idea about curve.

src/key.js
snowyu: Yes, It's used to make me a headache for a while.. I think `generate` and `add` is different meaning. This is why using `openpgp.generateKey` instead of `openpgp.addKey`.

But the `Key.prototype.generateSubkey` instance method which means  "generate the subkey on the key", the subkey relates(belongs) to the key.

The `SubKey.generate` static method which means "generate an independent subkey" if it is exists.

Maybe I should add the `Subkey.generate` method and modify the constructor of Subkey:
generate a subkey if no packet parameter passed.

src/key.js
snowyu: "generate and add a new subkey to a Key"

src/key.js
snowyu: Ok.

src/key.js
snowyu: No.  It just is reuse the `options.passphrase`.  The `needRelocked` only used to relock security key.

* Case 1:  Set the same password to subkey which is your case.
* Case 2：Set the different password to the subkey, not the same with secret key.
  * Case 2.1: It should raise error if the secret key is encrypted.
    * because the password is wrong for the secret key. So the user must decrypt it first.
  * Case 2.2: It will only encrypt the subkey with the password if the secret key is decrypted.
    * because the needRelocked is false: undefined.

src/key.js
snowyu: Ok, But I just follow the error format of the `Key.prototype.encrypt/decrypt`.

src/key.js
snowyu: I found the reason: the first byte of n(same security key) in V5 key is not the same in the V4 key.
So I can use the byteLength to check instead of BitLength.

snowyu:
Ok, almost done. review it.

test/general/key.js
```diff
@@ -2193,4 +2193,90 @@ VYGdb3eNlV8CfoEC
...
+    const privateKey = openpgp.key.readArmored(priv_key_rsa).keys[0];
+    const total = privateKey.subKeys.length;
+    return privateKey.primaryKey.decrypt('hello world').then(function(){
+      return privateKey.generateSubkey({passphrase: 'otherpass'});
+    }).then(function(answer){
+      expect(answer).to.exist;
+      expect(privateKey.subKeys.length).to.be.equal(total+1);
+      const subkeyN = answer.subKey.params[0];
+      const pkN = privateKey.primaryKey.params[0];
+      // console.log(subkeyN.data[0], pkN.data[0], openpgp.util.nbits(subkeyN.data[0]), openpgp.util.nbits(pkN.data[0]))
```
sanjanarajan: Remove this please

test/general/key.js
```diff
@@ -2193,4 +2193,90 @@ VYGdb3eNlV8CfoEC
...
+        expect(answer.subKey.isDecrypted).to.be.true;
+        return answer;
+      })
+    }).then(function(answer){
+      return answer.verify(privateKey.primaryKey);
+    }).then(function(answer){
+      expect(answer).to.be.equal(openpgp.enums.keyStatus.valid);
+    });
+  });
+
```
sanjanarajan: Would be good to have tests that verify that the new keys can encrypt/sign/decrypt/verify as expected

src/key.js
```diff
@@ -771,6 +771,92 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+ * @returns {Promise<module:key.SubKey>} return the subkey if successful.
+ * @async
+ */
+Key.prototype.generateSubkey = async function(options){
+  const defaultOptions = {};
+  const defaultAlgo = this.primaryKey.algorithm;
+  if (defaultAlgo.indexOf('rsa') === 0) {
+    defaultOptions.numBits = this.primaryKey.params[0].bitLength();
+  }
+  else if (defaultAlgo.indexOf('ec') === 0 || defaultAlgo.indexOf('ed') === 0) {
```
sanjanarajan: Could you instead check explicitly for ecdsa, eddsa, ecdh? In case there is some conflict in the future with algorithms that coincidentally have these prefixes.

src/key.js
```diff
@@ -996,6 +1082,50 @@ function SubKey(subKeyPacket) {
...
+ * @param {Number} [options.keyExpirationTime=0]
+ *                             The number of seconds after the key creation time that the key expires
+ * @param  {Date} options.date  Override the creation date of the key and the key signatures
+ * @param  {Object} userId                   (optional) user ID
+ *
+ * @returns {Promise<module:packet.Signature>} return the signature if successful.
+ * @async
+ */
+SubKey.prototype.bindSignature = async function(primaryKey, options) {
+  const subkeySignatureOpts = {};
```
sanjanarajan: Please find a way to share this code with wrapKeyObject, as most of it is duplicated from there

src/key.js
```diff
@@ -1030,6 +1116,70 @@ function SubKey(subKeyPacket) {
...
+/**
+ * generate a new SubKey.
+ * @param {Integer} options.numBits (optional) number of bits for the key creation.
+ * @param  {String} options.curve (optional) elliptic curve for ECC keys
+ * @param  {Date} options.date  Override the creation date of the key and the key signatures
+ *
+ * @returns {Promise<module:key.SubKey>} return the subkey if successful.
+ * @async
+ */
+SubKey.generate = async function(options) {
```
twiss: What's the use-case for generating a standalone subkey? Otherwise this can be merged into `addSubkey`.

twiss:
To summarize the above individual comments, in case they're confusing in isolation:

I think we should have two functions:
- `Key.prototype.addSubkey(options) -> Key`
- `createSubkeySignature` (private)

I think that would cover the most common use case of generating and adding a new subkey to a key, and then using that key. Are there any useful cases that that API misses?

src/key.js
snowyu: This question is simulate to what's the use-case for the Subkey Class? 

We abstract the data and functions to a class is used to reuse codes mostly.

snowyu:
First of all, this is not application or commercial service development, this is the encryption algorithm library. The user we are facing is the developer. It should be flexible, simple, and easy to use.
And Never think about the user-case instead of developers(The user scenarios they face are unimaginable). Just think about the class/data/function of the library.

Principle 1: A function(method) has one and only one responsibility to KISS(Keep It Simple and Stupid).
`KISS` make flexible, simple possible.

eg, The `add` and `generate` is different means, so it different responsibility. Purpose the developer wanna add a subkeypacket(coming from others key) to it.

eg for easy to use,  `The generation subkey` should encrypt itself if the passphrase given. And it can not generate subkey if main key is encrypted. The main key's password should be the same as the subkey most of the time. So Why not to decrypt the main key in the time.  

As developer, would you like this:

```ts
  // And it's Inefficient, if will decrypt all subkeys too.
  // if you known the internal you could use this to decrypt key
  // But the internal attribute name could be changed
  // await mainPrivateKey.primaryKey.decrypt('mypasswd'); 
  await mainPrivateKey.decrypt('mypasswd'); 
  const newSubkey = await mainPrivateKey.generateSubkey({passphrase: 'mypasswd'});
  await mainPrivateKey.encrypt('mypasswd'); // relock the main key and it will encrypt all subkeys again.
```

Or this:

```ts
  // if the passphrase is the same as the main key, it will unlock and lock it automatically.
  const newSubkey = await mainPrivateKey.generateSubkey({passphrase: 'mypasswd'});
```

Of cause, more easy means more weak(safety). But it is the developer's own choice.

tomholub:
This approach:
```js
  // if the passphrase is the same as the main key, it will unlock and lock it automatically.
  const newSubkey = await mainPrivateKey.generateSubkey({passphrase: 'mypasswd'});
```
Is not the best, because of the automatic "magic" behavior. For example, what if my key is already decrypted? What if I want my key to stay decrypted to do more operations?

Better to assume developer decrypted it first, and throw error if not:
```js
await prv.addSubkey(); // Error('Cannot add subkey: primary key is encrypted. Decrypt primary key first.');
```
Since as a developer (api user), I had to decrypt the key first, I should know to encrypt it again before saving, just like with the rest of OpenPGP.js API:
```js
await k.decrypt(passphrase);
await k.addSubkey();
await k.encrypt(passphrase);
```
No magic is better then some magic for a security library - my two cents.

snowyu:
Keep the key encrypted in memory is more friendly for beginner. And this `magic` does not prevent you from doing that.  You can still choose your way. Another can choose theirs.

tomholub:
I know. But that's not how the rest of the library works. Consistency is crucial. What if, based on this API, a developer assumes that the rest will do the same? 

It would make sense to discuss that for v4 of the library, but for now, let's keep it consistent.

snowyu:
I agree it.

sanjanarajan:
@snowyu when you have a chance could you please address the rest of the comments from @twiss? Then we can hopefully get this feature into a release soon.

To echo his main points:
1) Helper functions should only be public if there is a reasonable use case for them to be public
2) Variable names should be descriptive, even in tests, because developers need to read tests too
3) Simplify your code whenever possible

snowyu:
I think 

1. The case I've point out. And how to distinguish whether they are helper functions or methods(a responsibility/duty of a class), do you know?  Is it to judge whether they have an user case?
  * `SubKey.bindSignature` and  `SubKey.generate` are the duty of the SUbkey class. not a helper.
  * the SubKey  is not be exported(public). I think is right for it is  not designed to complete.
  * Maybe I am wrong, It is not OO Design. So a method added to a class called `public api`.
2. All right.
3. point it out.
4. great thanks reviews of @sanjanarajan.

Maybe I am of this type that is not welcome.

snowyu:
I think 

1. The case I've point out. And how to distinguish whether they are helper functions or methods(a responsibility/duty of a class), do you know?  Is it to judge whether they have an user case?
  * `SubKey.bindSignature` and  `SubKey.generate` are the duty of the SUbkey class. not a helper.
  * the SubKey  is not be exported(public). I think is right for it is  not designed to complete.
  * Maybe I am wrong, It is not OO Design. So a method added to a class called `public api`.
2. All right.
3. point it out.
4. great thanks reviews of @sanjanarajan.

Maybe I am of this type that is not welcome.

twiss:
1.) I assume you're referring to:

> Purpose the developer wanna add a subkeypacket(coming from others key) to it.

But why would they want to do that?

> And Never think about the user-case instead of developers(The user scenarios they face are unimaginable).

Usually, the way things work in an open source library is that users of the library (i.e. developers) add functionality when they have a use case for it. Adding functionality that nobody uses unnecessarily increases the size of the library and API, and hence maintenance and backwards compatibility work.

> the SubKey is not be exported(public). I think is right for it is not designed to complete.

The subkeys are available from `key.getSubkeys()` and a bunch of other functions. So the instances are public.

> SubKey.bindSignature and SubKey.generate are the duty of the SUbkey class. not a helper.
> Maybe I am wrong, It is not OO Design. So a method added to a class called `public api`.

I agree. This is a side effect of using JavaScript. If this were a full OO language, a private method on the class would be fine. However, in JS there's no straightforward way to do that.

3.) @sanjanarajan is probably referring to the comments I made on the code above, such as https://github.com/openpgpjs/openpgpjs/pull/732#discussion_r205434023. Also, please remove the commented out code and stuff like that. Generally just try to make the patch diff as small & simple as possible.

> Maybe I am of this type that is not welcome.

Sorry for making you feel that way. Generally, in an open source library it's a good idea to have a certain amount of consistency and cohesion. We're basically telling you the way we usually write code for openpgp.js. That's not to say your approach is unquestionably wrong, or ours is better, but it also means that we're not too interested in arguing about every single aspect of our approach. It's easier to change this 300 line patch than the whole of openpgp.js.

sanjanarajan:
Out of date

PR #734, opened 2018-07-13
Key revocation

### LINKED ISSUE #465: revoking a public key

twiss:
Yet another PR joining the party to fix #465. In addition to #730, this adds a high level `openpgp.revokeKey` function, and support for generating and applying standalone revocation certificates.

### DIFF
--- a/src/encoding/armor.js
+++ b/src/encoding/armor.js
@@ -95,15 +95,19 @@ function getType(text) {
  * packet block.
  * @author  Alex
  * @version 2011-12-16
+ * @param {String} customComment (optional) additional comment to add to the armored string
  * @returns {String} The header information
  */
-function addheader() {
+function addheader(customComment) {
   let result = "";
   if (config.show_version) {
     result += "Version: " + config.versionstring + '\r\n';
   }
   if (config.show_comment) {
     result += "Comment: " + config.commentstring + '\r\n';
+  }
+  if (customComment) {
+    result += "Comment: " + customComment + '\r\n';
   }
   result += '\r\n';
   return result;
@@ -326,22 +330,23 @@ function dearmor(text) {
  * @param body
  * @param {Integer} partindex
  * @param {Integer} parttotal
+ * @param {String} customComment (optional) additional comment to add to the armored string
  * @returns {String} Armored text
  * @static
  */
-function armor(messagetype, body, partindex, parttotal) {
+function armor(messagetype, body, partindex, parttotal, customComment) {
   const result = [];
   switch (messagetype) {
     case enums.armor.multipart_section:
       result.push("-----BEGIN PGP MESSAGE, PART " + partindex + "/" + parttotal + "-----\r\n");
-      result.push(addheader());
+      result.push(addheader(customComment));
       result.push(base64.encode(body));
       result.push("\r\n=" + getCheckSum(body) + "\r\n");
       result.push("-----END PGP MESSAGE, PART " + partindex + "/" + parttotal + "-----\r\n");
       break;
     case enums.armor.multipart_last:
       result.push("-----BEGIN PGP MESSAGE, PART " + partindex + "-----\r\n");
-      result.push(addheader());
+      result.push(addheader(customComment));
       result.push(base64.encode(body));
       result.push("\r\n=" + getCheckSum(body) + "\r\n");
       result.push("-----END PGP MESSAGE, PART " + partindex + "-----\r\n");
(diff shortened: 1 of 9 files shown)


src/key.js
```diff
@@ -624,9 +624,83 @@ async function mergeSignatures(source, dest, attr, checkFn) {
...
+/**
+ * Revokes the key
+ * @param  {module:key~Key} privateKey decrypted private key for revocation
+ * @param  {Object} reasonForRevocation optional, object indicating the reason for revocation
+ * @param  {module:enums.reasonForRevocation} reasonForRevocation.flag optional, flag indicating the reason for revocation
+ * @param  {String} reasonForRevocation.string optional, string explaining the reason for revocation
+ * @param  {Date} date optional, override the creationtime of the revocation signature
+ * @return {module:key~Key} new key with revocation signature
+ */
+Key.prototype.revoke = async function(privateKey, {
```
sanjanarajan: Let's just validate the Key object and make sure it's private rather than passing in a private key

src/key.js
```diff
@@ -624,9 +624,83 @@ async function mergeSignatures(source, dest, attr, checkFn) {
...
 
+/**
+ * Get revocation certificate from a revoked key.
+ *   (To get a revocation certificate for an unrevoked key, call revoke() first.)
+ * @return {String} armored revocation certificate
+ */
+Key.prototype.getRevocationCertificate = function() {
+  if (this.revocationSignatures.length) {
+    const commentstring = config.commentstring;
+    config.commentstring = 'This is a revocation certificate';
```
sanjanarajan: If I make a revocation cert in GPG I get two comment lines:
```
Comment: GPGTools - https://gpgtools.org
Comment: This is a revocation certificate
```
So instead of reassigning the comment and doing this try/finally stuff I would just pass a flag indicating the revocation into the armor.encode function.

src/key.js
```diff
@@ -624,9 +624,83 @@ async function mergeSignatures(source, dest, attr, checkFn) {
...
+};
+
+/**
+ * Applies a revocation certificate to a key
+ * @param  {String} revocationCertificate armored revocation certificate
+ * @return {module:key~Key} new revoked key
+ */
+Key.prototype.applyRevocationCertificate = async function(revocationCertificate) {
+  const input = armor.decode(revocationCertificate);
+  if (input.type !== enums.armor.public_key) {
```
sanjanarajan: Are rev certs also allowed to be armored signatures or just armored public keys? I feel like I've seen both

src/key.js
```diff
@@ -624,9 +624,83 @@ async function mergeSignatures(source, dest, attr, checkFn) {
...
+      return armor.encode(enums.armor.public_key, packetlist.write());
+    } finally {
+      // Restore comment string. armor.encode() shouldn't throw, but just to be sure it's wrapped in a try/finally
+      config.commentstring = commentstring;
+    }
+  }
+};
+
+/**
+ * Applies a revocation certificate to a key
```
sanjanarajan: It's probably obvious but just in case I would add a clarifying comment to say that it will only take the first signature from the certificate. Or we can support multiple to be consistent with the `getRevocationCertificate` function :)

src/openpgp.js
```diff
@@ -108,15 +108,16 @@ export function destroyWorker() {
  * @param  {Date} date               (optional) override the creation date of the key and the key signatures
  * @param  {Array<Object>} subkeys   (optional) options for each subkey, default to main key options. e.g. [{sign: true, passphrase: '123'}]
  *                                              sign parameter defaults to false, and indicates whether the subkey should sign rather than encrypt
+ * @param  {Boolean} revocationCertificate (optional) Whether the returned object should include a revocation certificate to revoke the public key
```
sanjanarajan: Probably unnecessary and can always be returned

src/key.js
```diff
@@ -1035,6 +1127,37 @@ SubKey.prototype.update = async function(subKey, primaryKey) {
...
+ * Revokes the subkey
+ * @param  {module:packet/signature} primaryKey primary key used for revocation
+ * @param  {module:key~Key} privateKey decrypted private key for revocation
+ * @param  {Object} reasonForRevocation optional, object indicating the reason for revocation
+ * @param  {module:enums.reasonForRevocation} reasonForRevocation.flag optional, flag indicating the reason for revocation
+ * @param  {String} reasonForRevocation.string optional, string explaining the reason for revocation
+ * @param  {Date} date optional, override the creationtime of the revocation signature
+ * @return {module:key~SubKey} new subkey with revocation signature
+ */
+SubKey.prototype.revoke = async function(primaryKey, privateKey, {
```
sanjanarajan: Same comment as above re: privateKey

src/key.js
```diff
@@ -1415,13 +1553,14 @@ function getExpirationTime(keyPacket, signature) {
 
 /**
  * Returns the preferred signature hash algorithm of a key
- * @param  {object} key
+ * @param  {module:key.Key} (optional) key
+ * @param  {module:packet.SecretKey|module:packet.SecretSubkey} keyPacket
```
sanjanarajan: Can you write a brief explanation of the purpose of the variable? Like "key packet used for signing" or something like that

src/encoding/armor.js
```diff
@@ -329,19 +332,19 @@ function dearmor(text) {
  * @returns {String} Armored text
```
sanjanarajan: Add customComment param to docs :)

src/encoding/armor.js
```diff
@@ -97,14 +97,17 @@ function getType(text) {
  * @version 2011-12-16
  * @returns {String} The header information
```
sanjanarajan: document customComment

PR #742, opened 2018-07-19
Fix verification of User Attributes

wiktor-k:
This change corrects verification of certifications over User Attributes (such as photos).

Before this change the code did not differentiate between User IDs and User Attributes as both of them are [stored in `data.userid`][0] and incorrectly used the User ID constant (0xB4) for both cases.

This change moves the signature certification tag value to User ID and User Attribute classes and uses that in calculation of data for the signature.

[0]: https://github.com/openpgpjs/openpgpjs/blob/11b2d2de3cf39949e5fdfb2b95d027d8329c41fe/src/key.js#L872

### DIFF
--- a/src/packet/signature.js
+++ b/src/packet/signature.js
@@ -594,14 +594,14 @@ Signature.prototype.toSign = function (type, data) {
       let packet;
       let tag;
 
-      if (data.userid !== undefined) {
+      if (data.userId) {
         tag = 0xB4;
-        packet = data.userid;
-      } else if (data.userattribute !== undefined) {
+        packet = data.userId;
+      } else if (data.userAttribute) {
         tag = 0xD1;
-        packet = data.userattribute;
+        packet = data.userAttribute;
       } else {
-        throw new Error('Either a userid or userattribute packet needs to be ' +
+        throw new Error('Either a userId or userAttribute packet needs to be ' +
           'supplied for certification.');
       }
 
--- a/src/key.js
+++ b/src/key.js
@@ -536,7 +536,7 @@ Key.prototype.getPrimaryUser = async function(date=new Date(), userId={}) {
   }).pop();
   const { user, selfCertification: cert } = primaryUser;
   const primaryKey = this.keyPacket;
-  const dataToVerify = { userid: user.userId , key: primaryKey };
+  const dataToVerify = { userId: user.userId, key: primaryKey };
   // skip if certificates is invalid, revoked, or expired
   // eslint-disable-next-line no-await-in-loop
   if (!(cert.verified || await cert.verify(primaryKey, dataToVerify))) {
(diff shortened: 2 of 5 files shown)


sanjanarajan:
@wiktor-k 
This solution works but it seems cleaner and more intuitive to pass around `{ userid, userAttribute }` rather than storing both IDs and attributes in `userId`. Is there a reason you chose not to go this route?

wiktor-k:
Only that the entire change would be bigger as I'd have to adjust all calls to the signature verification method (currently there are multiple places that use `userid: userID || userAttribute`).

I'll take a look at cleaning this up, I just didn't want to modify more code than necessary to fix the bug unless you're OK with that.

sanjanarajan:
Ah I see, that makes sense. 
I do think it's the cleaner solution and could avoid confusion in the future, so if you are willing to make this larger change that would be great.

wiktor-k:
Okay, I've split User IDs and Attributes to `userId` (I've used uppercase i instead of lowercase here) and `userAttribute`.

There is one more thing that could be improved.

In almost all cases I'm passing userId and userAttribute:

```js
const dataToSign = {
  userId: this.userId,
  userAttribute: this.userAttribute,
  key: primaryKey
};
```

Maybe it would be better to just pass `this` (User object)? e.g.:

```js
const dataToSign = {
  user: this,
  key: primaryKey
};
```

Or directly the packet itself:

```js
const dataToSign = {
  packet: this.userId || this.userAttribute,
  key: primaryKey
};
```

Actually `User` class constructor saves the packet in `userId` or `userAttribute` but there several places in the code when the direct access to the packet is needed and the code looks like this: `dataToSign.userId || dataToSign.userAttribute`.

Another approach would be splitting `dataToSign` into two parameters to `verify` et al. (one is packet and the other is key)...

(By the way I think Travis is stuck as it's in `running` state for some time...)

tomholub:
Travis tested it successfully now. Sometimes it may take a while as there may be timeouts and retries, etc.

It would be great to add a test at the end if you care about this not breaking in the future!

wiktor-k:
@tomholub, what test do you mean exactly? I've added a regression test for the bug I've fixed in the first commit.

wiktor-k:
Oh, no problem, I was wondering if there is some edge case I missed.

src/packet/signature.js
```diff
@@ -591,25 +591,18 @@ Signature.prototype.toSign = function (type, data) {
...
-
-      if (data.userid !== undefined) {
-        tag = 0xB4;
-        packet = data.userid;
-      } else if (data.userattribute !== undefined) {
-        tag = 0xD1;
-        packet = data.userattribute;
-      } else {
-        throw new Error('Either a userid or userattribute packet needs to be ' +
+      const packet = data.userId || data.userAttribute;
```
sanjanarajan: Now that the userId and userAttributes are passed separately, the old signature code can be reverted right?

wiktor-k: Yes, with adjustment to the character case. I think I'll just remove the first commit completely and extract a test case from there.

wiktor-k:
Okay, I've squashed the changes and minimized them.

One thing that I had to do additionally was changing `if (data.userid !== undefined)` to `if (data.userid)` as the User class assigns `null` to non-existent packets and `null` is `!== undefined`.

I used the short form but there are several alternatives (assigning `undefined` instead of `null` in User class or using non-strict equality `!=` that'd also match in the case of `null`).

sanjanarajan:
Merged! Thank you for making these changes so quickly :)

PR #761, opened 2018-09-02
+ [feat] add the signatureExpirationTime support

snowyu:
Now you can specify the expiration time of the signature.
Add the signatureExpirationTime option to openpgp.encrpt/sign.

### DIFF
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -29,7 +29,7 @@ import enums from './enums';
 import util from './util';
 import packet from './packet';
 import { Signature } from './signature';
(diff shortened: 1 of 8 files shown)


sanjanarajan:
Hi :) A couple general comments:

1) Please rebase this on the latest master, there are a lot of merge conflicts

2) You don't need to make new alternate signing functions (like `signDetachedEx`, `createSignaturePacketsEx`) - just pass in a new expiration time parameter to the existing functions. The default for this parameter could be to never expire (the current behavior) for backwards compatibility.

sanjanarajan:
Out of date

PR #772, opened 2018-09-19
Check validity and keyid of signatures before using them

### LINKED ISSUE #752: \r\n is being normalized to \n in the decrypted file content
If I'm decrypting a file, I want the original file contents, not a normalized version of the content. I can't imagine why that would be an agreed upon pattern for the library.

Anyway, the use-case I'm having trouble with is a file is based on a fixed-width file format.

Say for example someone sends me an encrypted file that is sending me strings of n bytes in length, and new lines are not a guarantee of a line ending.

For example, a file may look like this:

```
THIS IS 16 BYTES                             \r\n
THIS IS THE NEXT \r\n
16 BYTES .  \r\n
THIS IS THE THIRD SET OF BYTES \r\n
```

Obviously the above is not an accurate representation of a fixed length set of strings, but it illustrates the point perfectly. By normalizing the `\r\n` character to `\n` you are effectively shortening a line by 1 character and invalidating the entire schema of the fixed-width file.

I found the following two issues for reference:
https://github.com/openpgpjs/openpgpjs/issues/636
https://github.com/openpgpjs/openpgpjs/issues/211

### LINKED ISSUE #762: Missing verification of signatures in getLatestSignature
I'm currently migrating from OpenPGP.js 2 to 4 and checking how certificates are handled internally and it looks to me like the current procedure is suboptimal in some places and can lead to false negatives.

In `Key.prototype.getPrimaryUser` the primary user is found as follows:
1. only the latest signature of a user is considered: https://github.com/openpgpjs/openpgpjs/blob/master/src/key.js#L515
`getLatestSignature` returns the latest signature, but does not verify before returning that signature. I think __in general only valid signatures should be considered__, so `getLatestSignature` should be more a 'getLatestValidSignature'. In the current implementation the latest self certificate of a user id could be invalid or expired, which would render the complete user id invalid although the user id might have other valid self certificates.
2. In https://github.com/openpgpjs/openpgpjs/blob/master/src/key.js#L532 users are sorted by primary user id flag and created date. At that point in time the self certificates have not been verified, so what we get as most relevant user id might not have a valid self certificate and will lead to returning null after verification here: https://github.com/openpgpjs/openpgpjs/blob/master/src/key.js#L541

I would consider the procedure in OpenPGP.js v2 as more accurate: https://github.com/openpgpjs/openpgpjs/blob/v2.6.2/src/key.js#L511
First you build objects for every user id - self signature combination, sort, return the first valid combination.

The missing signature verification in `getLatestSignature` also affects the following methods:
```
Key.prototype.getSigningKey
Key.prototype.getEncryptionKey
Key.prototype.getRevocationCertificate
SubKey.prototype.verify
SubKey.prototype.getExpirationTime
```

Pinging @AndreHeinecke and @dd9jn for feedback on how this is implemented in GnuPG

### LINKED ISSUE #765: Verifying user revocation certificate (DSA) fails
One of my old test keys throws an exception in v4.0.1 with `key.getPrimaryUser`:
"TypeError: Cannot read property 'toBN' of undefined" at https://github.com/openpgpjs/openpgpjs/blob/master/src/crypto/signature.js#L49 when verifying a non-self user revoke certificate:
```
:user ID packet: "Passwort  \xc3\xa4\xc3\x84??^^12 2 \xc2\xb5  <pwd@test.com>"
# off=315 ctb=88 tag=2 hlen=2 plen=90
:signature packet: algo 17, keyid E1790A6F8C13574D
	version 4, created 1445956182, md5len 0, sigclass 0x30
	digest algo 10, begin of digest 0e 63
	hashed subpkt 2 len 4 (sig created 2015-10-27)
	hashed subpkt 29 len 18 (revocation reason 0x20 (Ich weis es nicht))
	subpkt 16 len 8 (issuer key ID E1790A6F8C13574D)
	data: [160 bits]
	data: [160 bits]
# off=407 ctb=89 tag=2 hlen=3 plen=306
:signature packet: algo 1, keyid 63649F53DA0BDE5C
	version 4, created 1421224226, md5len 0, sigclass 0x10
	digest algo 8, begin of digest fd ce
	hashed subpkt 2 len 4 (sig created 2015-01-14)
	hashed subpkt 11 len 5 (pref-sym-algos: 9 8 7 3 2)
	hashed subpkt 16 len 8 (issuer key ID 63649F53DA0BDE5C)
	hashed subpkt 21 len 3 (pref-hash-algos: 8 2 10)
	hashed subpkt 22 len 2 (pref-zip-algos: 2 1)
	hashed subpkt 27 len 1 (key flags: 03)
	hashed subpkt 30 len 1 (features: 01)
	data: [2046 bits]
# off=716 ctb=88 tag=2 hlen=2 plen=70
:signature packet: algo 17, keyid E1790A6F8C13574D
	version 4, created 1445955936, md5len 0, sigclass 0x10
	digest algo 10, begin of digest 33 48
	hashed subpkt 2 len 4 (sig created 2015-10-27)
	subpkt 16 len 8 (issuer key ID E1790A6F8C13574D)
	data: [156 bits]
	data: [160 bits]
```
Here is the key in question and the key that signed it:
[Passwort  äÄ??^^12 2 µ  (DA0BDE5C) – public.asc.txt](https://github.com/openpgpjs/openpgpjs/files/2353928/Passwort.aA.12.2.DA0BDE5C.public.asc.txt)
[Signing key public.asc.txt](https://github.com/openpgpjs/openpgpjs/files/2353929/Signing.key.public.asc.txt)

twiss:
Fix #762, fix #765, fix #752, fix all the things!

This also includes some API changes:

- `Key.prototype.getRevocationCertificate` is now async
- `SubKey.prototype.getExpirationTime` is now async and takes `primaryKey, [date]` as parameters instead of just `[date]`
- When parsing a key with an authorized revocation key subpacket, we now return an error in the `(await openpgp.key.read/readArmored(key)).err` array, but also still return the key in the `.keys` array. This is because we ignore valid third-party revocation signatures made by authorized keys, so the key might not be safe to use in some cases.


src/key.js
```diff
@@ -261,30 +261,37 @@ Key.prototype.armor = function() {
...
       signature = signatures[i];
     }
   }
   return signature;
 }
 
-function isValidSigningKeyPacket(keyPacket, signature, date=new Date()) {
+function isValidSigningKeyPacket(keyPacket, signature) {
+  if (!signature.verified || signature.revoked !== false) { // Sanity check
+    throw new Error("Signature wasn't verified before calling isValidSigningKeyPacket");
```
sanjanarajan: Instead of doing this maybe we can just make this a sub-function of `Key.prototype.getSigningKey` as it isn't used anywhere else and doesn't seem generically useful

src/key.js
```diff
@@ -302,32 +309,34 @@ Key.prototype.getSigningKey = async function (keyId=null, date=new Date(), userI
...
+        isValidSigningKeyPacket(primaryKey, primaryUser.selfCertification)) {
       return this;
     }
   }
   return null;
 };
 
-function isValidEncryptionKeyPacket(keyPacket, signature, date=new Date()) {
+function isValidEncryptionKeyPacket(keyPacket, signature) {
+  if (!signature.verified || signature.revoked !== false) { // Sanity check
```
sanjanarajan: Same comment as with `isValidSigningKeyPacket`

Issue #776, opened 2018-10-03
decryption just kills node process


spearmootz:

spearmootz:
the private key and password match, i tested them on an application and was able to decrypt the file im trying to decrypt.

private key
```
`-----BEGIN PGP PRIVATE KEY BLOCK-----
Version: BCPG C# v1.8.1.0

key here
-----END PGP PRIVATE KEY BLOCK-----`
```

spearmootz:
so im running a test and if i do a setTimeout it tells node its not done processing yet so it stays open. that tells me that node is not crashing but that the decrypt function is not resolving or rejecting which ive never seen before.

```
const fs = require('fs');
const path = require('path');
const pgpDecrypt = require('./pgpDecrypt').default;

fs.readFile(path.join(__dirname, x), async (error, file) => {
  try {
    console.log('here');
    const results = await pgpDecrypt('fileName');
    console.error(results);
  } catch (error) {
    console.error(error);
  }
  console.log('here');
});

setTimeout(() => null, 100000);

```

spearmootz:
ive narrowed the error to the following code. it looks like it should work but it doesnt. i tinkered with it and realized i could reduce the code while doing the same thing. the second except is my code

```
/**
           * Wait until signature objects have been verified
           * @param  {Object} signatures              list of signatures
           */
          async function prepareSignatures(signatures) {
            await Promise.all(
              signatures.map(async signature => {
                signature.signature = await signature.signature;
                signature.valid = await signature.verified;
              })
            );
          }
```



```
/**
           * Wait until signature objects have been verified
           * @param  {Object} signatures              list of signatures
           */
          async function prepareSignatures(signatures) {
            signatures.forEach(async signature => {
              signature.signature = await signature.signature;
              signature.valid = await signature.verified;
            });
          }
```

twiss:
Hi @spearmootz, thanks for digging into this!

What it sounds like might be happening is that your message contains a one-pass signature packet without a corresponding signature packet.

To confirm, could you perhaps either import the private key in GPG and post the output of `gpg --list-packets <filename>`, or post an example private key and message?

spearmootz:
@twiss i cannot post the private key and message because it contains information that i cannot share.  here is the output of the command you mentioned but i modified the email, the key name and the file name

```
gpg: encrypted with 2048-bit RSA key, ID D1542F5B97E91901, created 2017-08-08
      "name of key (080817_Current_Key) <some email>"
[...]

PR #777, opened 2018-10-03
fix promise not resolving or rejecting

spearmootz:
https://github.com/openpgpjs/openpgpjs/issues/776


spearmootz:
breaks test, and i think  i have to dig deeper into why the bug happens

PR #779, opened 2018-10-05
Throw on parse errors in integrity protected encrypted packets

twiss:


src/packet/packetlist.js
```diff
@@ -53,10 +53,9 @@ List.prototype.read = async function (bytes) {
             await packet.read(parsed.packet);
             await writer.write(packet);
           } catch (e) {
-            if (!config.tolerant ||
-                parsed.tag === enums.packet.symmetricallyEncrypted ||
-                parsed.tag === enums.packet.literal ||
-                parsed.tag === enums.packet.compressed) {
+            if (!config.tolerant || packetParser.supportsStreaming(parsed.tag)) {
```
sanjanarajan: In case this is not always a one-to-one relationship, maybe we can just continue to enumerate them here? I think this could cause confusion in the future

Issue #271, opened 2014-11-30
Support for multiple recipients? 


daum:
Does openpgpjs support multiple recipients from different public keys?  I saw #68 which has multiple keys from one armoredText, however is there a way to load more than one public key from different armoredText's for the encryption routine?  

Thanks!

ashcs:
Please example for latest version

avimar:
How do I do this? The link is no longer active.

I tried passing in an array of keys, that didn't work.

I tried putting both public keys in the same text for an import, and it encrypted it, but firebase.io said that it wasn't my key. Is that the only way? Is it just a bug with firebase.io?

bartbutler:
An array of keys should work. Please look at the tests, there should be one or more examples in there of using multiple public keys.

avimar:
I assume here https://github.com/openpgpjs/openpgpjs/blob/master/test/general/openpgp.js but despite the naming of `keys` it seems to only be one.

bartbutler:
What happens when you try an array with multiple keys?

avimar:
`(node:14700) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Error encrypting message: Cannot read property 'getPrimaryUser' of undefined`

Here's my code:

https://gist.github.com/avimar/b02c34a20d90d7476f615766b34246ed

bartbutler:
FIrst, your keys are invalid. There needs to be a newline after the key headers that your keys lack. openpgp.key.readArmored has an 'err' field that will be an Error object if the key parsing fails.

Second, instead of '.key' you needs to use '.keys[0]' or merge the key array, as there can be (though there usually aren't) multiple keys inside an armored block.

With those changes your code worked for me.

avimar:
Wow, `.keys` somehow worked fine -- that's what the getting started docs show: https://github.com/openpgpjs/openpgpjs/blob/master/README.md#encrypt-and-decrypt-string-data-with-pgp-keys

But when used in an array, it needs `[0]` specified.

So my code now seems to work fine, I just made that small change.
Thanks!

Code below for anyone who will need in the future:

```
options = {
    data: 'Hello, World!',                             // input as String (or Uint8Array)
    publicKeys: [                      // for encryption
	openpgp.key.readArmored(pubkey1).keys[0],
	openpgp.key.readArmored(pubkey2).keys[2]  
	]
   };
```

bartbutler:
[...]

PR #781, opened 2018-10-08
Multiple public keys code

evildvl:
Update README.md with the code example by @avimar (https://github.com/openpgpjs/openpgpjs/issues/271#issuecomment-274901731). Think that it's too hard to find and missed in documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,40 @@ const encryptDecryptFunction = async() => {
 encryptDecryptFunction()
 ```
 
+Encrypt with multiple public keys:
+
+```js
+const pubkey1 = `-----BEGIN PGP PUBLIC KEY BLOCK-----
+...
+-----END PGP PUBLIC KEY BLOCK-----`
+const pubkey2 = `-----BEGIN PGP PUBLIC KEY BLOCK-----
+...
+-----END PGP PUBLIC KEY BLOCK-----`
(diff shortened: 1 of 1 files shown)


sanjanarajan:
Could you please integrate this into the existing "Encrypt and decrypt String data with PGP keys" example? It's quite out of place as a standalone variable definition in between generation and revocation.

evildvl:
Sure, branch updated.

evildvl:
@avimar updated

README.md
```diff
@@ -198,6 +198,42 @@ const encryptDecryptFunction = async() => {
...
+    const privKeyObj = (await openpgp.key.readArmored(privkey)).keys[0]
+    await privKeyObj.decrypt(passphrase)
+    
+    pubkeys = pubkeys.map(async (key) => {
+    	return (await openpgp.key.readArmored(key)).keys[0]
+    });
+    
+    const options = {
+        message: openpgp.message.fromText(message),       
+        publicKeys: pubkeys,           				  // for encryption
```
niponchi: From line 220, `pubkeys` is array of `Promise`, not array of `Key`.
In this case, line 226 should be `publicKeys: await Promise.all(pubkeys)`

evildvl: Helllo, I will check it and fix if here is the error

PR #782, opened 2018-10-11
added helpful comment about default encryption and decryption of U8in…

ZelphirKaltstahl:
…tArray with password

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ openpgp.initWorker({ path:'openpgp.worker.js' }) // set the relative web worker 
 
 #### Encrypt and decrypt *Uint8Array* data with a password
 
-This is encryption/decryption method will use AES-256 by default.
+Encryption will use AES-256 by default. Decryption will use whatever was used when encrypting, as such information is readably stored in the encrypted message.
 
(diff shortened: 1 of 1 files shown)


sanjanarajan:
Could you please remove the extraneous newlines and edit the comment from "This is encryption/decryption method will use AES-256 by default" to "Encryption will use AES-256 by default"?
Decryption has no default, as we will use whatever algorithm was used for encryption.

ZelphirKaltstahl:
@sanjanarajan I will modify the explanation and also add the info about decryption not using a default, but using whatever was used in encryption (as such information is held in the encrypted message). I don't see the extraneous newlines. There is one after the heading and then one after the explanation I added and then comes the js code block. The format seems to be to have empty lines after the headings and before code block in the rest of the document. Or do you mean elsewhere?

And do you mean I should create a new pull request, or can I somehow update this one? (I haven't done so many pull requests yet and never had to modify anything after the fact.)

sanjanarajan:
Thanks :) I edited this to have a bit more detail about where the defaults come from, you can check master for the latest.

Issue #788, opened 2018-10-21
Lookup public key on HKP server async


nickmi:
While trying this in webstorm i do receive an error because of the await:
```
hkp.lookup(options).then(function(key) {
    var pubkey = await openpgp.key.readArmored(key);
});

```
This doesnt seem to fix it: 
```
async function do(){
var hkp = new openpgp.HKP('https://pgp.mit.edu');
var options = {
    query: 'alice@example.com'
};
hkp.lookup(options).then(function(key) {
    var pubkey = await openpgp.key.readArmored(key);
});}
```

Unless its a local problem on my part i think that the example  in the front page could be changed to the following:
```
hkp.lookup(options).then( async function(key) {
    var pubkey = await openpgp.key.readArmored(key);
});

```

tomholub:
True, as written, the example won't work. It should have probably been:
```js
let armoredPubkey = await hkp.lookup(options);
var pubkey = await openpgp.key.readArmored(armoredPubkey);
```

PR #790, opened 2018-10-23
fixed key lookup example on  README.md

nickmi:
Related to issue #788, i just changed the example on the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,7 +416,7 @@ var options = {
     query: 'alice@example.com'
 };
 
-hkp.lookup(options).then(function(key) {
+hkp.lookup(options).then( async function(key) {
     var pubkey = await openpgp.key.readArmored(key);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -416,7 +416,7 @@ var options = {
     query: 'alice@example.com'
 };
 
-hkp.lookup(options).then(function(key) {
+hkp.lookup(options).then( async function(key) {
```
tomholub: I do recommend to use either .then or async/await, but not both - it gets messy. Since we do most examples with async/await, the following would be better:
```js
let armoredPubkey = await hkp.lookup(options);
var pubkey = await openpgp.key.readArmored(armoredPubkey);
```

nickmi: Thank you for the feedback, made the changes as recommended

PR #795, opened 2018-11-01
Web Crypto hashing

### LINKED ISSUE #426: Native SHA hashing via WebCrypto
We already use native hashing under node.js. This would add native hashing on all platforms. A large part of the work would be to refactor many private api from a sync style to an async/promise style. The work was already started in the following branch:

https://github.com/openpgpjs/openpgpjs/tree/webcrypto_hashing

Once we use WebCrypto, we can also remove dependencies on Rusha (SHA-1) and asmCrypto's SHA-256. jsSHA would need to stay though as not all required SHA operations are supported by WebCrypto.

twiss:
Fix #426.


src/crypto/hash/index.js
```diff
@@ -68,11 +76,11 @@ if (nodeCrypto) { // Use Node native crypto for all hash functions
 } else { // Use JS fallbacks
   hash_fns = {
     md5: md5,
-    sha1: asmcrypto_hash(Sha1),
+    sha1: asmcrypto_hash(Sha1, navigator.userAgent.indexOf('Edge') === -1 && 'SHA-1'),
```
sanjanarajan: Why?

twiss: Edge doesn't support SHA-1: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#Parameters

PR #796, opened 2018-11-04
Return String in openpgp.revokeKey()

terrencew97:
The `revokeKey()` should return String, but not the readableStream.

I found that there is a function called `convertStreams()`, but I have no idea why it throws

> `Error revoking key: This stream has already been locked for exclusive reading by another reader at new e`

That's why I used `convertStream()` directly.

### DIFF
--- a/src/openpgp.js
+++ b/src/openpgp.js
@@ -209,14 +209,14 @@ export function revokeKey({
       const publicKey = key.toPublic();
       return {
         privateKey: key,
-        privateKeyArmored: key.armor(),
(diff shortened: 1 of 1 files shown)


twiss:
Hey! Thanks for catching this :)

The reason `convertStreams` would throw is because the object essentially contains the same stream twice. So I've fixed it in https://github.com/openpgpjs/openpgpjs/commit/2245df6023bdaa04f4d737c76d79c6c2bb6173f9 by calling it on `key` ahead of creating the object.

PR #800, opened 2018-11-06
ci: test Node.js 8, 10 and 11

DanielRuf:
Node.js 9 and other odd release numbers already reached their EOL and are short living.

https://github.com/nodejs/Release/blob/master/README.md

Only test 8, 10 and current stable (11).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,28 +13,30 @@ matrix:
       env: OPENPGP_NODE_JS='8' OPENPGPJSTEST='unit'
     - node_js: "10"
       env: OPENPGP_NODE_JS='10' OPENPGPJSTEST='unit'
-    - node_js: "9"
+    - node_js: "node"
+      env: OPENPGP_NODE_JS='10' OPENPGPJSTEST='unit'
+    - node_js: "10"
       env: BROWSER='Firefox' VERSION='61' PLATFORM='macOS 10.13' OPENPGPJSTEST='saucelabs'
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='Chrome' VERSION='49' PLATFORM='macOS 10.13' OPENPGPJSTEST='saucelabs' COMPAT=1
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='Chrome' VERSION='68' PLATFORM='macOS 10.13' OPENPGPJSTEST='saucelabs'
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='Internet Explorer' VERSION='11.103' PLATFORM='Windows 10' OPENPGPJSTEST='saucelabs' COMPAT=1
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='MicrosoftEdge' VERSION='17.17134' PLATFORM='Windows 10' OPENPGPJSTEST='saucelabs'
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='Safari' VERSION='10' PLATFORM='macOS 10.12' OPENPGPJSTEST='saucelabs' COMPAT=1
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='Safari' VERSION='11' PLATFORM='macOS 10.13' OPENPGPJSTEST='saucelabs'
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='Android' VERSION='6.0' OPENPGPJSTEST='saucelabs'
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='iPad' VERSION='10.0' OPENPGPJSTEST='saucelabs' COMPAT=1
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='iPad' VERSION='11.0' OPENPGPJSTEST='saucelabs'
   allow_failures:
-    - node_js: "9"
+    - node_js: "10"
 
 before_script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,28 +13,30 @@ matrix:
...
       env: BROWSER='Android' VERSION='6.0' OPENPGPJSTEST='saucelabs'
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='iPad' VERSION='10.0' OPENPGPJSTEST='saucelabs' COMPAT=1
-    - node_js: "9"
+    - node_js: "10"
       env: BROWSER='iPad' VERSION='11.0' OPENPGPJSTEST='saucelabs'
   allow_failures:
-    - node_js: "9"
+    - node_js: "10"
```
twiss: This causes the Node.js 10 test to be allowed to fail, in addition to the browser tests. Since it's a simple string match, you could change it (and the `node_js` versions of the browser tests above) to "10.0" or "11". But since we're testing browsers here, and not the node version we're running them on, leaving it at 9 should also be fine.

Or you could just remove the Node.js 10 test, since I think it's a bit unlikely that OpenPGP.js would work in Node 8 and 11 but be broken in 10. But I'll leave it up to you.

.travis.yml
```diff
@@ -13,28 +13,30 @@ matrix:
       env: OPENPGP_NODE_JS='8' OPENPGPJSTEST='unit'
     - node_js: "10"
       env: OPENPGP_NODE_JS='10' OPENPGPJSTEST='unit'
-    - node_js: "9"
+    - node_js: "node"
+      env: OPENPGP_NODE_JS='10' OPENPGPJSTEST='unit'
```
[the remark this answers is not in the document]
DanielRuf: > 'latest'

Not sure but latest / node is actually 11 on Travis CI and other platforms.

.travis.yml
DanielRuf: > leaving it at 9 should also be fine.

Which is not needed imho as Node.js 9 reached its EOL and should not be used in production.

DanielRuf: I'll probably remove the `allow_failures` entries.

.travis.yml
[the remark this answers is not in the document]
twiss: Yes, that's why setting the env variable `OPENPGP_NODE_JS='10'` under `node_js: "node"` here is confusing.

.travis.yml
twiss: The `allow_failures` entry is still needed here for the browser tests, which often fail for no particular reason. The reason it stated `node_js: "9"` is not because Node 9 is allowed to fail, but because all the browser tests ran on Node 9.

DanielRuf: > but because all the browser tests ran on Node 9

Well, most libraries are not tested against Node.js 9 (anymore) and it will not receive any bugfixes and security fixes. If so, we can filter by `OPENPGPJSTEST` or similar.

.travis.yml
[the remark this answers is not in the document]
DanielRuf: I just oversaw it and yes `11` or `stable` is the right value.

.travis.yml
twiss: Travis doesn't match by individual env variables.

DanielRuf: Mh, that might be a bit more complicated. See https://github.com/travis-ci/travis-ci/issues/6661

DanielRuf: But isn't Node.js also used for building the library when running the browser tests? So we can not just set the version to something different and think that this is enough. I understand that this is currently done to split tests and allow failing Saucelabs test. But should they be really allowed to fail?

twiss: Yes, but building the library is also done in the Node tests. Besides, that doesn't really differ at all between different versions of Node.

.travis.yml
```diff
@@ -37,7 +37,16 @@ matrix:
...
+    - env: BROWSER='Firefox' VERSION='61' PLATFORM='macOS 10.13' OPENPGPJSTEST='saucelabs'
+    - env: BROWSER='Chrome' VERSION='49' PLATFORM='macOS 10.13' OPENPGPJSTEST='saucelabs' COMPAT=1
+    - env: BROWSER='Chrome' VERSION='68' PLATFORM='macOS 10.13' OPENPGPJSTEST='saucelabs'
+    - env: BROWSER='Internet Explorer' VERSION='11.103' PLATFORM='Windows 10' OPENPGPJSTEST='saucelabs' COMPAT=1
+    - env: BROWSER='MicrosoftEdge' VERSION='17.17134' PLATFORM='Windows 10' OPENPGPJSTEST='saucelabs'
+    - env: BROWSER='Safari' VERSION='10' PLATFORM='macOS 10.12' OPENPGPJSTEST='saucelabs' COMPAT=1
+    - env: BROWSER='Safari' VERSION='11' PLATFORM='macOS 10.13' OPENPGPJSTEST='saucelabs'
+    - env: BROWSER='Android' VERSION='6.0' OPENPGPJSTEST='saucelabs'
+    - env: BROWSER='iPad' VERSION='10.0' OPENPGPJSTEST='saucelabs' COMPAT=1
+    - env: BROWSER='iPad' VERSION='11.0' OPENPGPJSTEST='saucelabs'
```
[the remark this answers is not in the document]
DanielRuf: As already replied, I'll revert this change ;-)

twiss:
Fixed by #889

PR #808, opened 2018-11-28
Move key revocation check to key class. Evaluate only self-signatures.

toberndo:
The strict check on the revocation key features as in https://github.com/openpgpjs/openpgpjs/commit/d314a20e0fd9261683517b55c4aede444a88f8f8 has I think some issues:

1. If there are multiple PGP keys in one armored block and one key contains the revocation key sub packet, then parsing stops and no keys are returned. We should skip here only the affected key but process the other keys in the armored block as normal.

2. [5.2.3.15. Revocation Key](https://tools.ietf.org/html/rfc4880#section-5.2.3.15) defines this feature to be restricted to self-signatures ("This is found on a self-signature."). With self-signatures being certification signatures (types 0x10-0x13), the direct-key signature (type 0x1F), and the subkey binding signature (type 0x18) ([5.2.3.3](https://tools.ietf.org/html/rfc4880#section-5.2.3.3)). Currently every signature of the key is checked for the flag, and as a result for example every user signature could have the flag set (illegally) and render the key unusable in OpenPGP.js

3. An information from the signature (the revocation key flag) is used before verifying the signature. The means a forged self-signature added to the key can be used to lock the key out of OpenPGP.js. The self-signature should first be verified and only valid signatures with the  revocation key feature should trigger an exception.

The PR addresses 1-2 but not 3.

### DIFF
--- a/src/key.js
+++ b/src/key.js
@@ -105,6 +105,7 @@ Key.prototype.packetlist2structure = function(packetlist) {
               continue;
             }
             if (packetlist[i].issuerKeyId.equals(primaryKeyId)) {
+              checkRevocationKey(packetlist[i], primaryKeyId);
               user.selfCertifications.push(packetlist[i]);
             } else {
               user.otherCertifications.push(packetlist[i]);
@@ -118,6 +119,7 @@ Key.prototype.packetlist2structure = function(packetlist) {
             }
             break;
           case enums.signature.key:
+            checkRevocationKey(packetlist[i], primaryKeyId);
             this.directSignatures.push(packetlist[i]);
             break;
           case enums.signature.subkey_binding:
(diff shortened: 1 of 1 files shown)


src/key.js
```diff
@@ -157,6 +160,16 @@ Key.prototype.toPacketlist = function() {
   return packetlist;
 };
 
+/**
+ * Throws error if signature has revocation key sub packet (not supported by OpenPGP.js)
+ */
+Key.prototype.checkRevocationKey = function(signature) {
```
twiss: This doesn't really seem like it needs to be a public function exported on the prototype, could you make it just a normal private function?

src/key.js
```diff
@@ -157,6 +160,16 @@ Key.prototype.toPacketlist = function() {
...
+  if (signature.revocationKeyClass !== null &&
+      signature.issuerKeyId.equals(this.getKeyId())) {
```
twiss: You have `this.getKeyId()` available as `primaryKeyId` in the calling locations, you could pass that instead of using `this` if you like

toberndo:
@twiss Yup, makes sense. Please see update.

PR #816, opened 2018-12-14
Security fixes

twiss:


src/message.js
```diff
@@ -181,23 +182,35 @@ Message.prototype.decryptSessionKeys = async function(privateKeys, passwords) {
...
-        if (!privateKeyPacket.isDecrypted()) {
-          throw new Error('Private key is not decrypted.');
-        }
-        try {
-          await keyPacket.decrypt(privateKeyPacket);
-          keyPackets.push(keyPacket);
-        } catch (err) {
-          util.print_debug_error(err);
+      await Promise.all(privateKeys.map(async function(privateKey) {
+        const primaryUser = await privateKey.getPrimaryUser(/* symEncryptedPacket.created, userId */);
```
sanjanarajan: What does this comment mean?

src/message.js
```diff
@@ -181,23 +182,35 @@ Message.prototype.decryptSessionKeys = async function(privateKeys, passwords) {
...
+        let algos = [
+          enums.symmetric.aes256, // OpenPGP.js default
```
sanjanarajan: Should we replace this with the current config value? Or perhaps add that one as well?

twiss: Yeah, maybe. The issue is that if you change the config value, other people using the normal default will throw -- so maybe it's better to find that out quickly.
Similarly, GPG will warn even with our normal default if we send an AES256 message to someone without AES256 in the preferred keys - so maybe this is just an argument to switch to 3DES as the fallback, as the spec mandates, make that non-configurable, and then add the config value to the start of the preferred algos when generating keys (and use it for password encryption).

src/message.js
sanjanarajan: Why can't we at least pass in the time though?

twiss: Yeah we can do that

twiss: Actually, it turns out the `symEncryptedPacket.created` property was wishful thinking, and it doesn't exist, so we don't have the time here.

src/message.js
sanjanarajan: Yes, let's do that

PR #824, opened 2018-12-23
[WIP] use duck-typing for util.isStream -- open discussion

tonymet:
Presently isStream tests for object.prototype  = Readable Stream, which makes it a bit strict when dealing with stream-like objects. 

I'm working on a pipeline which uses npm tar to generate encrypted tar files (see below). I would like to pass the tarStream to openpgp.message.fromBinary, but that fails in the current build. 

Just want to open the conversation and see if this duck-typing would be a good approach.  If you like it, I can add the remaining code and tests

```
var tarStream = tar.create({
    gzip: false,
    //file: 'tmp/infile.tar'
}, ['tmp/testtar'])

debug(tarStream)

const options = {
    // default  util.isStream() would error here because tarStream is not strictly ReadableStream
    message: openpgp.message.fromBinary(tarStream), // input as Message object
    passwords: ['secret stuff'], // multiple passwords possible
    armor: false // don't ASCII armor (for Uint8Array output)
};
openpgp.encrypt(options).then(async function(ciphertext) {
    const encrypted = ciphertext.message.packets.write();
    const reader = openpgp.stream.getReader(encrypted);
    await pipe(reader, writeStream)
    debug("writing a block", ciphertext.message.packets.length)
});
```


twiss:
I think this should probably be fixed in the `isStream` function itself, at https://github.com/openpgpjs/web-stream-tools/blob/84a497715c9df271a673f8616318264ab42ab3cc/lib/util.js#L8.

I think the issue is not that it's not a "real" stream, but rather that it's a [Duplex stream](https://nodejs.org/api/stream.html#stream_class_stream_duplex) which apparently implements the same interface as a Readable stream, but doesn't inherit from it. So you could just add a check for a Duplex stream to the `if` statement above.

tonymet:
thanks for the workaround I'll give that a shot.

PR #827, opened 2018-12-27
Double is readonly

estelendur:
When I was building my project that includes openpgpjs, I got an error:
```
SyntaxError: "double" is read-only
```

The error went away if I renamed the variable `double` inside the `double` function in `util.js`.


sanjanarajan:
Thank you for your contribution! Could you please remove all dist files from the PR?

estelendur:
Yes, sorry about that :)

PR #844, opened 2019-01-18
Make ephemeral secret available from ECDH module

wussler:
Split the session key encryption and decryption process in two steps, thus allowing an ECDHE exchange through functions genPublicEphemeralKey and genPrivateEphemeralKey.

### DIFF
--- a/src/crypto/public_key/elliptic/ecdh.js
+++ b/src/crypto/public_key/elliptic/ecdh.js
@@ -58,6 +58,30 @@ async function kdf(hash_algo, X, length, param) {
   return digest.subarray(0, length);
 }
 
+/**
+ * Generate ECDHE ephemeral key and secret from public key
+ *
+ * @param  {module:type/oid}        oid          Elliptic curve object identifier
+ * @param  {module:enums.symmetric} cipher_algo  Symmetric cipher to use
+ * @param  {module:enums.hash}      hash_algo    Hash algorithm to use
+ * @param  {Uint8Array}             Q            Recipient public key
+ * @param  {String}                 fingerprint  Recipient fingerprint
+ * @returns {Promise<{V: BN, C: Uint8Array}>}    Returns public part of ephemeral key and generated ephemeral secret
+ * @async
+ */
+async function genPublicEphemeralKey(oid, cipher_algo, hash_algo, Q, fingerprint) {
+  const curve = new Curve(oid);
+  const param = buildEcdhParam(enums.publicKey.ecdh, oid, cipher_algo, hash_algo, fingerprint);
+  cipher_algo = enums.read(enums.symmetric, cipher_algo);
+  const v = await curve.genKeyPair();
+  Q = curve.keyFromPublic(Q);
+  const S = v.derive(Q);
+  const Z = await kdf(hash_algo, S, cipher[cipher_algo].keySize, param);
+  return {
+    V: new BN(v.getPublic()),
+    Z: Z
+  };
+}
 
 /**
  * Encrypt and wrap a session key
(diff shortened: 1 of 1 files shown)


src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -58,56 +58,90 @@ async function kdf(hash_algo, X, length, param) {
...
+ * @param  {module:enums.symmetric} cipher_algo  Symmetric cipher to use
+ * @param  {module:enums.hash}      hash_algo    Hash algorithm to use
+ * @param  {module:type/mpi}        m            Value derived from session key (RFC 6637)
+ * @param  {Uint8Array}             Q            Recipient public key
+ * @param  {String}                 fingerprint  Recipient fingerprint
+ * @returns {Promise<{V: BN, C: BN}>}            Returns public part of ephemeral key and encoded session key
+ * @async
+ */
+async function encrypt(oid, cipher_algo, hash_algo, m, Q, fingerprint) {
+  const key = await genPublicEphemeralKey(oid, cipher_algo, hash_algo, Q, fingerprint);
```
twiss: Purely nitpicking here, but you can destructure this as
```suggestion
  const { V, Z } = await genPublicEphemeralKey(oid, cipher_algo, hash_algo, Q, fingerprint);
```

src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -58,56 +58,90 @@ async function kdf(hash_algo, X, length, param) {
...
+  const C = aes_kw.wrap(key.Z, m.toString());
+  return {
```
twiss: And then here
```suggestion
  return { V, C };
```

src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -58,56 +58,90 @@ async function kdf(hash_algo, X, length, param) {
...
+ * @returns {Promise<Uint8Array>}                Value derived from session
+ * @async
+ */
+async function decrypt(oid, cipher_algo, hash_algo, V, C, d, fingerprint) {
+  const Z = await genPrivateEphemeralKey(oid, cipher_algo, hash_algo, V, d, fingerprint);
   return new BN(aes_kw.unwrap(Z, C));
 }
 
-export default { encrypt, decrypt };
+export default { encrypt, decrypt, genPublicEphemeralKey, genPrivateEphemeralKey };
```
twiss: This is a bit a matter of opinion but since it's technically a public function it might be worth writing out `generate` in full, even though the names are already kinda long
```suggestion
export default { encrypt, decrypt, generatePublicEphemeralKey, generatePrivateEphemeralKey };
```

wussler: The name is made following the style of genKeyPair in ./Curves

src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -58,56 +58,86 @@ async function kdf(hash_algo, X, length, param) {
...
   cipher_algo = enums.read(enums.symmetric, cipher_algo);
   const v = await curve.genKeyPair();
   Q = curve.keyFromPublic(Q);
   const S = v.derive(Q);
-  const Z = await kdf(hash_algo, S, cipher[cipher_algo].keySize, param);
-  const C = aes_kw.wrap(Z, m.toString());
   return {
     V: new BN(v.getPublic()),
-    C: C
+    Z: await kdf(hash_algo, S, cipher[cipher_algo].keySize, param)
```
twiss: These are becoming increasingly nitpickier nitpicks, but if you add `const V = new BN(v.getPublic());` above and restore the `const Z = ...` as well, you can `return { V, Z };` here and the linter shouldn't complain.

PR #853, opened 2019-02-04
Fix ECDH message encryption for some session keys

twiss:


test/general/openpgp.js
```diff
@@ -304,6 +304,37 @@ const twoPasswordGPGFail = ['-----BEGIN PGP MESSAGE-----',
 '=cHCV',
 '-----END PGP MESSAGE-----'].join('\n');
 
+const feng_msg_bad = `-----BEGIN PGP MESSAGE-----
```
sanjanarajan: Maybe name these something more descriptive of the bug and less descriptive of feng 😂

PR #855, opened 2019-02-06
++ Add another Domain for HKP server

rash0:
the pgp.mit.edu domain is most of the time down and responds with time out...so i discovered this ubuntu domain for the same database...but its much faster and never falls....i think too much traffic over the bit one :(

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,8 +409,7 @@ openpgp.revokeKey(options).then(function(key) {
 #### Lookup public key on HKP server
 
 ```js
-//Another domain for same server also available, https://keyserver.ubuntu.com
-var hkp = new openpgp.HKP('https://pgp.mit.edu');
+var hkp = new openpgp.HKP('https://keyserver.ubuntu.com');
 
 var options = {
     query: 'alice@example.com'
--- a/README.md
+++ b/README.md
@@ -409,6 +409,7 @@ openpgp.revokeKey(options).then(function(key) {
 #### Lookup public key on HKP server
 
 ```js
+//Another domain for same server also available, https://keyserver.ubuntu.com
 var hkp = new openpgp.HKP('https://pgp.mit.edu');
 
 var options = {
--- a/README.md
+++ b/README.md
@@ -409,7 +409,7 @@ openpgp.revokeKey(options).then(function(key) {
 #### Lookup public key on HKP server
 
 ```js
-var hkp = new openpgp.HKP('https://keyserver.ubuntu.com');
+var hkp = new openpgp.HKP(); // Defaults to https://keyserver.ubuntu.com, or pass another keyserver URL as a string
 
 var options = {
     query: 'alice@example.com'
--- a/src/hkp.js
+++ b/src/hkp.js
@@ -27,7 +27,8 @@ import config from './config';
  * Initialize the HKP client and configure it with the key server url and fetch function.
  * @constructor
  * @param {String}    keyServerBaseUrl  (optional) The HKP key server base url including
- *   the protocol to use e.g. https://keyserver.ubuntu.com
+ *   the protocol to use, e.g. 'https://pgp.mit.edu'; defaults to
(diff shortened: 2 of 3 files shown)


tomholub:
You might as well straight up replace them in the code. At FlowCrypt we do sometimes check SKS servers on the backend, and the MIT server is basically useless. ubuntu is the most stable server of the 20 we tried.

tomholub:
@rash0 just to clarify, I'm only an user of this library and a contributor, not maintainer, so this is obviously up to the maintainers to evaluate. For me it's thumbs up.

rash0:
Done :)

README.md
```diff
@@ -409,7 +409,7 @@ openpgp.revokeKey(options).then(function(key) {
 #### Lookup public key on HKP server
 
 ```js
-var hkp = new openpgp.HKP('https://pgp.mit.edu');
+var hkp = new openpgp.HKP('https://keyserver.ubuntu.com');
```
twiss:
```suggestion
var hkp = new openpgp.HKP(); // Defaults to https://keyserver.ubuntu.com, or pass another keyserver URL as a string
```

src/hkp.js
```diff
@@ -27,7 +27,7 @@ import config from './config';
  * Initialize the HKP client and configure it with the key server url and fetch function.
  * @constructor
  * @param {String}    keyServerBaseUrl  (optional) The HKP key server base url including
- *   the protocol to use e.g. https://pgp.mit.edu
+ *   the protocol to use e.g. https://keyserver.ubuntu.com
```
twiss:
```suggestion
 *   the protocol to use, e.g. 'https://pgp.mit.edu'; defaults to
 *   openpgp.config.keyserver (https://keyserver.ubuntu.com)
```

tomholub:
Ah, who would have known. +1

src/hkp.js
rash0: I think its cool to leave the mit.edu one too...having options and choice is always good

rash0: you don't think that it should be mentioned in the readme file also ??

rash0:
Also i was reading that the HKP server accepts revocation certificates on the same url '/pks/add' ...
but didn't find anything anywhere about it in the documentation ?? ....did i miss it somewhere ??

src/hkp.js
twiss: There are two examples in the readme, one mentions the Ubuntu server and one mentions the PGP server.

twiss:
> Also i was reading that the HKP server accepts revocation certificates on the same url '/pks/add' ...
but didn't find anything anywhere about it in the documentation ?? ....did i miss it somewhere ??

Yes, you should be able to add individual packets to an existing keys, including revocation signatures. So you can probably just use the existing `hkp.add` function for that, although I haven't tested it.

rash0:
> Yes, you should be able to add individual packets to an existing keys, including revocation signatures. So you can probably just use the existing `hkp.add` function for that, although I haven't tested it.
> 
> Alternatively you can apply the revocation certificate to the key with `openpgp.revokeKey` and then pass the result of that to `hkp.add` to update the key.

Yes, this what i meant, just to mention that  it in the readme, that the `hkp.upload()` could be used also to upload revocation certificate....I didn't test it also using the openpgp library, but i uploaded a revocation certificate few days ago to ubuntu using the same web interface for uploading a new key...and it worked.

So, Should i add a comment that the  `hkp.upload()` could be used for uploading a new key and also uploading revocation certificate....or.... anyone have a suggestions of how to test uploading it using openpgp and sharing the results here ???

I just assumes that it works....because from my understanding is that the revocation certificate is itself some sort of a public key with a the same fingerprint as the public key. 🤔🤔

rash0:
times interested in my changes or should i close this ?!

twiss:
Yes, I am, thanks for your contributions :) We'll merge them soon. Sorry for the delay.

PR #858, opened 2019-02-14
Expand git hashes in dependencies to fix pnpm install

ssimpo:
Need to switch git hashes in dependencies from short to long so **[pnpm](https://github.com/pnpm)** install packages.  The pnpm installer does not support short hashes (yet).  This is a simple changes, which makes packages that depend on this (such as [WildDuck](https://github.com/nodemailer/wildduck)) installable.

**See:**
- pnpm/git-resolver#15
- pnpm/pnpm#1127


twiss:
Thanks!

PR #865, opened 2019-02-24
Support GNU export-secret-subkeys extension

### LINKED ISSUE #251: Feature Request: Handle 'export-secret-subkeys' keyrings from GPG
It would be awesome to see OpenPGP.js support GPG's extension to PGP whereby you remove the private key to the master key of a keyring.

I understand this is not formally part of the PGP standard, however it's a valuable security feature and appears to be [growing](https://github.com/toberndo/mailvelope/issues/186) in [popularity](https://github.com/toberndo/mailvelope/issues/199).
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4013557-feature-request-handle-export-secret-subkeys-keyrings-from-gpg?utm_campaign=plugin&utm_content=tracker%2F290945&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F290945&utm_medium=issues&utm_source=github).
</bountysource-plugin>

twiss:
Fix #251.


tomholub:
Woohoo!

I was trying to test this, but cannot build it:
```
$ npm install && npm test
// ... cropped

> openpgp@4.4.7 pretest git/openpgpjs-d60efff61d024421e7985e25e7156cb4856d9d9a
> grunt

(node:14135) ExperimentalWarning: The http2 module is an experimental API.
Running "browserify:openpgp" (browserify) task
>> Error: Cannot find module 'asmcrypto.js/dist_es5/aes/cfb' from '/home/luke/git/openpgpjs-d60efff61d024421e7985e25e7156cb4856d9d9a/src/crypto'
Warning: Error running grunt-browserify. Use --force to continue.

Aborted due to warnings.
npm ERR! Test failed.  See above for more details.
```

twiss:
Hmm, that's weird. `npm install && npm run build` works fine for me, and the CI tests also do more or less that and passed.

> Error: Cannot find module 'asmcrypto.js/dist_es5/aes/cfb'

That also sounds pretty unrelated to this PR, and I don't really know what could cause that - maybe just delete `node_modules` and try again? :/

tomholub:
It works!

I have no clue how to treat decryption of such primary keys, and keep it not confusing. I see you went for `k.primaryKey.decrypt('anything') -> true`. That may be ok, though some weird behavior will follow:

```js
k.primaryKey.isDecrypted() // false 
k.primaryKey.decrypt('anything') // true
k.primaryKey.isDecrypted() // false
```
So I wonder if
```js
k.primaryKey.isDecrypted() // false 
k.primaryKey.decrypt('anything') // null
k.primaryKey.isDecrypted() // false
```
may be more appropriate. This may mean that `k.decrypt(...)` would have to be updated to either skip such primary keys, or skip over any nulls. I don't know which other code would need to be updated, so I have no idea how big a change would that be.

Really excited that this is finally happening!

twiss:
> It works!

Cool! 👍 

> I see you went for `k.primaryKey.decrypt('anything') -> true`. That may be ok, though some weird behavior will follow

Myeah. I was thinking of the return value as sort of a "success/failure" boolean, rather than "key was decrypted". But actually the function throws on failure, and `k.decrypt` doesn't even look at the return value contrary to what I thought, so I've changed it to just `return false` 🙃

test/general/signature.js
```diff
@@ -340,15 +340,12 @@ describe("Signature", function() {
...
-        expect(verified).to.have.length(1);
-        expect(await verified[0].verified).to.be.true;
-        expect((await verified[0].signature).packets.length).to.equal(1);
-      });
+    await priv_key_gnupg_ext.decrypt("abcd");
+    return openpgp.decrypt({ message: msg, privateKeys: [priv_key_gnupg_ext], publicKeys: [pub_key] }).then(function(msg) {
+      expect(msg.signatures).to.exist;
+      expect(msg.signatures).to.have.length(1);
+      expect(msg.signatures[0].valid).to.be.true;
+      expect(msg.signatures[0].signature.packets.length).to.equal(1);
```
sanjanarajan: Let's add a test for signing in case we ever switch the order of signing key packet selection to prioritize primary over subkeys

sanjanarajan: Actually also, maybe we should just add filtering for decrypted packets in the functions that choose the packet to use for signing/decrypting. For example, wouldn't the current behavior be weird if I imported a key with no primary key and no signing-capable subkey and tried to sign with it?

sanjanarajan: Adding the filtering may just make the error handling more consistent

twiss:
There's also the question of what `key.isDecrypted()` should return - currently it has the same weird behavior of
```js
key.isDecrypted() // false 
await key.decrypt('password') // true
key.isDecrypted() // false
```
It sort of depends on what the real question you want to ask is - e.g. if it's "can I sign something with this key?", that now depends on the subkeys. So it wouldn't be entirely unreasonable to add a `capabilities` param similar to the one `Key.prototype.getExpirationTime` has. Although I think the only really reliable way to find out the answer is to just try signing something and see if it fails.

Or if it's "should I encrypt this key before exporting" then it should be something more like `key.getKeys().some(key => key.isDecrypted())`.

Lmk if you have an opinion on this / any other usecases for that function.

test/general/signature.js
twiss: > Actually also, maybe we should just add filtering for decrypted packets in the functions that choose the packet to use for signing/decrypting.

We can't do that really easily because `getSigningKey` is also used for getting the public signing key, so we'd have to add a parameter for whether we want a decrypted key. I think just sorting by `isDecrypted()` is easier.

> For example, wouldn't the current behavior be weird if I imported a key with no primary key and no signing-capable subkey and tried to sign with it?

The current behavior would be throwing "Private key is not decrypted." when you try to sign with it, whereas with that change it would throw "Could not find valid signing key packet in key". I agree that in that specific case that would make sense, but it would also be a pretty weird key, and the same change would apply to the case where it's a regular old key with a perfectly valid signing key but you just forgot to decrypt it.

tomholub: `Could not find valid signing key packet in key` to me sounds like a matching error for encrypt-subkey-only type of key that @sanjanarajan mentions.

tomholub: But then `Private key is not decrypted` for this weird edge case of a key (missing primary, encrypt only subkey, trying to sign) is still miles better compared to previous behavior (which was: cannot do anything at all with any key unless primary key is present).

tomholub:
> Or if it's "should I encrypt this key before exporting" then it should be something more like `key.getKeys().some(key => key.isDecrypted())`.

I think this is the main usecase to optimize for, because here the stakes are the highest. Exporting a (partly) unprotected key that the developer thought was encrypted is a disaster, and this particular way of looking at the return values should probably overwhelm all other interpretations (imo). Less crucial situation: If I have a partly decrypted key that I thought was fully decrypted, and I use it, all I get is an exception (not so bad).

We could have our cake and eat it too, with some caveats, if the user-facing method throws with partially decrypted keys

```js
const { keys: [key] } = await openpgp.key.readArmored(' ... ');
if(key.isDecrypted()) { // could throw: "Inconsistent key state: only partially decrypted"
  // all available private information is certainly decrypted (empty primary keys ignored)
} else {
  // all available private information is certainly encrypted (empty primary keys ignored)
}
```

tomholub:
@twiss @sanjanarajan  Do you think that throwing on partially decrypted keys would be a reasonable way to resolve this?

If not, does it make sense to cater to the safety of exporting (unknowingly partly) unprotected keys, and have `key.isDecrypted()` return `true` if any if the subkeys are decrypted?

twiss:
@tomholub What I've settled on is to make `key.primaryKey.isDecrypted()` (and thus `key.isDecrypted()`) return true after calling `decrypt()` on it. This also makes sure we don't throw "Private key is not decrypted" in various places, instead we now throw "Missing private key parameters" which makes a bit more sense.

tomholub:
Awesome, I'm still planning to throw some bounties on other issues that pain us, as I discover them, to attract outside contributors. Thanks again!

PR #868, opened 2019-02-26
Fix merging multiple subkey binding signatures

### LINKED ISSUE #863: SubKey.prototype.update overwrites newer binding signature
We got an issue where subkey binding signatures are not merged correctly when using the Key.prototype.update method. I think the line `that.bindingSignatures[i] = srcBindSig;` at https://github.com/openpgpjs/openpgpjs/blob/master/src/key.js#L1187 has to be omitted. The `checkFn`of `mergeSignatures` should only return true or false but not change the signatures. And the purpose of the for loop at https://github.com/openpgpjs/openpgpjs/blob/master/src/key.js#L1184 is to determine if a binding signature on the destination key has a newer creation date as on the source  key in which case we do not need to merge.

toberndo:
Fixes #863

### DIFF
--- a/src/key.js
+++ b/src/key.js
@@ -1183,10 +1183,10 @@ SubKey.prototype.update = async function(subKey, primaryKey) {
     }
     for (let i = 0; i < that.bindingSignatures.length; i++) {
       if (that.bindingSignatures[i].issuerKeyId.equals(srcBindSig.issuerKeyId)) {
-        if (srcBindSig.created < that.bindingSignatures[i].created) {
+        if (srcBindSig.created > that.bindingSignatures[i].created) {
           that.bindingSignatures[i] = srcBindSig;
-          return false;
         }
+        return false;
       }
     }
     return true;
--- a/test/general/key.js
+++ b/test/general/key.js
@@ -2413,6 +2413,20 @@ describe('Key', function() {
     });
   });
 
+  it('update() - merge multiple subkey binding signatures', async function() {
+    const source = (await openpgp.key.readArmored(multipleBindingSignatures)).keys[0];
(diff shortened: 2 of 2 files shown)


twiss:
Hi @toberndo! Thanks for the PR, and sorry for the delay.

I imagine what the code should be doing is something like: "If a binding signature from the same key already exists, don't add it to the list. If the one in the passed key is newer, overwrite it." If that's correct, I think the patch should instead be something like:

```patch
       if (that.bindingSignatures[i].issuerKeyId.equals(srcBindSig.issuerKeyId)) {
-        if (srcBindSig.created < that.bindingSignatures[i].created) {
+        if (srcBindSig.created > that.bindingSignatures[i].created) {
           that.bindingSignatures[i] = srcBindSig;
-          return false;
         }
+        return false;
       }
```

However, that makes your test fail, unless I also change

```patch
-      expect(dest.subKeys[0].bindingSignatures.length).to.equal(2);
+      expect(dest.subKeys[0].bindingSignatures.length).to.equal(1);
```

So I might be wrong. Is there some reason the result should have two binding signatures?

toberndo:
> "If a binding signature from the same key already exists, don't add it to the list. If the one in the passed key is newer, overwrite it."

You are right, I think that's the way it should be. Actually we tested with Kleopatra and on merge here also only the newest binding signature is preserved. I don't see a reason to keep the older signature in the key.

Would you like me to change the PR? Or, if you have the change already done, please go on.

twiss:
Yeah, could you update this PR, please? Thanks!

toberndo:
@twiss Done :)

PR #879, opened 2019-03-31
Require uid self cert - configurable - updated

### LINKED ISSUE #740: gpg --allow-non-selfsigned-uid equivalent
PR #753

Unfortunately, there are still many pubkeys in use today (mostly coming from Symantec PGP Desktop) that do not have uid self signatures. Naturally, OpenPGP.js is refusing to use these for encryption.

I would like to make a PR that would allow this with a flag. I would like to be able to set an attribute on a `Key` object, eg `requireUidSelfSignature` to `false`. After that, when I do `k.getEncryptionKeyPacket()`, it will return one even if there are no self uid signatures. Same for `k.getPrimaryUser()` etc. This would allow such key to be used in `openpgp.encrypt` and elsewhere.

The flag would have to be set on each key individually, allowing the application know which key requires this and show warnings to the user.

This functionality would allow encrypting for legacy keys. Creation of such keys is deprecated, but using such keys should be possible with a flag / user warning for compatibility reasons. Would a PR for this be accepted?

tomholub:
replaces #753
fixes #740 

addressed comments:
https://github.com/openpgpjs/openpgpjs/pull/753#pullrequestreview-146325484
https://github.com/openpgpjs/openpgpjs/pull/753#discussion_r210166781
https://github.com/openpgpjs/openpgpjs/pull/753#discussion_r210176390

@sanjanarajan 

Thanks for the endless patience to get this done :roll_eyes:

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -123,6 +123,12 @@ export default {
 
   /**
    * @memberof module:config
+   * @property {Boolean} require_uid_self_cert If false, user is valid even without self certification
+   */
+  require_uid_self_cert: true,
+
+  /**
+   * @memberof module:config
    * @property {Boolean} use_native Use native Node.js crypto/zlib and WebCrypto APIs when available
    */
   use_native:               true,
--- a/src/key.js
+++ b/src/key.js
@@ -1735,8 +1735,7 @@ export async function getPreferredAlgo(type, keys, date=new Date(), userIds=[]) 
   const prioMap = {};
   await Promise.all(keys.map(async function(key, i) {
     const primaryUser = await key.getPrimaryUser(date, userIds[i]);
-    const noSelfCertOk = !config.require_uid_self_cert && primaryUser && !primaryUser.selfCertification;
-    if (!primaryUser || noSelfCertOk || !primaryUser.selfCertification[prefProperty]) {
+    if (!primaryUser || !primaryUser.selfCertification || !primaryUser.selfCertification[prefProperty]) {
       return defaultAlgo;
     }
     primaryUser.selfCertification[prefProperty].forEach(function(algo, index) {
--- a/src/key.js
+++ b/src/key.js
@@ -362,7 +362,7 @@ Key.prototype.getEncryptionKey = async function(keyId, date=new Date(), userId={
     }
     // if no valid subkey for encryption, evaluate primary key
     const primaryUser = await this.getPrimaryUser(date, userId);
-    if (primaryUser && (!keyId || primaryKey.getKeyId().equals(keyId)) &&
+    if (primaryUser && (!keyId || primaryKey.getKeyId().equals(keyId)) && primaryUser.selfCertification &&
         isValidEncryptionKeyPacket(primaryKey, primaryUser.selfCertification)) {
       return this;
     }
@@ -1699,7 +1699,7 @@ export async function getPreferredHashAlgo(key, keyPacket, date=new Date(), user
   let pref_algo = hash_algo;
   if (key instanceof Key) {
     const primaryUser = await key.getPrimaryUser(date, userId);
-    if (primaryUser && primaryUser.selfCertification.preferredHashAlgorithms) {
+    if (primaryUser && primaryUser.selfCertification && primaryUser.selfCertification.preferredHashAlgorithms) {
       [pref_algo] = primaryUser.selfCertification.preferredHashAlgorithms;
       hash_algo = crypto.hash.getHashByteLength(hash_algo) <= crypto.hash.getHashByteLength(pref_algo) ?
         pref_algo : hash_algo;
@@ -1774,7 +1774,7 @@ export async function isAeadSupported(keys, date=new Date(), userIds=[]) {
   // TODO replace when Promise.some or Promise.any are implemented
   await Promise.all(keys.map(async function(key, i) {
     const primaryUser = await key.getPrimaryUser(date, userIds[i]);
-    if (!primaryUser || !primaryUser.selfCertification.features ||
+    if (!primaryUser || !primaryUser.selfCertification || !primaryUser.selfCertification.features ||
         !(primaryUser.selfCertification.features[0] & enums.features.aead)) {
       supported = false;
     }
--- a/src/key.js
+++ b/src/key.js
@@ -316,23 +316,22 @@ Key.prototype.getSigningKey = async function (keyId=null, date=new Date(), userI
       }
     }
     const primaryUser = await this.getPrimaryUser(date, userId);
-    const noSelfCertOk = !primaryUser.selfCertification && !config.require_uid_self_cert;
     if (primaryUser && (!keyId || primaryKey.getKeyId().equals(keyId)) &&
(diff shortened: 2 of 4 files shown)


tomholub:
Ping, thanks :)

tomholub:
Another gentle ping

tomholub:
Let me know if there's anything I can do to move this forward. Thanks! :+1:

twiss:
Hi! :wave: Sorry for the delay. I've become a bit apprehensive to add this feature. I know it was basically already approved, but I think we should reconsider for a moment. The main reason is the following:

Currently, openpgp.js only places (and looks at) key preferences in UID self certifications, not in subkey self certs. So, with this option enabled, a MITM attacker can strip the key preferences from a key simply by removing the UID self cert, forcing us to fall back to the spec's defaults. This works even if the user verifies the fingerprint. Our current default preferred algorithm is AES256; we fall back to AES128, which is not so bad, even though the non-draft spec mandates falling back to 3DES, which would be pretty bad. Still, let's say someone discovers a vulnerability in our implementation of AES128, but not in AES256, this gives them a way to exploit that vulnerability. (We also can't easily change the fallback algorithm without changing the spec, but we can easily change our default, so these might divert further in the future.)

There's also some considerations around the primary key expiry, key flags, revocation key, etc, although I think that's less easily exploitable. In general, though, it's just nice to be able to verify all information in a key by verifying the fingerprint and a bunch of signatures.

I know client software should show a warning when using this option, and I know you will, but I'm afraid other developers will just globally enable this option when they first encounter a key like this. The problem with that is not that they'll accept that specific key, but rather that it exposes them to these unrelated subtle attacks on all other keys.

Even when a warning is shown, it's difficult for users to distinguish between MITM and a legacy key, and the existence of this option makes the path of least resistance to just use the key, rather than go check with the key owner why there's no self cert, which is what they should do.

Sorry for not bringing this up earlier. What do you think?

tomholub:
> Sorry for not bringing this up earlier. What do you think?

No worries, thank you for taking the time to verbalize it above.

> Even when a warning is shown, it's difficult for users to distinguish between MITM and a legacy key, and the existence of this option makes the path of least resistance to just use the key, rather than go check with the key owner why there's no self cert, which is what they should do.

This makes me think that I could further restrict it - only allow this option when emailing users on specific pre-defined domains (the ones that are known to be using such legacy keys). When sender wants a recipient domain whitelisted, they'd need to write the app developer (me) to verify they are using keys of this type, before they can use this functionality. Which brings me to..

> I'm afraid other developers will just globally enable this option when they first encounter a key like this

I have no good answer to this. One habit I've been following is to name things `DANGEROUS`, with some comments around it, which should give developers a pause: `openpgp.config.require_uid_self_cert_DANGEROUS_IF_FALSE = false;`, or perhaps maintain our own fork.

> Currently, openpgp.js only places (and looks at) key preferences in UID self certifications, not in subkey self certs. So, with this option enabled, a MITM attacker can strip the key preferences from a key simply by removing the UID self cert, forcing us to fall back to the spec's defaults. This works even if the user verifies the fingerprint. [...]

If I read this correctly, in a future scenario where there was a vulnerability discovered in the AES128 implementation, you are worried about a MITM attacker being able to trick users into using AES128 instead of AES256 to encrypt their outgoing messages.

In general, MITM attacker can replace and manipulate exchanged pubkeys at will, and users can protect themselves against this by diligently cross-checking fingerprints. You are worried that fingerprint checks don't help in this scenario - coupled with app developers not being diligent in showing alerts or users being lazy: increasing the attack scope of the app for non-legacy keys too.

> Our current default preferred algorithm is AES256; we fall back to AES128, which is not so bad [...]. Still, let's say someone discovers a vulnerability in our implementation of AES128, but not in AES256, this gives them a way to exploit that vulnerability.

Here I don't fully follow. If there are no self certs, I would expect OpenPGP.js to use default algo - AES256. Based on what would it decide to use AES128 instead when given no preference options?

> (We also can't easily change the fallback algorithm without changing the spec, but we can easily change our default, so these might divert further in the future.)

I'm trying to find relevant spec - I see that https://tools.ietf.org/html/rfc4880#section-13.2 considers missing preferences as `3DES`. OpenPGP.js default is AES256 (intentionally not following spec - a good choice I think). Am I looking at the right spec? Cannot find the word "fallback" in it.

**Thanks for looking into it!**

twiss:
> Thanks for looking into it!

No problem 😊

> If there are no self certs, I would expect OpenPGP.js to use default algo - AES256. Based on what would it decide to use AES128 instead when given no preference options?

OpenPGP.js used to do that, but we changed it a while back. The draft version of the spec replaces 3DES with AES128 in the section you found: https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-07#section-14.2. This is hardcoded here: https://github.com/openpgpjs/openpgpjs/blob/b035f2ea4638c4cb52b736ec5ec38181add38167/src/key.js#L1724-L1726

The "default algorithm to use when a key doesn't specify preferences" and the "default algorithm to put in the preferences of keys we generate" are thus now separate, with only the latter being configurable, as it should be.

> This makes me think that I could further restrict it - only allow this option when emailing users on specific pre-defined domains (the ones that are known to be using such legacy keys). When sender wants a recipient domain whitelisted, they'd need to write the app developer (me) to verify they are using keys of this type, before they can use this functionality.

Interesting. I didn't realize there were entire domains like that, I thought it was just individual users with old keys. Do you have examples of domains like that?

Also, GPG 2.2.10 will still use 3DES for a key like this. Presumably, if it's an entire domain, they're running pretty old software to decrypt messages as well - are you sure that they'll accept AES128?

> One habit I've been following is to name things DANGEROUS, with some comments around it, which should give developers a pause: `openpgp.config.require_uid_self_cert_DANGEROUS_IF_FALSE = false;`

Yeah, that'd definitely be an improvement. I guess the other half of that would be to clearly document all the risks.

> or perhaps maintain our own fork.

Yeah, maybe that's the safest option. Perhaps, if you have a dialog for a while telling people to email you with domains that do this, we could try to reach out to the owners of those domains and get them to update their software? Then hopefully at some point in the future this option won't be necessary anymore.

tomholub:
Thanks for the explanation, now I see how it would use AES128.

> I didn't realize there were entire domains like that, I thought it was just individual users with old keys. Do you have examples of domains like that?

If it were individual users, I'd just tell them to upgrade, or otherwise not worry about it much. I cannot share a particular domain (privacy), but it's large orgs, and they are slow moving..

> Perhaps, if you have a dialog for a while telling people to email you with domains that do this, we could try to reach out to the owners of those domains and get them to update their software?

I hope so.

> Presumably, if it's an entire domain, they're running pretty old software to decrypt messages as well - are you sure that they'll accept AES128?

I'll test AES128 as well as AES256, and see what can they decrypt.

-----------------

Looks like we'll be forking :) That way we minimize the chances of another app developer enabling this option without due diligence. Thanks for the comments!

twiss:
Thanks! Good luck, and let me know if you discover anything ~on your travels~ with the experiment ;)

PR #889, opened 2019-04-29
Add Node 12 to CI

twiss:


twiss:
Hey :wave: No particular rationale - since it's only used in the browser tests I thought it didn't really matter, but if you'd like to bump it up I'd be happy to merge a PR for that. Due to the way that `allow_failures` matching works, you could even do something like `"12.0"` or `"node"` (latest version) I think - as long as the node test for 12 and the browser tests have a different string value, the `allow_failures` should work correctly. Or you could do it the other way around - `node` instead of `12` for testing node, and then `12` instead of `9` for testing browsers, if you prefer.

PR #892, opened 2019-05-01
Fix uid format

### LINKED ISSUE #891: uid generated by address-rfc2822 is not compatible with GnuPG
`address-rfc2822` generates uid (reformatted address) in the following format:
```
$name <$email> ($comment)
```
Though it's legal per RFC, `GnuPG` prefers a different format, which is also more nature:
```
$name ($comment) <$email>
```
As a result, `GnuPG` treats the entire `$name <$email>` part as the name, causing email to be empty:

image

```
$ gpg --list-keys

/Users/si/.gnupg/pubring.kbx
----------------------------
pub   ed25519 2019-05-01 [SC]
      98526DA660B1E0AB0664F84A07E8AEA47F54E033
uid           [ultimate] Si Feng <_@feng.si> (冯思)
sub   cv25519 2019-05-01 [E]
```

**And this is what should show up in GnuPG if the uid is formatted correct:**

image

```
$ gpg --list-keys

/Users/si/.gnupg/pubring.kbx
----------------------------
pub   ed25519 2019-05-01 [SC]
      41D7B303919795417785F37D0D1DF53EE67E5867
uid           [ultimate] Si Feng (冯思) <_@feng.si>
sub   cv25519 2019-05-01 [E]
```

fengsi:
This fixes #891.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -652,11 +652,23 @@ export default {
    * Format user id for internal use.
    */
   formatUserId: function(id) {
-    // name and email address can be empty but must be of the correct type
-    if ((id.name && !util.isString(id.name)) || (id.email && !util.isEmailAddress(id.email))) {
+    // name, email address and comment can be empty but must be of the correct type
+    if ((id.name && !util.isString(id.name)) ||
+        (id.email && !util.isEmailAddress(id.email)) ||
+        (id.comment && !util.isString(id.comment))) {
       throw new Error('Invalid user id format');
     }
-    return new rfc2822.Address(id.name, id.email, id.comment).format();
+    const components = [];
+    if (id.name) {
+      components.push(id.name);
+    }
+    if (id.comment) {
+      components.push(`(${id.comment})`);
+    }
+    if (id.email) {
+      components.push(`<${id.email}>`);
+    }
+    return components.join(' ');
   },
 
   /**
--- a/test/general/key.js
+++ b/test/general/key.js
@@ -1737,7 +1737,49 @@ function versionSpecificTests() {
     return openpgp.generateKey(opt).then(function(key) {
       key = key.key;
       expect(key.users.length).to.equal(1);
-      expect(key.users[0].userId.userid).to.equal('test <a@b.com> (test comment)');
+      expect(key.users[0].userId.userid).to.equal('test (test comment) <a@b.com>');
+      expect(key.users[0].userId.name).to.equal(userId.name);
+      expect(key.users[0].userId.email).to.equal(userId.email);
+      expect(key.users[0].userId.comment).to.equal(userId.comment);
+    });
+  });
+
+  it('Generate key - single userid (all missing)', function() {
+    const userId = { name: '', email: '', comment: '' };
+    const opt = {numBits: 512, userIds: userId, passphrase: '123'};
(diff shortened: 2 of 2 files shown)


twiss:
:+1: Thanks!

toberndo:
RFC2822 defines [quoted strings](https://tools.ietf.org/html/rfc2822#section-3.2.5). A name that has special characters will be put in `"`. rfc2822-address does this here: https://github.com/haraka/node-address-rfc2822/blob/master/index.js#L107
Was it intentional that this PR removes quoting of names?

fengsi:
@toberndo no - this PR only tried to fix comment location per GnuPG conventions.  Please feel free to fix the name quoting.  Sorry for missing that.
BTW, regarding [RFC 2822](https://tools.ietf.org/html/rfc2822) - shouldn't we use [RFC 5322](https://tools.ietf.org/html/rfc5322) instead?  That was one of the reasons I wanted to get rid of `address-rfc2822`...

twiss:
We could also fork address-rfc2822, switch the order there, and go back to using that. Or use a different library. Proposals or PRs welcome, but I'll probably do the former if nobody makes one ^.^

toberndo:
To my understanding GPG does not do any quoting. And as a result applications based on GPG like GPGTools will show quoted names between `"` (which is problematic in the UI when you want to sort by name). So in that sense this PR aligns OpenPGP.js with the GPG encoding.
So far in Mailvelope we have used the Google Closure library for name-addr parsing which seems to closer to RFC5322 (?). For this I extracted the `goog.format.EmailAddress` with all dependencies into a new file: https://github.com/mailvelope/mailvelope/blob/master/src/modules/closure-library/closure/goog/emailaddress.js. But it does not support a separate comment property.

twiss:
> To my understanding GPG does not do any quoting. And as a result applications based on GPG like GPGTools will show quoted names between " (which is problematic in the UI when you want to sort by name). So in that sense this PR aligns OpenPGP.js with the GPG encoding.

Thanks for the info. That then makes me think we should keep it as-is, although we should probably also throw when trying to generate a user ID with invalid characters (e.g. `"`) in names and such.

> So far in Mailvelope we have used the Google Closure library for name-addr parsing

OpenPGP.js exposes address-rfc2822's parser as `openpgp.util.parseUserId`, which seems to support both, perhaps you could use that? It also parses user IDs by default, you can access the result with `key.users[*].userId.name`, `.comment`, `.email`.

toberndo:
> OpenPGP.js exposes address-rfc2822's parser as openpgp.util.parseUserId, which seems to support both, perhaps you could use that?

It does not parse some edge cases correctly, in my set of test keys I got better results with `goog.format.EmailAddress` (would need to check what exactly was the issue). Currently we check if `key.users[*].userId.name` or `.email` is available and if not fall back to `goog.format.EmailAddress`.

By the way, a related issue: I think it is not optimal to use at https://github.com/openpgpjs/openpgpjs/blob/master/src/key.js#L1481 the format method (which uses util.parseUserId and util.formatUserId) to create a reformatted userId. For a userId  `<text@example.com>` for example the input userId will not be the same as the output. Here I think the original userId should be used, without parsing.

PR #899, opened 2019-05-09
add options.fetchOpts to HKP.lookup() method for fetch() call

nicholatian:
Hi. This is a quick PR to allow secure key server lookups to work with custom CAs, using the agent option in fetch(). I need to use SKS’ custom CA for HKPS, and prior to this change OpenPGP.js inadvertently obstructs me from informing `fetch` of these CAs to accept. If there’s anything else needing done please let me know, but otherwise I look forward to seeing this merged for everyone.


twiss:
Hi! It seems like you could monkey-patch this by doing something like
```js
const hkp = new openpgp.HKP();
const fetch = hkp._fetch;
hkp._fetch = url => fetch(url, fetchOptions);
hkp.lookup(...);
```
The option you mentioned is specific to node-fetch, whereas OpenPGP.js can also be used in the browser, where it is not available. So adding an option just for that seems a bit overkill if we can do without.

nicholatian:
Alright, noted. I managed to get HKPS working some other way still, so I guess this wasn't needed after all. Thanks for the response! Closing...

Issue #897, opened 2019-05-07
Support for multiple notations with the same key


wiktor-k:
Hello,

Currently it's impossible to access multiple [notations](https://tools.ietf.org/html/rfc4880#section-5.2.3.16) if they have the same key as the last one always overwrites all previous values.

I've added multiple signature notations with `gpg` and they are displayed well in CLI and in packet dump:

```
$ gpg --list-packets test.gpg 
# off=0 ctb=99 tag=6 hlen=3 plen=269
:public key packet:
	version 4, algo 1, created 1557150010, expires 0
	pkey[0]: [2048 bits]
	pkey[1]: [17 bits]
	keyid: 579D01B3ABE1540E
# off=272 ctb=b4 tag=13 hlen=2 plen=16
:user ID packet: "Testt <test@exa>"
# off=290 ctb=89 tag=2 hlen=3 plen=394
:signature packet: algo 1, keyid 579D01B3ABE1540E
(...truncated...)
	hashed subpkt 33 len 21 (issuer fpr v4 198C722A4BAC336E9DAAAE44579D01B3ABE1540E)
	hashed subpkt 2 len 4 (sig created 2019-05-06)
	hashed subpkt 20 len 25 (notation: test@example.com=2) <---- here
	hashed subpkt 20 len 25 (notation: test@example.com=3) <---- and here
	subpkt 16 len 8 (issuer key ID 579D01B3ABE1540E)
	data: [2048 bits]
```

Reading the same key with OpenPGP.js with this snippet of code:

```js
openpgp.key.read(fs.readFileSync('test.gpg')).then(k => k.keys[0].users[0].selfCertifications).then(console.log.bind(console), console.error.bind(console));
```

Prints the following info:

```
[ Signature {
    tag: 2,
    version: 4,
(...truncated...)
    issuerKeyId: Keyid { bytes: 'W },
    notation: { 'test@example.com': '3' }, <--- here
    preferredHashAlgorithms: [ 10, 9, 8, 11, 2 ],
    preferredCompressionAlgorithms: [ 2, 3, 1 ],
```

As you can see only the last notation value for that key was preserved (`test@example.com=3`) while the second one (`test@example.com=2`) is nowhere to be found.

[RFC 4880](https://tools.ietf.org/html/rfc4880#section-5.2.3.16) doesn't have any guidance as of what to do when there are multiple notations with the same key.

I see three possible solutions to this problem:
[...]

Issue #807, opened 2018-11-28
OpenPgpJs version 4.2.1 Decrypt fails on verify signature with error 'Error decrypting message: Corresponding signature packet missing at OnePassSignature.verify (node_modules/openpgp/dist/openpgp.js:35786:11)'


LukeMcCarthy-TW:
When trying to decrypt a message I get the following error

Error: Error decrypting message: Decryption failed due to missing MDC.
  at SymmetricallyEncrypted.decrypt (node_modules/openpgp/dist/openpgp.js:38899:11)
at <anonymous>

If I set
```
  openpgp.config.ignore_mdc_error = true
```
I get the following error

  Error: Error decrypting message: Corresponding signature packet missing
at OnePassSignature.verify (node_modules/openpgp/dist/openpgp.js:35786:11)
at <anonymous>
at process._tickCallback (internal/process/next_tick.js:188:7)

If I remove the public key from the options then the file decrypts correctly, however, I then don't know who sent the message

The file is sent from another organisation and it would be very difficult to change anything on their side

Any help would be greatly appreciated

Sample code below

```
const expect = require('chai').expect;
const openpgp = require('openpgp')

const pgpKeys = require('./pgp_keys_decrypt_unit_test');

const encryptedText = `-----BEGIN PGP MESSAGE-----
Version: SDS E-Business Server - 64 Bit - v9.0.1 - Full License

hQEMA1Fa0FFyaYAZAQgAouMX/4Vd0JWrWsmR6mzT/RLyK16Sd449R6r0H74msyas
sX6YI5Qu+inIJxWQhLF5fgaxr7+N9mwKM/G2FrErG2MDT/XTwjH9WOH4AudK9lcX
NWO2ELJyxwOJhqBlxL0bzjpmL83uufIjfEA7+QsTzGnMgg122aExVs2CVacxQob1
mouEpA7hp3HHBosrO/zIX773sVrOknlYNbzB6lX/+5l14FN19HRXCI7vQIX0G/D1
lF+kKmOodIZFD+5k/J1FDBwwOuHXai86Nv/l0ktvUG18yXrAxq96sTFgZ6xI8Yh0
4mUlZ2Pr2o82Vr5hBbyuLLmD7FMnyg+lubDwxkJXhtLAswHWpzrQ3gMlZVdrneN/
fZmp1yVJWSCTo0KguUZ7OZfjBMOrFPIwk02934EdZ0jP73OMVYS0/06vavRSFDIl
[...]

PR #901, opened 2019-05-09
Expose all signature notations

wiktor-k:
Previous implementation used an object to hold notations so if multiple notations had the same key name only the last one was visible.

After this change notations are exposed as an array of key-value pairs that can be converted to a map through `new Map(notations)`.

See #897.

If this change looks good I'll follow up with fixes to other issues mentioned by @twiss in #897.

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -183,5 +183,12 @@ export default {
    * @memberof module:config
    * @property {Integer} max_userid_length
    */
-  max_userid_length: 1024 * 5
+  max_userid_length: 1024 * 5,
+  /**
+   * Contains notatations that are considered "known". Known notations do not trigger
+   * validation error when the notation is marked as critical.
+   * @memberof module:config
+   * @property {Array} known_notations
+   */
+  known_notations: ["preferred-email-encoding@pgp.com", "pka-address@gnupg.org"]
 };
--- a/src/packet/signature.js
+++ b/src/packet/signature.js
@@ -70,7 +70,7 @@ function Signature(date=new Date()) {
   this.revocationKeyAlgorithm = null;
   this.revocationKeyFingerprint = null;
   this.issuerKeyId = new type_keyid();
-  this.notation = null;
+  this.notations = [];
   this.preferredHashAlgorithms = null;
   this.preferredCompressionAlgorithms = null;
   this.keyServerPreferences = null;
@@ -228,18 +228,16 @@ Signature.prototype.write_hashed_sub_packets = function () {
     bytes = util.concat([bytes, this.revocationKeyFingerprint]);
     arr.push(write_sub_packet(sub.revocation_key, bytes));
   }
-  if (this.notation !== null) {
-    Object.entries(this.notation).forEach(([name, value]) => {
-      bytes = [new Uint8Array([0x80, 0, 0, 0])];
-      // 2 octets of name length
-      bytes.push(util.writeNumber(name.length, 2));
-      // 2 octets of value length
-      bytes.push(util.writeNumber(value.length, 2));
-      bytes.push(util.str_to_Uint8Array(name + value));
-      bytes = util.concat(bytes);
-      arr.push(write_sub_packet(sub.notation_data, bytes));
-    });
-  }
+  this.notations.forEach(([name, value]) => {
+    bytes = [new Uint8Array([0x80, 0, 0, 0])];
+    // 2 octets of name length
+    bytes.push(util.writeNumber(name.length, 2));
+    // 2 octets of value length
+    bytes.push(util.writeNumber(value.length, 2));
+    bytes.push(util.str_to_Uint8Array(name + value));
+    bytes = util.concat(bytes);
+    arr.push(write_sub_packet(sub.notation_data, bytes));
+  });
   if (this.preferredHashAlgorithms !== null) {
     bytes = util.str_to_Uint8Array(util.Uint8Array_to_str(this.preferredHashAlgorithms));
     arr.push(write_sub_packet(sub.preferred_hash_algorithms, bytes));
@@ -447,8 +445,7 @@ Signature.prototype.read_sub_packet = function (bytes, trusted=true) {
         const name = util.Uint8Array_to_str(bytes.subarray(mypos, mypos + m));
         const value = util.Uint8Array_to_str(bytes.subarray(mypos + m, mypos + m + n));
 
-        this.notation = this.notation || {};
-        this.notation[name] = value;
+        this.notations.push([name, value]);
       } else {
         util.print_debug("Unsupported notation flag "+bytes[mypos]);
       }
--- a/src/packet/signature.js
+++ b/src/packet/signature.js
@@ -32,6 +32,7 @@ import type_mpi from '../type/mpi.js';
 import crypto from '../crypto';
 import enums from '../enums';
 import util from '../util';
+import config from '../config';
 
 /**
  * Implementation of the Signature Packet (Tag 2)
@@ -446,6 +447,10 @@ Signature.prototype.read_sub_packet = function (bytes, trusted=true) {
         const value = util.Uint8Array_to_str(bytes.subarray(mypos + m, mypos + m + n));
 
         this.notations.push([name, value]);
+
+        if (critical && (config.known_notations.indexOf(name) === -1)) {
+          throw new Error("Unknown critical notation: " + name);
+        }
       } else {
         util.print_debug("Unsupported notation flag "+bytes[mypos]);
       }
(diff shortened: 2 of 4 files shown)


twiss:
Looks good to me so far :+1:

wiktor-k:
Sorry for the delay but I want to get good test samples first.

GnuPG allows creating critical notations but the list of known notations [can be modified before signature validation](https://dev.gnupg.org/T4060). That is the application doing signature verification can mark which notations are known and should not fail the validation even when encountered (`gpg --known-notation`).

As for non-human readable notations I didn't see an option in GnuPG so I guess I'd have to modify OpenPGP.js to emit them to save them as test samples :)

In the mean time the built with this PR applied successfully allows me to render [keys with multiple notations](https://keyserver.ubuntu.com/pks/lookup?op=vindex&fingerprint=on&search=0x579D01B3ABE1540E) with the same name: https://metacode.biz/openpgp/key#0x579D01B3ABE1540E (this is rendered only via OpenPGP.js from the key, there are no server side-components).

twiss:
> GnuPG allows creating critical notations but the list of known notations can be modified before signature validation. That is the application doing signature verification can mark which notations are known and should not fail the validation even when encountered (gpg --known-notation).

The equivalent of this that I would do is to make a config option, an array that defaults to the names known to GPG, and that the application can push other values to.

> As for non-human readable notations I didn't see an option in GnuPG so I guess I'd have to modify OpenPGP.js to emit them to save them as test samples :)

If no one uses it and you don't have a use case either, feel free to leave this be, and continue to throw for non-text values, if you like :)

> In the mean time the built with this PR applied successfully allows me to render keys with multiple notations with the same name: https://metacode.biz/openpgp/key#0x579D01B3ABE1540E (this is rendered only via OpenPGP.js from the key, there are no server side-components).

Cool! ^.^

wiktor-k:
> The equivalent of this that I would do is to make a config option, an array that defaults to the names known to GPG, and that the application can push other values to.

Good. That's exactly what I was thinking but I did want to hear your opinion untainted.

> If no one uses it and you don't have a use case either, feel free to leave this be, and continue to throw for non-text values, if you like :)

Sounds like a plan :+1: although now I'm curious what GnuPG will do when it encounters this kind of notation... damn you curiosity!

wiktor-k:
Okay, I did use the approach of collecting critical notation names in an array and then failing the validation. Throwing an error while reading the key produces really bizarre errors.

Note that this doesn't write critical bit for notations (I don't need it but just mentioning).

twiss:
Failing the validation is clever :+1: though it raises the question why we're not doing that for critical signature subpackets in general. Probably what you're seeing when throwing during reading is that `packetlist.read` just silently drops the packet, which leads to other errors down the line. You can set `openpgp.config.tolerant = false` to fix this, but you don't necessarily want that in all contexts; you can set `openpgp.config.debug = true` to at least see the error in the console, but I also understand that you'd want access to it in code.

There's at least one other thing we should throw on signature verification rather than reading: https://github.com/openpgpjs/openpgpjs/blob/54f3eb5870e4e0611965003e276af786ec419470/src/key.js#L1672 (#808 point 3). So I think your solution is good precedent despite the inconsistency with critical subpackets for now.

On to the more nitpicky stuff: I believe you _can_, and probably should just throw in `signature.verify`. `openpgp.verify` has a try/catch which, in non-streaming mode, sets https://github.com/openpgpjs/openpgpjs/blob/54f3eb5870e4e0611965003e276af786ec419470/src/openpgp.js#L694-L695
For streaming mode and when using `message.verify` directly, `await signature.verified` will throw. I think all of that is relatively reasonable.

There's also a subtle issue with sticking an error object on the signature packet which is that it'll break the worker unless you handle it specifically. https://github.com/openpgpjs/openpgpjs/pull/894/files?w=1#diff-26a5faed328afd2f1b0e5800b759ec23 added a small number of tests for this, you could add yours to those to see.

wiktor-k:
> Failing the validation is clever +1 though it raises the question why we're not doing that for critical signature subpackets in general.

I mimicked the behavior of GnuPG w.r.t critical packets. As far as I remember X.509 and some other protocols also handle critical stuff in the same manner....

> Probably what you're seeing when throwing during reading is that packetlist.read just silently drops the packet, which leads to other errors down the line. You can set openpgp.config.tolerant = false to fix this, but you don't necessarily want that in all contexts; you can set openpgp.config.debug = true to at least see the error in the console, but I also understand that you'd want access to it in code.

Yes, and thanks for tips for debugging. I missed them :+1:

> On to the more nitpicky stuff: I believe you can, and probably should just throw in signature.verify. openpgp.verify has a try/catch which

Good idea, I'll rework the patch.

> There's also a subtle issue with sticking an error object on the signature packet which is that it'll break the worker unless you handle it specifically.

Okay, first I'll adjust the commit and then I'll see what's up with workers. I guess errors are not Transferrable objects :man_shrugging:

twiss:
> I mimicked the behavior of GnuPG w.r.t critical packets.

Alrighty, thanks for the info. Yeah we should do that for other critical subpackets too.

> I guess errors are not Transferrable objects

Hmhm. I think if you throw in `verify` it should work out by itself, that case is handled since the PR linked above.

wiktor-k:
> Alrighty, thanks for the info. Yeah we should do that for other critical subpackets too.

Just for the record the [relevant passage of RFC 4880](https://tools.ietf.org/html/rfc4880#section-5.2.3.1):

>   Bit 7 of the subpacket type is the "critical" bit.  If set, it
>   denotes that the subpacket is one that is critical for the evaluator
>   of the signature to recognize.  If a subpacket is encountered that is
>   marked critical but is unknown to the evaluating software, the
>   evaluator SHOULD consider the signature to be in error.

twiss:
Yeah :+1: I mean, not that I disagree with you, but that section doesn't necessarily say much about whether we should reject signatures with unknown critical subpackets during parsing (like we do now) or verification, nor should it necessarily make a difference, right?

I guess the main practical difference (other than what kind of error shows up if you have one signature) is what happens if you have e.g. two self-signatures, and only the newest one has the critical subpacket. Currently we'd use the old self-signature and the key would be valid, afterwards we'd use the new one and the key would be invalid. That's because we currently only try the latest self-signature, I think unlike GnuPG, which I believe will use the older signature packet even on regular old verification failure of the new one - perhaps we should also do that, which would make this distinction largely moot.

At that point, if we still want to throw on parse for simplicity but still have reasonable error messages, we could think about a mechanism to store the error somewhere and then later on throw that instead of "missing signature" - that would also perhaps be useful for https://github.com/openpgpjs/openpgpjs/issues/807#issuecomment-442626252.

wiktor-k:
Okay Daniel, excuse me zooming out of this discussion for a moment but are we reaching some kind of solution w.r.t. critical notations here? As far as I understood it throwing on signature parse would indeed produce bad errors but at least it would leave the key usable. Is this the direction you want it to go now?

I initially didn't want to mix in changes about multiple-key notations and further changes (critical bit & human-readable flag) because multiple-key notations, in my opinion, are good to be merged and currently the critical bit changes are blocking the first set of changes.

While I do enjoy discussion about solving it the optimal way I have a feeling it will take a while to adjust the code base and as you can see I'm quite slow (a lot of work with other endeavors recently, sorry :bowing_man: ).

twiss:
Yes, apologies for the stream of consciousness.

> As far as I understood it throwing on signature parse would indeed produce bad errors but at least it would leave the key usable. Is this the direction you want it to go now?

Yes, I think that'd be best for now. If we still want to test the error we could set `openpgp.config.tolerant = false` in the test case for now.

> I initially didn't want to mix in changes

Ah ok, that wasn't clear to me, sorry. Do you want me to merge the first commit now and then follow up later, if you have time? Or I can also make the changes proposed above in this PR, if you prefer. You've already done most of the work, after all (thanks!) ^.^

wiktor-k:
> Yes, apologies for the stream of consciousness.

No problem! I actually enjoy discussions about the bigger picture and had some ideas for more improvements in OpenPGP.js myself but currently I'm writing some tooling for [my proof system](https://metacode.biz/openpgp/key#0x653909A2F0E37C106F5FAF546C8857E0D8E8F074) and I need notations fixed on npm :)

> Yes, I think that'd be best for now. If we still want to test the error we could set openpgp.config.tolerant = false in the test case for now.

Fixed to do just that.

> You've already done most of the work, after all (thanks!) ^.^

Yep, no problem. I guess we're close to the finish line :checkered_flag:

twiss:
:+1: Thanks a lot!

wiktor-k:
No problem. Thank you for taking the time to review and discuss it!

PR #911, opened 2019-06-21
Node detection: base on process instead of window

schantaraud:


twiss:
Thanks! Could you un-commit the `dist` files? They'll be updated at release.

OTOH, feel free to update https://github.com/openpgpjs/openpgpjs/blob/b035f2ea4638c4cb52b736ec5ec38181add38167/package.json#L83 and `npm-shrinkwrap.json` in this PR, if you like.

PR #921, opened 2019-07-01
Node crypto ecdh

chesnokovilya:


src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -340,10 +353,50 @@ async function ellipticPublicEphemeralKey(curve, Q) {
...
 }
 
+/**
+ * Generate ECDHE secret from private key and public part of ephemeral key using nodeCrypto
+ *
+ * @param  {Curve}                  curve          Elliptic curve object
+ * @param  {Uint8Array}             V              Public part of ephemeral key
+ * @param  {Uint8Array}             d              Recipient private key
+ *
+ * @returns {Promise<{secretKey: Uint8Array, sharedKey: Uint8Array}>}
```
twiss: I agree that these extra empty lines make it nice and tidy, but, these comments are already very large, and personally I almost never look at them, so it's a bit a wasted space. Also I think it's better to improve the parameter names instead of improving these comments (don't have to do that now, though). It's a bit nitpicky but could you remove the empty lines?

src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -340,10 +353,50 @@ async function ellipticPublicEphemeralKey(curve, Q) {
...
+ * @param  {Uint8Array}             Q            Recipient public key
+ *
+ * @returns {Promise<{publicKey Uint8Array, sharedKey Uint8Array}>}
+ * @async
+ */
+async function nodePublicEphemeralKey(curve, Q)
+{
+  const sender = nodeCrypto.createECDH(curve.node.node);
+  sender.generateKeys();
+  //below would throw different errors depending on version of node
```
twiss: It's probably better to explain this in the test case (something like "computeSecret inside nodePublicEphemeralKey throws different...") instead of here, as it's otherwise not really relevant to this code.

src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -259,8 +269,9 @@ async function webPrivateEphemeralKey(curve, V, Q, d) {
  * Generate ECDHE ephemeral key and secret from public key using webCrypto
  *
  * @param  {Curve}                  curve        Elliptic curve object
- * @param  {Uint8Array}             Q                   Recipient public key
- * @returns {Promise<{V: Uint8Array, S: BN}>}   Returns public part of ephemeral key and generated ephemeral secret
+ * @param  {Uint8Array}             Q            Recipient public key
+ *
+ * @returns {Promise<{publicKey Uint8Array, sharedKey Uint8Array}>}
```
twiss:
```suggestion
 * @returns {Promise<{publicKey: Uint8Array, sharedKey: Uint8Array}>}
```
Some other functions too

test/crypto/elliptic.js
```diff
@@ -500,10 +499,28 @@ describe('Elliptic Curve Cryptography', function () {
     return Z;
   }
 
+  async function genNodePrivateEphemeralKey(curve, V, d, fingerprint) {
+    const curveObj = new openpgp.crypto.publicKey.elliptic.Curve(curve);
+    const oid = new openpgp.OID(curveObj.oid);
+    const { sharedKey } = await openpgp.crypto.publicKey.elliptic.ecdh.nodePrivateEphemeralKey(
```
twiss: Could you deduplicate these functions a bit? E.g. by passing 'nodePrivateEphemeralKey' as the first parameter and then doing something like
```suggestion
    const { sharedKey } = await openpgp.crypto.publicKey.elliptic.ecdh[functionName](
```
Then it makes it more clear that they're all doing the same thing

test/crypto/elliptic.js
```diff
@@ -464,30 +463,26 @@ describe('Elliptic Curve Cryptography', function () {
...
+    switch (fun) {
+      case 'webPrivateEphemeralKey': {
+        result = await openpgp.crypto.publicKey.elliptic.ecdh[fun](
+          curveObj, V, Q, d
+        );
+        break;
+      }
+      case 'nodePrivateEphemeralKey':
+      case 'ellipticPrivateEphemeralKey': {
+        result= await openpgp.crypto.publicKey.elliptic.ecdh[fun](
```
twiss:
```suggestion
        result = await openpgp.crypto.publicKey.elliptic.ecdh[fun](
```

PR #928, opened 2019-07-08
adds support to reuse primaryKeyPacket when generating new keys

jpbochi:
I was looking for a way to generate new subkeys for an existing key. In order for me to `update` an existing key, the fingerprint of the newly generated key has to match the one with the new subkeys. The only way to support that, as far as I could tell, was to reuse the primary key of the key to be updated. And this is not currently supported by the library. Here's my implementation.

If there's a exsiting way to generate new subkeys and add them to an existing key that I couldn't find, please tell me.

### DIFF
--- a/src/key.js
+++ b/src/key.js
@@ -643,6 +643,7 @@ Key.prototype.update = async function(key) {
       this.subKeys.push(srcSubKey);
     }
   }));
+  return this;
 };
 
 /**
@@ -1366,6 +1367,14 @@ export async function generate(options) {
   }
 
   async function generateSecretKey(options) {
+    if (options.primaryKeyPacket) {
+      if (options.primaryKeyPacket.tag !== enums.packet.secretKey) {
+        throw new Error(
+          'primaryKeyPacket.tag must be enums.packet.secretKey, was ' + enums.read(enums.packet, options.primaryKeyPacket.tag) + '.'
+        );
+      }
+      return options.primaryKeyPacket;
+    }
     const secretKeyPacket = new packet.SecretKey(options.date);
     secretKeyPacket.packets = null;
     secretKeyPacket.algorithm = enums.read(enums.publicKey, options.algorithm);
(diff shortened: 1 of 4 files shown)


twiss:
This is a clever approach to generate subkeys for a key, but I think it might have some unwanted side-effects. For example, it re-generates the primary key's self-signature. If you're gonna upload the updated key to a keyserver, you'll want it to still have the old self-signature, as you can only add packets, not change them.

I think it wouldn't be unreasonable to add a `Key.prototype.addSubkey(s?)` function, and extract some of the common functionality with `generate()` into private helper functions. What do you think?

jpbochi:
@twiss Yeah. It isn't perfect. I really wanted to minimize the surface area of the change. Would your `Key.prototype.addSubkey(...)` generate new subkeys or add existing ones generated in some other method?

To clarify, what I did in my case was more or less like this:

```
const tempKey = openpgp.generateKey({ primaryKeyPacket: key.primaryKey, ...otherArgs });
tempKey.key.users.pop();
existingKey.update(tempKey.key);
```

The second line avoids any merging with the new user data being generated. It's definitely a bit wasteful. A `Key.prototype.generateAndMergeSubKeys` function would be the best way to avoid unnecessary computation.

twiss:
Hey @jpbochi, sorry for the delay. Yeah, I was imagining a single function that would return a clone of the key with a newly generated subkey added to it; most methods like `toPublic` and `revoke` currently return a clone too. It's difficult to come up with a perfect name for this, but I think `generateAndMergeSubKeys` is kinda verbose 🙃 If a comment explains what it does I think something like `addSubkey` is clear enough :)

twiss:
Superseded by #963.

jpbochi:
sorry I didn't reply to this one earlier. I do like the `generateSubkey` and `addSubkey` pair of methods suggested at https://github.com/openpgpjs/openpgpjs/pull/732. They certainly can be implemented in separate PRs. I'll try to keep track of this one you just started. cheers

PR #931, opened 2019-07-11
Creating light openpg version without indutny/elliptic

chesnokovilya:
Full support for x25519 cryptography. Limited support for NIST cruves and brainpool, since web crypto and node crypto does not support sreaming.


test/crypto/elliptic.js
```diff
@@ -151,63 +151,49 @@ describe('Elliptic Curve Cryptography', function () {
...
-        const curve = new elliptic_curves.Curve(name);
-        expect(curve).to.exist;
-        const keyPair = curve.keyFromPrivate(pair.priv);
-        expect(keyPair).to.exist;
-        const pub = keyPair.getPublic();
-        expect(pub).to.exist;
-        expect(openpgp.util.Uint8Array_to_hex(pub)).to.equal(openpgp.util.Uint8Array_to_hex(pair.pub));
+    it('Creating KeyPair full build', function () {
+      if (!openpgp.util.getFullBuild() && !openpgp.util.getNodeCrypto()) {
+        this.skip;
```
twiss:
```suggestion
        this.skip();
```
🙃

test/crypto/elliptic.js
```diff
@@ -151,63 +151,49 @@ describe('Elliptic Curve Cryptography', function () {
       }
       done();
     });
-    it('Creating KeyPair', function () {
-      const names = ['p256', 'p384', 'p521', 'secp256k1', 'curve25519', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1'];
+    it('Creating KeyPair light build', function () {
+      const names = ['p256', 'p384', 'p521', 'curve25519'];
```
twiss: Could you either skip this test when the next one is not skipped, or (probably cleaner) merge them into one test but just select a different `names` array depending on the build?

src/crypto/public_key/elliptic/indutnyKey.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import enums from '../../../enums';
+
+/**
+ * @constructor
+ */
+function KeyPair(curve, options) {
+  if (!curve.indutnyCurve) {
+    throw(new Error('This curve is supported only in the full build of OpenPGP.js'));
```
twiss: Bit nitpicky, but:
```suggestion
    throw new Error('This curve is supported only in the full build of OpenPGP.js');
```
(Some other places too)

src/crypto/public_key/elliptic/curves.js
```diff
@@ -152,66 +154,83 @@ function Curve(oid_or_name, params) {
...
+      const privateKey = await random.getRandomBytes(32);
+      const keyPair = nacl.sign.keyPair.fromSeed(privateKey);
+      const publicKey = util.concatUint8Array([new Uint8Array([0x40]), keyPair.publicKey]);
+      return {
+        getPublic: () => publicKey,
+        getPrivate: () => privateKey
+      };
     }
+    default:
+      break;
```
twiss: This default case is superfluous

src/crypto/public_key/elliptic/ecdh.js
```diff
@@ -87,28 +90,30 @@ async function kdf(hash_algo, X, length, param, stripLeading=false, stripTrailin
...
           return await webPublicEphemeralKey(curve, Q);
         } catch (err) {
           util.print_debug_error(err);
         }
       }
+      break;
+    }
+    case 'node': {
+      return nodePublicEphemeralKey(curve, Q);
     }
```
twiss: Very nitpicky, but: the curly brackets `{}` are only necessary when you have `let/const` declarations in them; in the web and node case they're unnecessary. (Some other places too - I know they were there already, it was because of the `const result = await ...` that used to be there.)

src/crypto/public_key/elliptic/curves.js
```diff
@@ -251,21 +270,78 @@ async function webGenKeyPair(name) {
...
-    pub: ecdh.getPublicKey().toJSON().data,
-    priv: ecdh.getPrivateKey().toJSON().data
+    publicKey: new Uint8Array(ecdh.getPublicKey()),
+    privateKey: new Uint8Array(ecdh.getPrivateKey())
+  };
+}
+
+const indutnyCurve = util.getFullBuild() ? function (name) {
+  return new indutnyEc.ec(name);
+} : undefined;
```
twiss: Given that you already wrap the calling location of this function in `util.getFullBuild()` and how short it is, I think you could inline it.

src/crypto/public_key/elliptic/curves.js
```diff
@@ -152,66 +154,83 @@ function Curve(oid_or_name, params) {
...
+    this.type = 'node';
+  } else if (this.name === 'curve25519') {
+    this.type = 'curve25519';
+  } else if (this.name === 'ed25519') {
+    this.type = 'ed25519';
   }
-  return keyPair;
-};
+  if (util.getFullBuild()) {
+    this.indutnyCurve = indutnyCurve(this.name);
```
twiss: We don't have to do it now, but in the future, we could optimize this by only creating the curve object when we're actually gonna use it, as this is quite slow. 🙂

test/crypto/elliptic.js
chesnokovilya: We could separate test cases for light-build and full-build to different files in the future. Then different tests would be easier to understand.

twiss: I'm not entirely sure that would help; by having an explicit `if` statement inside the test, it becomes very obvious what the difference between the lightweight and the full build is expected to be. Whereas if you have a similar test in two different files, you need to manually compare them to tell the difference.

src/crypto/public_key/elliptic/curves.js
chesnokovilya: Yeah, I would do it in this PR.

twiss:
Superseded by #956.

PR #933, opened 2019-07-15
bump asn1.js and node-fetch

knoxcard:


twiss:
Hey :wave: Thanks for the PR. Keeping up to date is always good, but is there something specific that these versions add/improve?

knoxcard:
Not that I know of.
Figure if it passes all the travis tests, why not?
Up to date code can include security/optimization fixes as well.

Issue #854, opened 2019-02-05
Intermittent sign+verify failure


twiss:
Two test failures that each happened once:

```
Brainpool Cryptography
  Encrypt and sign message:
      AssertionError: expected false to be true      
      at Context.<anonymous> (test/general/brainpool.js:213:45)
```

~and Unit Tests Crypto API functional testing Sign and verify DSA~ (fixed by #938)

chesnokovilya:
I will try to investigate this, since it prevents travis test builds from successfully passing sometimes.

twiss:
I think the brainpool one might have something to do with https://github.com/openpgpjs/openpgpjs/blob/6d626ea70ca080af410e1dfb1c4a9f0dda99bec3/src/crypto/public_key/elliptic/key.js#L225 stripping leading zeroes or some such.

chesnokovilya:
example non-working signed + encrypted data for brainpool is here https://gist.github.com/chesnokovilya/accb5138bd077616067eff10ea35c1f6

PR #938, opened 2019-07-29
fix issue 854 partially

chesnokovilya:
Solve https://github.com/openpgpjs/openpgpjs/issues/854
DSA signature elements are made constant length.

### DIFF
--- a/src/crypto/public_key/dsa.js
+++ b/src/crypto/public_key/dsa.js
@@ -74,7 +74,6 @@ export default {
       // See Appendix B here: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf
       k = await random.getRandomBN(one, q); // returns in [1, q-1]
       r = gred.redPow(k).fromRed().toRed(redq); // (g**k mod p) mod q
-
       if (zero.cmp(r) === 0) {
         continue;
       }
--- a/src/crypto/public_key/dsa.js
+++ b/src/crypto/public_key/dsa.js
@@ -62,9 +62,7 @@ export default {
     // of leftmost bits equal to the number of bits of q.  This (possibly
     // truncated) hash function result is treated as a number and used
     // directly in the DSA signature algorithm.
-    const h = new BN(
-      util.getLeftNBits(hashed, q.bitLength()))
-      .toRed(redq);
+    const h = new BN(hashed.subarray(0, q.byteLength())).toRed(redq);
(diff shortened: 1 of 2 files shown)


chesnokovilya:
brainpool bug is inside undutny/elliptic. Library makes incorrect verify calculations for some subset of signatures.

chesnokovilya:
@twiss I think we could merge as it is. At least it fixes dsa test failure.

src/crypto/public_key/dsa.js
```diff
@@ -85,8 +85,8 @@ export default {
       break;
     }
     return {
-      r: r.toArrayLike(Uint8Array),
-      s: s.toArrayLike(Uint8Array)
+      r: r.toArrayLike(Uint8Array, 'be', 32),
+      s: s.toArrayLike(Uint8Array, 'be', 32)
```
twiss: Thanks for the PR! :+1: I think these should be `q.bitLength() / 32` instead of 32 (the signature size varies by key size).

twiss: P.S. A bit unrelated but the two calls to `util.getLeftNBits(hashed, q.bitLength()))` could be replaced by `hashed.subarray(0, q.bitLength() / 8)`; q's bit length must be a multiple of 8. Then `getLeftNBits` can be removed. You don't have to do that in this PR of course, only if you feel like it.

tomholub: > Thanks for the PR! :+1: I think these should be `q.bitLength() / 32` instead of 32 (the signature size varies by key size).

Is there a way to produce failure with the currently implemented code `r.toArrayLike(Uint8Array, 'be', 32)` ? If yes, it would be wise to add a test specifically for that (with appropriate key size), as this kind of thing would be easy to miss in the future.

chesnokovilya:
I will change to `q.byteLength` then. It should be equal to `bitLength / 8`  under condition that length is multiple of 8.

src/crypto/public_key/dsa.js
twiss: Hmm. I think it would actually still work with this code, as 32 bytes is enough to fit every signature size. Nevertheless, I think we don't have any tests for that. I agree it would be nice to have one.

tomholub: Maybe not as crucial to have a test for this if it would not cause a hard failure (and is not a security concern). Thanks for the explanation :+1:

PR #941, opened 2019-08-12
fix expiration date in flowcrypt key

chesnokovilya:


tomholub:
Ah, sorry about that. A time bomb in your test suite!

twiss:
No problem ^.^ We've had the exact same thing before with other test keys 🙃

PR #947, opened 2019-08-15
use native node RSA crypto for performance

### LINKED ISSUE #530: Node performance is unacceptably slow.
When I try to generate a keypair in nodejs (7.2.1). The process takes well over 30 seconds and this is for a key of 4096 bits. GPG2 took < 2. This was done on an 2015 MacBook Pro 15", I am wondering if there is anything else I can enable to improve it. 

In my current config, I do have `aead_protect: true` and its using native nodejs.

tomholub:
fixes #530

### DIFF
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -228,8 +228,8 @@ export default {
       return key;
     } else if (nodeCrypto && nodeCrypto.generateKeyPair) {
       const opts = {
-        modulusLength: B.toNumber(),
-        publicExponent: E.toNumber(),
+        modulusLength: Number(B.toString(10)),
+        publicExponent: Number(E.toString(10)),
         publicKeyEncoding: { type: 'pkcs1', format: 'der' },
         privateKeyEncoding: { type: 'pkcs1', format: 'der' }
       };
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -172,6 +172,7 @@ export default {
     let key;
     E = new BN(E, 16);
     const webCrypto = util.getWebCryptoAll();
+    const nodeCrypto = util.getNodeCrypto();
 
     // Native RSA keygen using Web Crypto
     if (webCrypto) {
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -228,8 +228,8 @@ export default {
       return key;
     } else if (nodeCrypto && nodeCrypto.generateKeyPair) {
       const opts = {
-        modulusLength: Number(B.toString(10)),
-        publicExponent: Number(E.toString(10)),
+        modulusLength: B.toNumber(),
+        publicExponent: E.toNumber(),
         publicKeyEncoding: { type: 'pkcs1', format: 'der' },
         privateKeyEncoding: { type: 'pkcs1', format: 'der' }
       };
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -31,9 +31,6 @@ import random from '../random';
 import config from '../../config';
 import util from '../../util';
 
-const nodeCrypto = util.getNodeCrypto(); // for native node keygen
-const asn1 = nodeCrypto ? require('asn1.js') : undefined;
-
 // Helper for IE11 KeyOperation objects
 function promisifyIE11Op(keyObj, err) {
   if (typeof keyObj.then !== 'function') { // IE11 KeyOperation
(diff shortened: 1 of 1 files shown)


tomholub:
It failed on Node 10 only, for brainpool crypto. Maybe a flake? (unsure)
```
  1) Unit Tests
       General
         Brainpool Cryptography
           Encrypt and sign message:
      AssertionError: expected false to be true
```

src/crypto/public_key/rsa.js
```diff
@@ -31,6 +31,9 @@ import random from '../random';
 import config from '../../config';
 import util from '../../util';
 
+const nodeCrypto = util.getNodeCrypto(); // for native node keygen
```
twiss: The reason `getWebCryptoAll` is called inside the function, is so that you can change its behavior using `config.use_native`. Could you do the same thing with `getNodeCrypto`?

tomholub: Sure. That means I'd run `asn1.define('RSAPrivateKey'` on top regardless of `config.use_native` setting, then do `getNodeCrypto` inside `generate` so that it's sensitive to `config.use_native` which user may have set sometime after the whole lib was initiated. Did I understand correctly?

This is not a nitpick, I think it's very much expected that keygen stays non-native when `config.use_native` is false.

tomholub: Sure. That means I'd run `asn1.define('RSAPrivateKey'` on top regardless of `config.use_native` setting, then do `getNodeCrypto` inside `generate` so that it's sensitive to `config.use_native` which user may have set sometime after the whole lib was initiated. Did I understand correctly?

This is not a nitpick, I think it's very much expected that keygen stays non-native when `config.use_native` is false.

twiss: Oh, right. You could do both actually (call the function in both places, here to check whether we're on node and inside the function to check `config.use_native`).

tomholub: Yup, that's what I was thinking. Will do.

tomholub:
One day it would be nice to have a test that checks that on node, with `use_native`, keygen takes < 1/3 the time of without `use_native`, confirming that in one case it's using native crypto and in one case it's not.

One day..

twiss:
Hmhm :+1: Alternatively, it would be nice to have some benchmarks for both (and a bunch of other stuff), with alerts if the performance of either of them changed significantly.

tomholub:
As a datapoint, on mobile, generating a 4096 bit key in js takes about 90 seconds :roll_eyes: (presumably ~ 2x 45 seconds)

tomholub:
Whoops - I stage my edits line by line, and I left one out.

This is because vscode wanted to autoformat the doc, so I'd be checking in unrelated white space changes. It would be good to know what is the intended formatting and then include vscode settings in the repo, so that formatting can work in our favor for people using vscode also. That would be for another day, too.

twiss:
Hmhm :+1: The consensus spacing in this repository seems to be:
```js
const test = function(a, b=1) { // "function (" also appears quite often
  if (a + 1 === b) {
    try {
      someFn(b);
    } catch(e) {}
  }
  const arr = [a];
  const obj = { b };
};
```
Taken together it's all a bit inconsistent and ad-hoc, so calling this formatting "intended" is perhaps undeserved praise, but a PR codifying something close to this (and perhaps fixing code that doesn't conform) would nevertheless be welcome :)

tomholub:
Here's what happens when I paste that into vscode (default settings):

```js
const test = function (a, b = 1) { // "function (" also appears quite often
  if (a + 1 === b) {
    try {
      someFn(b);
    } catch (e) { }
  }
  const arr = [a];
  const obj = { b };
};
```
That's really close, maybe except for spaces around `=` in args. But if the above is cool with you, the advantage is that it's the default of a very popular JS IDE. I can contribute the auto-format whitespace-only-PR to bootstrap some consistency. Thoughts?

twiss:
Hmhm :+1: The only downside / argument for keeping it the way it is is to keep `git blame` clean. But that spacing does make a bit more sense so I think it's OK. Perhaps wait a bit until #931 is merged to make a PR, though, since that PR is rather large?

Also, if we're talking about ideals, it would be nice to make `npm lint` enforce this spacing at some point :)

tomholub:
There's git blame history, but I know what you mean. I'm not sure there's a way to avoid that - one way or another, if there is to be consistency, there will be the same level of git blame pollution whether it's done all at once or in incremental steps over time.

chesnokovilya:
yeah I think we should switch to `const test = function (a, b = 1) ` since this is airbnb style. There is eslint rule for it `space-infix-ops=error` now it is `off`.

tomholub:
Oh, and thanks for merging! :tada:

PR #948, opened 2019-08-16
Fixes brainpool sign/verify bug. Closing issue #854

chesnokovilya:
Update elliptic to the version with correct signature creation and verification in the case of leading zero in hash.
Closing issue https://github.com/openpgpjs/openpgpjs/issues/854


larabr:
For future ref, the fixes are in https://github.com/openpgpjs/elliptic/pull/2 and https://github.com/openpgpjs/elliptic/pull/3 .

PR #956, opened 2019-08-30
loading indutny/elliptic library to browser on demand

chesnokovilya:


src/openpgp.js
```diff
@@ -93,6 +94,41 @@ export function destroyWorker() {
...
+//////////////////////////
+
+
+let elliptic;  // instance of the indutny/elliptic
+
+/**
+ * Load elliptic by path or from node_modules
+ * @param {String} path relative path to elliptic browserified package
+ */
+export async function loadElliptic(path) {
```
twiss: I would put this function somewhere else, maybe indutnyKey.js?

src/crypto/public_key/elliptic/curves.js
```diff
@@ -152,75 +154,76 @@ function Curve(oid_or_name, params) {
...
+    this.type = 'curve25519';
+  } else if (this.name === 'ed25519') {
+    this.type = 'ed25519';
   }
-  return keyPair;
-};
+  this.getIndutnyCurve = util.getUseElliptic() ? async name => {
+    const elliptic = getElliptic() || await loadElliptic(build.external_indutny_elliptic_path);
+    return new elliptic.ec(name);
+  } : undefined;
```
twiss: You could also put this function in indutnyKey.js - it doesn't use `this`, so it doesn't have to be a method.

src/crypto/public_key/elliptic/indutnyKey.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+/**
+ * @constructor
+ */
+function KeyPair(indutnyCurve, options) {
+  this.keyPair = indutnyCurve.keyPair(options);
+  if (this.keyPair.validate().result !== true) {
+    throw new Error('Invalid elliptic public key');
+  }
```
twiss: I'm not sure that we should always call `validate` here - before, we were only calling it in `keyFromPublic`, now we're also calling it in what used to be `keyFromPrivate`. Could you maybe briefly test how long `validate` takes to see whether it makes any difference at all?

test/crypto/elliptic.js
```diff
@@ -5,7 +5,13 @@ chai.use(require('chai-as-promised'));
 
 const expect = chai.expect;
 
-describe('Elliptic Curve Cryptography', function () {
+describe('Elliptic Curve Cryptography @lightweight', function () {
+  if (typeof window !== 'undefined') {
+    before(async function() {
+      await openpgp.loadElliptic('../dist/elliptic.min.js');
+    })
```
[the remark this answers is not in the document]
chesnokovilya: It is not so simple. This path is only valid from `test` folder. So it could not be universal.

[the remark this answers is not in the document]
twiss: Yes, I meant that when you have a config option (that defaults to `'./elliptic.min.js'`), you could then do
```js
openpgp.config.elliptic_path = '../dist/elliptic.min.js';
```
here.

src/openpgp.js
```diff
@@ -93,6 +94,41 @@ export function destroyWorker() {
...
+  if(typeof window !== 'undefined' && build.external_indutny_elliptic) {
+    // Fetch again if it fails, mainly to solve chrome bug "body stream has been lost and cannot be disturbed"
+    const ellipticPromise = util.dl({ filepath: path }).catch(() => util.dl({ filepath: path }));
```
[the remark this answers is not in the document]
chesnokovilya: So you want to eliminate `path` option here? It would not work so simple as you could not load elliptic by static path stored in config. For example, you could look at `initWorker` function.

src/util.js
```diff
@@ -753,5 +758,18 @@ export default {
...
+  },
+
+  /**
+   * download script from filepath
+   * @param {Object} params download parameters
+   */
+  dl: async function({ filepath, integrity }) {
+    const options = {
+      integrity,
+      credentials: 'include'
```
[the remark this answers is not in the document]
chesnokovilya: Not clear for me what I should remove =).

src/openpgp.js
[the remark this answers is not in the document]
twiss: I'm proposing that you call
```js
dl(config.elliptic_path, config.elliptic_fetch_options);
```
here. I know that's not what `initWorker` does, but `initWorker` is a function that's meant to always be called explicitly, whereas `loadElliptic` is meant to be called automatically by OpenPGP.js itself, so a config option makes sense.

src/build.env.js
```diff
@@ -0,0 +1,22 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+export default {
+  use_indutny_elliptic: true,
```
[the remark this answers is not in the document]
chesnokovilya: This option prevents openpgp from tryng to load elliptic. I think this is useful. Fro example `browserify` has `exclude` and `external` options. At runtime we could check if `elliptic` should be loaded but not loaded.

src/util.js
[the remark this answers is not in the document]
twiss: The entire options object :) and make it a parameter instead. So
```js
dl: async function(path, options) {
```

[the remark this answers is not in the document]
twiss: The entire options object :) and make it a parameter instead. So
```js
dl: async function(path, options) {
```

src/build.env.js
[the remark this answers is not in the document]
twiss: The example code I posted would also prevent OpenPGP.js from trying to load elliptic, right? It is a bit more work than passing a flag, but the advantage is that you don't have to create a special build of OpenPGP.js, and that it simplifies OpenPGP.js because we can always just call `loadElliptic` before checking first.

src/crypto/public_key/elliptic/curves.js
chesnokovilya: IndutnyKey.js is not public now, I could make it public and move this function here.

src/util.js
```diff
@@ -753,5 +758,18 @@ export default {
       result += ALPHABET[MASK & (buffer >> bitsLeft)];
     }
     return result;
+  },
+
+  /**
+   * download script from filepath
+   * @param {Object} params download parameters
+   */
+  dl: async function({ filepath, integrity }) {
```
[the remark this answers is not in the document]
chesnokovilya: One use case case could be expanding on-demand web loading to hash functions, for example deprecated md5.

src/openpgp.js
chesnokovilya: I could move it here, but should we move `loadScript` from openpgp? We could use it to load other rarely used code, like md5 hash.

src/build.env.js
```diff
@@ -0,0 +1,22 @@
...
+  external_indutny_elliptic: false,
+  external_indutny_elliptic_path: 'elliptic.min.js'
```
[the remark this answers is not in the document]
chesnokovilya: removed `build.env.js` entirely

test/crypto/elliptic.js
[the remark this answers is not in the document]
chesnokovilya: ok, changed to your suggestion.

src/util.js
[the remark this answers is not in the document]
chesnokovilya: moved to src/lightweight_helper.js

src/openpgp.js
chesnokovilya: moved to lightweight_helper.js

src/build.env.js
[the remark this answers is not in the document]
chesnokovilya: Do you want to substitute function content using grunt replace command? It is not clear what do you want =).

src/crypto/public_key/elliptic/indutnyKey.js
chesnokovilya: I wanted to explicitly cache `indutnyCurve` here. I know that nodeJs is caching require, however I still think it is useful to have explicit cache, for various browser environments.

src/openpgp.js
```diff
@@ -714,3 +750,40 @@ function onError(message, error) {
...
+  const script = document.createElement('script');
+
+  script.src = path;
+  if (integrity) {
+    script.integrity = integrity;
+  }
+  script.onload = e => cb(e);
+  script.onerror = e => cb(undefined, e);
+
+  document.head.appendChild(script);
```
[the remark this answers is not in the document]
chesnokovilya: I have removed loadScriptHelper.

src/config/config.js
```diff
@@ -190,5 +190,24 @@ export default {
...
+   * Path to load elliptic from external file
+   */
+  external_indutny_elliptic_path: 'elliptic.min.js',
+
+  /*
+   * indutny elliptic fetch options
+   */
+  indutny_elliptic_fetch_options: {
+    credentials: 'include'
+  }
```
twiss: This default `credentials: 'include'` is not necessary, it can be set by the application if needed. You can just default to `{}` for the options, I think.

package.json
```diff
@@ -64,22 +70,16 @@
...
-    "elliptic": "github:openpgpjs/elliptic#6b7801573b8940a49e7b8253176ece2725841efd",
-    "hash.js": "^1.1.3",
     "pako": "^1.0.6",
     "seek-bzip": "github:openpgpjs/seek-bzip#6187fc025851d35c4e104a25ea15a10b9b8d6f7d",
+    "sinon": "^4.3.0",
+    "text-encoding-utf-8": "^1.0.2",
     "tweetnacl": "github:openpgpjs/tweetnacl-js#1ef755f2b252a3e328ac739848d00e0dad76be2d",
     "web-stream-tools": "github:openpgpjs/web-stream-tools#dc4b05e8a272b45819233f3df735423432beacfc",
-    "email-addresses": "github:openpgpjs/email-addresses#686743c6452b44bafcd06d47db7f36ddf3f3f118"
+    "whatwg-fetch": "^2.0.3"
```
twiss: This is pretty nitpicky, but could you keep these packages in the order they were in? I know they're usually in alphabetic order, but the order they were in was to roughly separate what was needed only for development (everything up to `sinon`) and what packages end up in a build (everything starting from `text-encoding-utf-8`).
Also, even if you disagree with that, keeping them in the same order in this PR would make it easier to see what changed here.

src/lightweight_helper.js
```diff
@@ -0,0 +1,31 @@
...
+  // eslint-disable-next-line
+    if(self.importScripts) {
+    return importScripts(path);
+  }
+  return new Promise((resolve, reject) => {
+    const script = document.createElement('script');
+    script.src = path;
+    if (integrity) {
+      script.integrity = integrity;
+    }
```
twiss: You're never passing `integrity`, so I think this can be removed

src/lightweight_helper.js
```diff
@@ -0,0 +1,31 @@
...
+    script.onload = e => resolve(e);
```
twiss:
```suggestion
    script.onload = () => resolve();
```

src/lightweight_helper.js
```diff
@@ -0,0 +1,31 @@
...
+    script.onerror = e => reject(e);
```
twiss:
```suggestion
    script.onerror = () => reject(new Error('Failed to load ' + path));
```

src/lightweight_helper.js
```diff
@@ -0,0 +1,31 @@
+
+/**
+ * @param {String} path
+ * @param {String} integrity
+ */
+export const loadScript = (path, integrity) => {
+  // eslint-disable-next-line
+    if(self.importScripts) {
```
twiss:
```suggestion
  if (typeof importScripts !== 'undefined') {
```
Maybe then you can also remove the `eslint-disable-next-line`?

src/crypto/public_key/elliptic/indutnyKey.js
```diff
@@ -0,0 +1,84 @@
...
+    throw new Error('elliptic library has not loaded correctly');
+  }
+  URL.revokeObjectURL(mainUrl);
+  return window.openpgp.elliptic;
+}
+
+function loadElliptic() {
+  if(util.detectNode() && config.external_indutny_elliptic) {
+    // eslint-disable-next-line
+    return require('./' + config.external_indutny_elliptic_path);
```
twiss: I would move this `'./'` to the default value of the config value, as it should always work with it.
```suggestion
    return require(config.external_indutny_elliptic_path);
```

src/config/config.js
```diff
@@ -190,5 +190,24 @@ export default {
...
+   */
+  external_indutny_elliptic: false,
+  /*
+   * use indutny/elliptic library
+   */
+  use_indutny_elliptic: true,
+  /**
+   * Path to load elliptic from external file
+   */
+  external_indutny_elliptic_path: 'elliptic.min.js',
```
twiss: Pretty nitpicky but I would drop the "external", since it's pretty obvious that a path refers to an external file.
```suggestion
  indutny_elliptic_path: './elliptic.min.js',
```

Gruntfile.js
```diff
@@ -131,6 +128,45 @@ module.exports = function(grunt) {
...
+          'dist/openpgp.js'
+        ],
+        overwrite: true,
+        replacements: [
+          {
+            from: "use_indutny_elliptic: true",
+            to: "use_indutny_elliptic: false"
+          }
+        ]
+      },
```
twiss: I don't think you have to create a grunt command and option for this. If they want to disable elliptic, they can do themselves:
```js
openpgp.config.use_indutny_elliptic = false;
```

src/crypto/public_key/elliptic/curves.js
```diff
@@ -152,75 +153,72 @@ function Curve(oid_or_name, params) {
...
+      const privateKey = await random.getRandomBytes(32);
+      const keyPair = nacl.sign.keyPair.fromSeed(privateKey);
+      const publicKey = util.concatUint8Array([new Uint8Array([0x40]), keyPair.publicKey]);
+      return { publicKey, privateKey };
     }
   }
-  return new KeyPair(this, keyPair);
+  if (!util.getUseElliptic()) {
+    throw new Error('This curve is only supported in the full build of OpenPGP.js');
+  }
```
twiss: I think you could move this check into the `getIndutnyCurve` function that you're calling below, and then also remove it in the other places you're checking this (since they will eventually all call `getIndutnyCurve`).

src/util.js
```diff
@@ -638,6 +638,10 @@ export default {
     return navigator.hardwareConcurrency || 1;
   },
 
+  getUseElliptic: function() {
+    return config.use_indutny_elliptic;
+  },
+
```
twiss: I'm not sure this function is really necessary - especially once you're only checking it in one place, you can just check
```js
if (!config.use_indutny_elliptic) {
```
directly.

Then, in the test suite where you're checking this in a bunch of places, I'm not sure it's really necessary? We're never calling the test suite with `config.use_indutny_elliptic = false`, so it should always be available in the test suite.

src/config/config.js
```diff
@@ -190,5 +190,24 @@ export default {
...
-  known_notations: ["preferred-email-encoding@pgp.com", "pka-address@gnupg.org"]
+  known_notations: ["preferred-email-encoding@pgp.com", "pka-address@gnupg.org"],
+  /**
+   * use external indutny/elliptic library
+   */
+  external_indutny_elliptic: false,
+  /*
+   * use indutny/elliptic library
+   */
+  use_indutny_elliptic: true,
```
twiss: Even more nitpicky but I would swap these two options around, to make the order more logical (first we check whether to use it, then we check where to find it).

src/config/config.js
```diff
@@ -190,5 +190,24 @@ export default {
...
+  /**
+   * Path to load elliptic from external file
+   */
+  external_indutny_elliptic_path: 'elliptic.min.js',
+
```
twiss:
```suggestion
```

src/lightweight_helper.js
```diff
@@ -0,0 +1,31 @@
+
```
twiss:
```suggestion
```

src/crypto/public_key/elliptic/indutnyKey.js
```diff
@@ -0,0 +1,84 @@
...
+
+export async function getIndutnyCurve(name) {
+  let elliptic;
+  if(typeof window !== 'undefined' && config.external_indutny_elliptic) {
+    if (!ellipticPromise) {
+      ellipticPromise = loadEllipticPromise().catch(e => {
+        ellipticPromise = undefined;
+        throw e;
+      });
+    }
```
twiss: I know I suggested to put this here so it's my bad, but I would move this bit of code to `loadElliptic`, given the name of that function. And then to make the logic a bit more clear I would do something like
```js
if (!config.external_indutny_elliptic) {
  return require('elliptic');
}
if (util.detectNode()) {
  // eslint-disable-next-line
  return require(config.external_indutny_elliptic_path);
}
// (this code)
```
there.

And then this function can just be
```js
  const elliptic = await loadElliptic();
  return new elliptic.ec(name);
```

src/lightweight_helper.js
chesnokovilya: I have researched a bit error that are thrown here, they are mostly network related.  This promise would be resolved even if path is wrong, and actually elliptic has failed to load.

chesnokovilya: Example here https://codesandbox.io/s/intelligent-darwin-w9cxx

src/config/config.js
chesnokovilya: fixed

src/config/config.js
chesnokovilya: fixed

src/lightweight_helper.js
chesnokovilya: This is kinda general function, so in the future we could use this attribute, even if it is not used now.

package.json
chesnokovilya: Npm kinda do it by itself, once I run `npm install elliptic`. May be we could switch to the new ordering convention? It is bothersome to rearrange lines manually after each change, or switching to another git branch.

src/lightweight_helper.js
twiss: Ok. But `e` is not an error object, it is an event object. You should throw an error instead of an event. Even for network errors, "failed to load" is pretty descriptive.

src/config/config.js
chesnokovilya: fixed

src/lightweight_helper.js
chesnokovilya: It is actually error event is passed here, I would better catch it in loadEllipticPromise.

src/lightweight_helper.js
chesnokovilya: fixed

src/crypto/public_key/elliptic/indutnyKey.js
chesnokovilya: fixed

src/lightweight_helper.js
chesnokovilya: fixed

src/config/config.js
chesnokovilya: hmm, not obvious for me. There could be internal path, leading to node_modules for example.

Gruntfile.js
chesnokovilya: Since we have version without elliptic, let's have an option for an automatic build of this version, and then testing it.

src/util.js
chesnokovilya: I have this function, so I could avoid importing config in bunch of places. I think it is valid use-case. 
Regarding tests, since we have option to disable elliptic, let's have test suite which would work with this option.

src/crypto/public_key/elliptic/indutnyKey.js
chesnokovilya: fixed

src/crypto/public_key/elliptic/curves.js
chesnokovilya: If we move this check to the `getIndutnyCurve`, then sometimes in ecdh and ecdsa we would get rejected promise, and sometime we would get thrown error. Than we would need to catch those errors. It is considered anti-pattern.

```A function is hard to use when it sometimes returns a promise and sometimes throws an exception. When writing an async function, prefer to signal failure by returning a rejected promise```

https://stackoverflow.com/questions/33445415/javascript-promises-reject-vs-throw

File: src/lightweight_helper.js (hunk 15 above)
Comment (twiss) [reply src/lightweight_helper.js]: It is an error event, but not an error object. You could do
```js
    script.onerror = e => reject(new Error(e.message));
```
if you prefer, but you should always throw an error object.

src/config/config.js
twiss: That would still be external (to the dist file) right? I.e. `external_indutny_elliptic` would be true. Also, I'm not entirely sure that loading indutny externally from node_modules using a path on Node.js is very important, as file size is not a concern there and `require('elliptic')` should just work in any case.

Gruntfile.js
twiss: If you want to test it, you can set the config option in the test suite, no? We shouldn't have a separate build for every single config option, it would run out of hand.

src/util.js
twiss: > I have this function, so I could avoid importing config in bunch of places.

That's why I mentioned "especially once you're only checking it in one place", because then you only have to import config in one place. I don't think that importing `util` instead of importing `config` is better, when it is `config` that you really need to check.

The test suite already uses `config` in lots of places, you could search for `config` in the test directory to check.

src/crypto/public_key/elliptic/curves.js
twiss: All of the functions are `async`, right? So it would always cause a rejected promise no matter where you put the `throw`.

src/util.js
chesnokovilya: removed

package.json
twiss: > May be we could switch to the new ordering convention?

Yeah, we could do that, but preferably not in this PR, since it makes it harder to see what changed.
Once we switch to ES6 modules, we'll move the `devDependencies` starting from `text-encoding-utf-8` to `dependencies`, so then is probably a good time to fix the ordering.

PR #963, opened 2019-09-10
Feature: add new sub key to the key object

chesnokovilya:
Implements addSubkey function.
Inspired by https://github.com/openpgpjs/openpgpjs/pull/732

### DIFF
--- a/src/key.js
+++ b/src/key.js
@@ -826,6 +826,37 @@ Key.prototype.verifyAllUsers = async function(keys) {
   return results;
 };
 
+Key.prototype.addSubkey = async function(options = {}) {
+  if (!this.isPrivate()) {
+    throw new Error("Cannot add a subkey to a public key");
+  }
+  const defaultOptions = {};
+  const defaultAlgo = this.primaryKey.algorithm;
+  if (defaultAlgo.indexOf('rsa') === 0) {
+    defaultOptions.numBits = this.primaryKey.getAlgorithmInfo().bits;
+  } else if (defaultAlgo === 'ecdh' || defaultAlgo === 'ecdsa' || defaultAlgo === 'eddsa') {
+    defaultOptions.curve = this.primaryKey.getAlgorithmInfo().curve;
+  }
+  const secretKeyPacket = this.primaryKey;
+  if (!secretKeyPacket.isDecrypted()) {
+    throw new Error("Key is not decrypted");
+  }
+  options = sanitizeKeyOptions(options, defaultOptions);
+  const keyPacket = await generateSecretSubkey(options);
+
+  if (options.passphrase) {
+    await keyPacket.encrypt(options.passphrase);
+  }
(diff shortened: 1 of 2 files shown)


src/key.js
```diff
@@ -826,6 +826,30 @@ Key.prototype.verifyAllUsers = async function(keys) {
...
+  if (!this.isPrivate()) {
+    throw new Error("Cannot add a subkey to a public key");
+  }
+  const defaultOptions = {};
+  const defaultAlgo = this.primaryKey.algorithm;
+  if (defaultAlgo.indexOf('rsa') === 0) {
+    defaultOptions.numBits = this.primaryKey.getAlgorithmInfo().bits;
+  } else if (defaultAlgo === 'ecdh' || defaultAlgo === 'ecdsa' || defaultAlgo === 'eddsa') {
+    defaultOptions.curve = this.primaryKey.getAlgorithmInfo().curve;
+  }
```
twiss: It seems like all of this could be simplified to
```js
const defaultOptions = this.primaryKey.getAlgorithmInfo();
defaultOptions.numBits = defaultOptions.bits;
```
Or, we could even remove the last line if we rename `numBits` to `bits` everywhere to make it consistent within the library, and then do something like
```js
export function generateKey({ userIds = [], passphrase = "", numBits = 2048, bits = numBits, keyExpirationTime = 0, curve = "", date = new Date(), subkeys = [{}] }) {
```
in src/openpgp.js (to keep it backwards compatible). But maybe that's something for another day. Up to you :)

src/key.js
```diff
@@ -1769,3 +1773,27 @@ export async function isAeadSupported(keys, date = new Date(), userIds = []) {
   }));
   return supported;
 }
+
+async function bindSignature(subkey, primaryKey, options) {
```
twiss: Perhaps name this `createBindingSignature`?

Also, bit nitpicky but could you put this function a bit closer to the rest of the related code? E.g. under `wrapKeyObject`.

src/key.js
```diff
@@ -1329,52 +1353,52 @@ export async function generate(options) {
...
-      }
-    } else if (options.numBits) {
-      options.algorithm = enums.publicKey.rsa_encrypt_sign;
-    } else {
-      throw new Error('Unrecognized key type');
-    }
-    return options;
-  }
-
   async function generateSecretKey(options) {
```
twiss: This is purely a cosmetic question, but could you pull this function out as well, and put it above `generateSecretSubkey`, so that they're together and that the diff is smaller (when ignoring whitespace changes)?

src/key.js
chesnokovilya: I want to refactor this module anyway, let it stay here for now, that way file diff is minimised.

src/key.js
chesnokovilya: Updated

PR #970, opened 2019-09-17
Rename `numBits` and `bits` to `rsaBits`

### LINKED ISSUE #969: Rename `numBits` and `bits` to `rsaBits`
It would be nice to be consistent between `generateKey`'s `numBits` and `getAlgorithmInfo`'s `bits`. `numBits` is redundant with its type. `bits` is fine. `rsaBits` would indicate that it only applies to RSA keys, which is nice and makes it so that `openpgp.generateKey({ rsaBits: 4096 })` contains an explicit indication that RSA is being used.

We might want to do this in a backwards compatible way (not just so that we can perhaps do this before v5, but also so that `openpgp.generateKey({ numBits: 4096 })` doesn't silently start generating 2048-bit keys for people who don't read the release notes), e.g.

```js
export function generateKey({ userIds = [], passphrase = "", numBits = 2048, rsaBits = numBits, keyExpirationTime = 0, curve = "", date = new Date(), subkeys = [{}] }) {
```

(And `Key.prototype.addSubkey`, depending on how long we wait with this :) And maybe `key.generate` as well? E.g. @tomholub do you use the low-level API for this?)

twiss:
Fix #969.

Keep supporting the old names as well though in `openpgp.generateKey` and `getAlgorithmInfo`, but not in `openpgp.key.generate` (as it is recommended that developers use `openpgp.generateKey` instead, and it now throws when using `numBits` instead of `rsaBits`, so there's no risk of silent key security downgrade).

The old names can now be considered deprecated, and might be removed in v5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ RSA keys:
 ```js
 var options = {
     userIds: [{ name:'Jon Smith', email:'jon@example.com' }], // multiple user IDs
-    numBits: 4096,                                            // RSA key size
+    rsaBits: 4096,                                            // RSA key size
     passphrase: 'super long and hard to guess secret'         // protects the private key
 };
 ```
--- a/src/key.js
+++ b/src/key.js
@@ -829,7 +829,7 @@ Key.prototype.verifyAllUsers = async function(keys) {
 /**
  * Generates a new OpenPGP subkey, and returns a clone of the Key object with the new subkey added.
  * Supports RSA and ECC keys. Defaults to the algorithm and bit size/curve of the primary key.
- * @param {Integer} options.numBits    number of bits for the key creation.
+ * @param {Integer} options.rsaBits    number of bits for the key creation.
  * @param {Number} [options.keyExpirationTime=0]
  *                             The number of seconds after the key creation time that the key expires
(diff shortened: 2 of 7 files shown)


tomholub:
I should fix the TypeScript definitions, too, but that's for another issue

chesnokovilya:
Hmm, maybe we could remove numerous ways of doing same thing. For example, remove public access to `key.generate` and leave only `openpgp.generateKey`.

tomholub:
We use openpgp.generateKey, I was not aware of key.generate

twiss:
> Hmm, maybe we could remove numerous ways of doing same thing. For example, remove public access to `key.generate` and leave only `openpgp.generateKey`.

Yeah. After we remove the Worker, the difference between `openpgp.generateKey` and `key.generate` will be even smaller, so we could remove one of them. OTOH, `openpgp.generateKey` gives you some stuff, like a revocation certificate, that is good to have for new users of the API, but that you might not necessarily need if you know what you're doing.

For purposes of tree shaking, we probably want to keep a function at `key.generate`, so that you can do

```js
import { key } from 'openpgpjs';
key.generate(...);
```

without importing everything else. But this function could of course have an API more like `openpgp.generateKey` than the current one, if we want.

PR #983, opened 2019-10-02
Support non-human-readable notation values

wiktor-k:
This change adds support for binary (non-human-readable) values in signature notations. Signature notations that have the human-readable flag (0x80) will be deserialized into string values while anything else into raw Uint8Array buffers.

Additionally the check for human-readable flag was modified to check the existence of the flag instead of comparison with the whole value.

Some of these changes were suggested back then: https://github.com/openpgpjs/openpgpjs/issues/897#issuecomment-490030917

For the record, the signature with binary notations displayed in an OpenPGP packet parser on [dump.sequoia-pgp.org](https://dump.sequoia-pgp.org/?data=-----BEGIN%20PGP%20SIGNATURE-----%0D%0AVersion%3A%20OpenPGP.js%20v4.6.2%0D%0AComment%3A%20https%3A//openpgpjs.org%0D%0A%0D%0AwncEARYKAB8FAl2TS9MYFAAAAAAADAADdGVzdEBrZXkuY29tAQIDAAoJEGZ9%0D%0AgtV/iL8hrhMBAOQ/UgqRTbx1Z8inGmRdUx1cJU1SR4Pnq/eJNH/CFk5DAP0Q%0D%0AhUhMKMuiM3pRwdIyDOItkUWQmjEEw7/XmhgInkXsCw%3D%3D%0D%0A%3DZGXr%0D%0A-----END%20PGP%20SIGNATURE-----%0D%0A).

### DIFF
--- a/src/packet/signature.js
+++ b/src/packet/signature.js
@@ -71,7 +71,8 @@ function Signature(date = new Date()) {
   this.revocationKeyAlgorithm = null;
   this.revocationKeyFingerprint = null;
   this.issuerKeyId = new type_keyid();
-  this.notations = [];
+  this.rawNotations = [];
+  this.notations = {};
   this.preferredHashAlgorithms = null;
   this.preferredCompressionAlgorithms = null;
   this.keyServerPreferences = null;
@@ -233,13 +234,14 @@ Signature.prototype.write_hashed_sub_packets = function () {
     bytes = util.concat([bytes, this.revocationKeyFingerprint]);
     arr.push(write_sub_packet(sub.revocation_key, bytes));
   }
-  this.notations.forEach(([name, value]) => {
-    bytes = [new Uint8Array([0x80, 0, 0, 0])];
+  this.rawNotations.forEach(([{ name, value, humanReadable }]) => {
+    bytes = [new Uint8Array([humanReadable ? 0x80 : 0, 0, 0, 0])];
     // 2 octets of name length
     bytes.push(util.writeNumber(name.length, 2));
     // 2 octets of value length
     bytes.push(util.writeNumber(value.length, 2));
-    bytes.push(util.str_to_Uint8Array(name + value));
+    bytes.push(util.str_to_Uint8Array(name));
+    bytes.push(value);
     bytes = util.concat(bytes);
     arr.push(write_sub_packet(sub.notation_data, bytes));
   });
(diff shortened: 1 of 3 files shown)


twiss:
Hey Wiktor, thanks for the PR!

I know I more or less suggested this, but I'm not entirely sure whether it's a good idea to return either a string or a Uint8Array in the same place depending on the packet. Especially since binary notations are probably rare, developers might test their code with only text notations, and never realize that our API can return Uint8Arrays, and then break when it does.

There's a few ways I can think of to fix this, none of them perfect. We could have a `notations` and `binaryNotations` property (one with strings, same as now, and one with Uint8Arrays). Or we could have something like `rawNotations = [{name: String, value: Uint8Array, humanReadable: Boolean}, ...]`, and then `notations` as a convenience API (perhaps with only the human-readable notations, which could then just be a `Map` of `name => string`, or even an object like it used to be before - which is kinda similar to the other option I suggested in the comment you linked.) Thoughts?

---

Side note: for names, the spec says that the string must be UTF-8. For human-readable values, it's not specified, but I kind of think that was probably the intention given that everything else in the spec is UTF-8. So, we should be using `util.decode_utf8` and `util.encode_utf8` instead of `util.Uint8Array_to_str` and `util.str_to_Uint8Array`. That doesn't necessarily have to happen in this PR, of course. But maybe the `rawNotations` would be nice to have on the off chance that there exists a human-readable notation out there that is not UTF-8 (perhaps generated by us?)

Side question: do you know of any keys with binary notations in the wild? Can other software generate them, or are these the first? Is it well-supported by other software?

wiktor-k:
> I know I more or less suggested this, but I'm not entirely sure whether it's a good idea to return either a string or a Uint8Array in the same place depending on the packet. Especially since binary notations are probably rare, developers might test their code with only text notations, and never realize that our API can return Uint8Arrays, and then break when it does.

Yep. Actually I had the same thoughts at the back of my head but I figured out it's better to talk with some solid code as a template rather than with pure words :)

> There's a few ways I can think of to fix this, none of them perfect.

I'll take a moment to implement them and feel which one is the best. Usually sooner or later one of solutions looks like a clear winner...

> Side note: for names, the spec says that the string must be UTF-8.

That's a good point, I'll make sure it's included.

> Side question: do you know of any keys with binary notations in the wild? Can other software generate them, or are these the first? Is it well-supported by other software?

No, but I've seen people stuffing base64-encoded data here and there and it kind of annoys me when directly embedding binary data is possible (base64 increases storage with no clear gain). GnuPG displays the value as `[non human readable]` but lower level software (libraries) such as Sequoia works with them just fine. Take a look at raw dump [here](https://dump.sequoia-pgp.org/?data=-----BEGIN%20PGP%20SIGNATURE-----%0D%0AVersion%3A%20OpenPGP.js%20v4.6.2%0D%0AComment%3A%20https%3A//openpgpjs.org%0D%0A%0D%0AwncEARYKAB8FAl2TS9MYFAAAAAAADAADdGVzdEBrZXkuY29tAQIDAAoJEGZ9%0D%0AgtV/iL8hrhMBAOQ/UgqRTbx1Z8inGmRdUx1cJU1SR4Pnq/eJNH/CFk5DAP0Q%0D%0AhUhMKMuiM3pRwdIyDOItkUWQmjEEw7/XmhgInkXsCw%3D%3D%0D%0A%3DZGXr%0D%0A-----END%20PGP%20SIGNATURE-----%0D%0A).

twiss:
Hey @wiktor-k, it's been a while, how are you? :blush:

Are you, by any chance, still interested in finishing this PR? It turns out it would also fix the bug linked above ^.^

twiss:
Ah, I see! Hope you enjoy your new house :D

Awesome, thanks! And no worries :blush:

wiktor-k:
Okay @twiss, let's take it on...

I like the `rawNotations` approach but with some tweaks:
  - we can make `notations` an Object like before but how about we `Object.freeze` it to disallow modifications and then when serializing only take into account `rawNotations`? If we want to allow modifying `notations` then we need some smart algorithm to merge it in case people modify `rawNotations` and `notations` simultaneously.
  - not having string value in `rawNotations` can be problematic. For example, the largest user of notations today is https://keyoxide.org/ and they use string notations and multiple notations of the same key. If we exposed only Uint8Arrays as values they'd need to convert them to strings too. Not a terrible thing but definitely an inconvenience. 
  - the `critical` bit. In this PR it's read only and never persisted so it's impossible to emit critical notations.

Based on this I propose the following structure:
```js
notations: Object.freeze({
  "a": "1",
  "b": "2"
}),
rawNotations: [
  {
    name: "a",
    value: {
      bytes: [49],
      string: "1"
    },
    humanReadable: true,
    critical: false
  },
  {
    name: "b",
    value: {
      bytes: [50],
      string: "2"
    },
    humanReadable: true,
    critical: false
  },
]
```

Yep, `rawNotations` would be very verbose but one could do anything with them, and that's the part I like :) If the string value cannot be computed it would be `null`.

What do you think about it? Suggestions welcome!

twiss:
Hey @wiktor-k :D

> * we can make `notations` an Object like before but how about we `Object.freeze` it to disallow modifications and then when serializing only take into account `rawNotations`?

Yeah, I'm fine with that. I'd also be fine with it without the `Object.freeze`, as it's nowhere promised that all unfrozen properties are taken into account when re-serializing a signature. :man_shrugging: But up to you.

> * not having string value in `rawNotations` can be problematic. For example, the largest user of notations today is https://keyoxide.org/ and they use string notations and multiple notations of the same key. If we exposed only Uint8Arrays as values they'd need to convert them to strings too. Not a terrible thing but definitely an inconvenience.

In my opinion, repeating the string value in two places is a bit redundant, and creating an extra object for that is a bit inefficient. Once people are using `rawNotations` as opposed to `notations`, I think then manually calling `openpgp.util.decode_utf8` is not so much to ask.

> * the `critical` bit. In this PR it's read only and never persisted so it's impossible to emit critical notations.

Hmhm. I think, ideally it would be nice to have `critical` flags stored for all subpackets, not just notations. But ok, that sounds very nontrivial, so probably this is fine for now. :+1:

twiss:
Thanks, looks good to me! :+1: Is it ready for merge?

wiktor-k:
Yep, it's OK to merge. I was pondering if I want to put the critical bit support for writing too but this is not important now. (and probably it should be addressed with support for more packets too). This change as it is now is self-contained and already an improvement IMHO.

Once we merge that I'm planning to include Justus' key in a unit test that will close his issue as this is related to this PR.

twiss:
:+1: Thanks again!

PR #997, opened 2019-10-23
Fix 996 : Own Webworker and encryption: No secure Random number generator available

### LINKED ISSUE #996: Own Webworker and encryption: No secure Random number generator available
Hello,

Thanks the advice of @twiss, I tried to use my own WebWorker with openpgp.
At the moment, it works well with String decryption but didn't work with string encryption.

Here is the simplified code I use : 

Here is the code in the main thread:

```
import PgpWorker from './pgp.worker.js'
const sendMessage = async function (message) {
  return new Promise((resolve, reject) => {
    let sharedPgpWorker = new SharedWorker((fn=>fn.match(/"(.*?)"/)[1])(""+PgpWorker))
    sharedPgpWorker.onerror = function(event) {
      reject(event)
    }
    sharedPgpWorker.port.onmessage = function (event) {
      if (event.data.err) {
        reject(event.data.err)
      } else {
        resolve(event.data)
      }
    }
    sharedPgpWorker.port.postMessage(message)
  }) 
}
const encryptMessageAndSign = async function (publicKeyReceiverList, message) {
  const workerMessage = {
    method: 'encryptMessageAndSign',
    arguments: {
      publicKeyReceiverList: publicKeyReceiverList,
      message: message
    }
  }
  return sendMessage(workerMessage)
}
```

Here is the worker code: 

```
import * as openpgp from 'openpgp'
const encryptMessageAndSign = async function (publicKeyReceiverList, message) {
  const publicKeyObjList = []
  for (const publicKey of publicKeyReceiverList) {
    const publicKeyObj = (await openpgp.key.readArmored(publicKey)).keys[0]
    publicKeyObjList.push(publicKeyObj)
  }
  const options = {
    message: openpgp.message.fromText(message), // input as Message object
    publicKeys: publicKeyObjList, // for encryption
    privateKeys: [pgpData.decryptedPrivateKey] // for signing (optional)
  }
  const cypertext = await openpgp.encrypt(options)
  return cypertext.data
}

self.onconnect = async function(event) {
  var port = event.ports[0]

  port.onmessage = async function (event) {
    try {
      const method = event.data.method
      switch (method) {
        case 'encryptMessageAndSign': {
          const { publicKeyReceiverList, message } = event.data.arguments
          const encryptedMessage = await encryptMessageAndSign(publicKeyReceiverList, message)
          port.postMessage(encryptedMessage)
          break
        }
        default: {
          throw new Error(`This method ${method} is unknow by the worker.`)
        }
      }
    } catch (err) {
      port.postMessage({ err: err })
    }
  }
}
```

And the error is: 
image

Do you have an idea ?

descampsk:
Pull request to fix #996


src/crypto/random.js
```diff
@@ -41,6 +41,8 @@ export default {
     const buf = new Uint8Array(length);
     if (typeof window !== 'undefined' && window.crypto && window.crypto.getRandomValues) {
       window.crypto.getRandomValues(buf);
+    } else if (typeof crypto !== 'undefined' && crypto && crypto.getRandomValues) {
+      crypto.getRandomValues(buf);
```
twiss: Instead of adding these two lines, could you replace the first two by them? (I.e. remove the first if and just keep this one)

descampsk: If i do that, the test aren't passing anymore.

descampsk: Done :)

src/crypto/random.js
```diff
@@ -39,8 +39,8 @@ export default {
    */
   getRandomBytes: async function(length) {
     const buf = new Uint8Array(length);
-    if (typeof window !== 'undefined' && window.crypto && window.crypto.getRandomValues) {
-      window.crypto.getRandomValues(buf);
+    if (typeof crypto !== 'undefined' && crypto && crypto.getRandomValues) {
```
twiss:
```suggestion
    if (typeof crypto !== 'undefined' && crypto.getRandomValues) {
```

descampsk:
Hehe, i'm not an expert with git :'(. I didn't manage to restore the file at its initial state.
It still stays in the pull request...

twiss:
No worries 😊 You could do e.g.
```bash
git checkout v4.6.2 dist
```
to restore the dist folder to the latest release version.

descampsk:
> 
> 
> No worries 😊 You could do e.g.
> 
> ```shell
> git checkout v4.6.2 dist
> ```
> 
> to restore the dist folder to the latest release version.

Didn't know that git checkout can work on folder ^^'. Thanks

twiss:
Thanks! :blush:

PR #999, opened 2019-10-25
use web + node crypto rsa for sign and verify operations

chesnokovilya:
Use web crypto for rsa sign and verify operations.

### DIFF
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -30,6 +30,10 @@ import prime from './prime';
 import random from '../random';
 import config from '../../config';
 import util from '../../util';
+import pkcs1 from '../pkcs1';
+import enums from '../../enums';
+
+const webCrypto = util.getWebCrypto();
 
 // Helper for IE11 KeyOperation objects
 function promisifyIE11Op(keyObj, err) {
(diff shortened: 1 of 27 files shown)


tomholub:
Thanks! This should also speed up key verification, which can be slow when importing many keys.

src/crypto/public_key/rsa.js
```diff
@@ -274,5 +313,169 @@ export default {
...
+      // switch p and q
+      prime1: new BN(q),
+      prime2: new BN(p),
+      exponent1: dq,
+      exponent2: dp,
+      coefficient: new BN(u)
+    }, 'pem', {
+      label: 'RSA PRIVATE KEY'
+    });
+    return new Uint8Array(sign.sign(pem));
```
twiss: Could you deduplicate this a bit? E.g. pull out the object.

src/crypto/public_key/rsa.js
```diff
@@ -274,5 +313,169 @@ export default {
...
+      modulus: new BN(n),
+      publicExponent: new BN(e)
+    }, 'pem', {
+      label: 'RSA PUBLIC KEY'
+    });
+    try {
+      return await verify.verify(key, s);
+    } catch (err) {
+      return false;
+    }
```
twiss: And this as well. Maybe deduplicate both the object and the try/catch, if possible.

PR #1005, opened 2019-11-20
Switch code coverage reporter to nyc

chesnokovilya:
Istambul code reporter contains critical vulnerability, it is no longer maintained, so in this PR openpgpjs would be switched to `nyc` coverage reporter. In the future we could update to automatic coverage reports using coveralls.io


srozell4:
Is there a status on getting this merged? It would be great to clear out the critical vulnerability in the nested dependencies.

chesnokovilya:
It would probably be merged around the end of January. I need to add some functionality to it.

PR #1017, opened 2019-12-16
Only throw on authorized revocation key when verifying self-signatures

twiss:
This also has the effect that we only throw on them when trying to use the key, instead of when parsing it, and that we don't throw when the authorized revocation key is specified in a separate direct-key signature instead of a User ID self-signature. Since rfc4880bis-05, the spec only specifies including it in a direct-key signature (see https://gitlab.com/openpgp-wg/rfc4880bis/commit/ac655aa4f2e52cdbd691eb2ad4d036fd1dc02581), so that means that we effectively don't reject them anymore for implementations that conform to that. This is because users that wanted to use the key could remove this separate signature, anyway.

This addresses point 3 of #808.

@toberndo Since you wrote that PR, lmk if you see any issues with this one. Specifically, this code no longer explicitly checks that the signature is a self-signature, however, I think it shouldn't be an issue since third-party signatures are not verified during normal usage of a key.

### DIFF
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -275,19 +275,6 @@ export function getExpirationTime(keyPacket, signature) {
 }
 
 /**
- * Check if signature has revocation key sub packet (not supported by OpenPGP.js)
- * and throw error if found
- * @param {module:packet.Signature} signature The certificate or signature to check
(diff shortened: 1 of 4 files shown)


twiss:
(Lmk if any issues come up)

PR #1043, opened 2020-02-07
Reject signatures using insecure hash algorithms

twiss:
- Reject all signatures using MD5 or RIPEMD by default
- Reject message signatures using MD5, RIPEMD or SHA1 by default

Also, switch from returning false to throwing errors in most `verify*()` functions, as well as `await signatures[*].verified`, in order to be able to show more informative error messages.

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -210,5 +210,15 @@ export default {
    * @memberof module:config
    * @property {Object} indutny_elliptic_fetch_options Options object to pass to `fetch` when loading the indutny/elliptic library. Only has an effect if `config.external_indutny_elliptic` is true.
    */
-  indutny_elliptic_fetch_options: {}
+  indutny_elliptic_fetch_options: {},
+  /**
+   * @memberof module:config
+   * @property {Set<Integer>} reject_hash_algorithms Reject insecure hash algorithms {@link module:enums.hash}
+   */
+  reject_hash_algorithms: new Set([enums.hash.md5, enums.hash.ripemd]),
+  /**
+   * @memberof module:config
+   * @property {Set<Integer>} reject_message_hash_algorithms Reject insecure message hash algorithms {@link module:enums.hash}
+   */
+  reject_message_hash_algorithms: new Set([enums.hash.md5, enums.hash.ripemd, enums.hash.sha1])
 };
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -215,10 +215,10 @@ export default {
    * @memberof module:config
    * @property {Set<Integer>} reject_hash_algorithms Reject insecure hash algorithms {@link module:enums.hash}
    */
-  reject_hash_algorithms: new Set([enums.hash.md5, enums.hash.ripemd]),
+  reject_hash_algorithms: new global.Set([enums.hash.md5, enums.hash.ripemd]),
   /**
    * @memberof module:config
    * @property {Set<Integer>} reject_message_hash_algorithms Reject insecure message hash algorithms {@link module:enums.hash}
    */
-  reject_message_hash_algorithms: new Set([enums.hash.md5, enums.hash.ripemd, enums.hash.sha1])
+  reject_message_hash_algorithms: new global.Set([enums.hash.md5, enums.hash.ripemd, enums.hash.sha1])
 };
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -38,16 +38,28 @@ export async function generateSecretKey(options) {
  */
 export async function getLatestValidSignature(signatures, primaryKey, signatureType, dataToVerify, date = new Date()) {
   let signature;
+  let exception;
   for (let i = signatures.length - 1; i >= 0; i--) {
-    if (
-      (!signature || signatures[i].created >= signature.created) &&
-      // check binding signature is not expired (ie, check for V4 expiration time)
-      !signatures[i].isExpired(date) &&
-      // check binding signature is verified
-      (signatures[i].verified || await signatures[i].verify(primaryKey, signatureType, dataToVerify))
-    ) {
-      signature = signatures[i];
+    try {
+      if (
+        (!signature || signatures[i].created >= signature.created) &&
+        // check binding signature is not expired (ie, check for V4 expiration time)
+        !signatures[i].isExpired(date) &&
+        // check binding signature is verified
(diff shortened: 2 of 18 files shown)


src/config/config.js
```diff
@@ -210,5 +210,15 @@ export default {
    * @memberof module:config
    * @property {Object} indutny_elliptic_fetch_options Options object to pass to `fetch` when loading the indutny/elliptic library. Only has an effect if `config.external_indutny_elliptic` is true.
    */
-  indutny_elliptic_fetch_options: {}
+  indutny_elliptic_fetch_options: {},
+  /**
+   * @memberof module:config
+   * @property {Array<Integer>} reject_hash_algorithms Reject insecure hash algorithm {@link module:enums.hash}
+   */
+  reject_hash_algorithms: new global.Set([enums.hash.md5, enums.hash.ripemd]),
```
chesnokovilya: Why Use global.Set here? Seems inconsistent with how other config values are set.

chesnokovilya: if other config variable is reject message hash, than may be this one should be "reject_key_hash_algorithms"?

src/key/helper.js
```diff
@@ -38,17 +38,29 @@ export async function generateSecretKey(options) {
...
+        (!signature || signatures[i].created >= signature.created) &&
+        // check binding signature is not expired (ie, check for V4 expiration time)
+        !signatures[i].isExpired(date) &&
+        // check binding signature is verified
+        (signatures[i].verified || await signatures[i].verify(primaryKey, signatureType, dataToVerify))
+      ) {
+        signature = signatures[i];
+      }
+    } catch (e) {
+      exception = e;
```
chesnokovilya: If there is array of signatures, than there could be an array of exceptions?

src/key/helper.js
```diff
@@ -106,7 +118,7 @@ export async function getPreferredHashAlgo(key, keyPacket, date = new Date(), us
   let pref_algo = hash_algo;
   if (key) {
     const primaryUser = await key.getPrimaryUser(date, userId);
-    if (primaryUser && primaryUser.selfCertification.preferredHashAlgorithms) {
+    if (primaryUser.selfCertification.preferredHashAlgorithms) {
```
chesnokovilya: May be we could get rid of "if inside if" here. This is up to you.

src/key/key.js
```diff
@@ -413,7 +418,7 @@ Key.prototype.validate = async function() {
   const signatureType = enums.signature.binary;
   signature.signatureType = signatureType;
   await signature.sign(signingKeyPacket, data);
-  return signature.verify(signingKeyPacket, signatureType, data);
+  await signature.verify(signingKeyPacket, signatureType, data);
```
chesnokovilya: on line 417 `hashAlgorithm` is set statically. Should it be checked if it is safe one? It seems that better way to set it would be in config.

test/general/signature.js
```diff
@@ -478,41 +478,52 @@ vwjE8mqJXetNMfj8r2SCyvkEnlVRYR+/mnge+ib56FdJ8uKtqSxyvgA=
...
 
   it('Testing signature checking on CAST5-enciphered message', async function() {
-    const priv_key = (await openpgp.key.readArmored(priv_key_arm1)).keys[0];
-    const pub_key = (await openpgp.key.readArmored(pub_key_arm1)).keys[0];
-    const msg = await openpgp.message.readArmored(msg_arm1);
-    await priv_key.decrypt("abcd");
-    return openpgp.decrypt({ privateKeys: priv_key, publicKeys:[pub_key], message:msg }).then(function(decrypted) {
+    const { reject_message_hash_algorithms } = openpgp.config;
+    Object.assign(openpgp.config, { reject_message_hash_algorithms: new Set([openpgp.enums.hash.md5, openpgp.enums.hash.ripemd]) });
+    try {
```
chesnokovilya: Why use try in tests? It makes tests more confusing in a way.

src/config/config.js
twiss: > Why Use global.Set here? Seems inconsistent with how other config values are set.

I initially made an array, but I switched to a Set so that you can easily do:

```js
openpgp.config.reject_message_hash_algorithms.delete(openpgp.enums.hash.sha1);
```

The only other array in config.js is `known_notations`, there I think it's less necessary to be able to easily remove them?

The `global.Set` instead of `Set` is because of https://github.com/openpgpjs/openpgpjs/commit/7bb8b69044f9b3309b5f31f0ba02d03de43276eb.

> if other config variable is reject message hash, than may be this one should be "reject_key_hash_algorithms"?

Maybe, but we might in the future also want to use it for other places where hashes are used (like S2K). I think long term the other config (`reject_message_hash_algorithms`) should eventually be removed again and we'll keep just this one.

src/key/helper.js
twiss: Theoretically yes, but there's a lot of other places in the library where we only give one error even if multiple can occur, e.g. https://github.com/openpgpjs/openpgpjs/blob/9bdeaa927ace1ef32ce025140651966e2e7751eb/src/key/key.js#L378-L389
I think in practice it's uncommon enough that it's not worth the effort to concatenate all the error messages or something. If someone wants more info they can debug the key manually.

src/key/helper.js
twiss: This `if` depends on `primaryUser`, which depends on the other `if` - so we can't merge them, no?

src/key/key.js
twiss: We could use `config.prefer_hash_algorithm` there, yeah. But that signature is just temporary, we don't export it, and we should eventually check that the key parameters correspond directly without making a signature there.

test/general/signature.js
twiss: The `try/finally` is to make sure that the config gets restored if the test fails (throws).

Maybe since it's happening a lot we could organize it in a bit more structured way (e.g. with `beforeEach`/`afterEach`, as is already done in some places) but when it's just for some of the tests I think it's still fairly clear what's going on.

src/key/helper.js
chesnokovilya: I think javascript has lazy if evaluation, so we could create composite condition inside if.

twiss: Yes but you'd need to create an inline assignment inside the if condition, I don't think that's really cleaner.

PR #1044, opened 2020-02-08
Optimize reading and writing armored messages

twiss:


src/encoding/armor.js
```diff
@@ -126,53 +126,64 @@ function getCheckSum(data) {
...
+  crc_table[1][i] = (crc_table[0][i] >> 8) ^ crc_table[0][crc_table[0][i] & 0xFF];
+}
+for (let i = 0; i <= 0xFF; i++) {
+  crc_table[2][i] = (crc_table[1][i] >> 8) ^ crc_table[0][crc_table[1][i] & 0xFF];
+}
+for (let i = 0; i <= 0xFF; i++) {
+  crc_table[3][i] = (crc_table[2][i] >> 8) ^ crc_table[0][crc_table[2][i] & 0xFF];
+}
+
+const isLittleEndian = (function() {
```
chesnokovilya: Why use IEFE here? Could ordinary function or closure work here?

src/encoding/armor.js
```diff
@@ -126,53 +126,64 @@ function getCheckSum(data) {
...
-  0x51f6d10c,
-  0x50fa48fa, 0x507c0401, 0x5542fa2f, 0x55c4b6d4, 0x54c82f22, 0x544e63d9, 0x56d11cce, 0x56575035, 0x575bc9c3,
-  0x57dd8538
+  new Array(0xFF),
+  new Array(0xFF),
+  new Array(0xFF),
+  new Array(0xFF)
 ];
 
+for (let i = 0; i <= 0xFF; i++) {
```
chesnokovilya: May be it would be cleaner to create single function or closure for crc_table generation, instead of dumping all these FOR in the main body?

src/encoding/base64.js
```diff
@@ -13,121 +13,82 @@
...
-        r.push('=');
-        l += 1;
-      }
-    }
-    if (s === 1 && !u) {
-      if ((l % 60) === 0 && !u) {
-        r.push("\r\n");
-      }
-      r.push('=');
+    const lines = Math.floor(buf.length / 45);
```
chesnokovilya: What is 45 by the way? May be you could specify it as variable?

src/encoding/armor.js
twiss: The idea is because the function only has to be called once (because the endianness of the platform never changes) it is not necessary to create a function for this. The IEFE is to encapsulate the temporary `buffer` variable.
Btw this code was more or less copied from MDN, I'll add a comment.

src/encoding/armor.js
twiss: Yeah, maybe. I think since they're not creating any extra variables, I think it's not really necessary to encapsulate this in an IEFE, for example (and again the code only needs to run once).

src/encoding/base64.js
twiss: It's the number of bytes per line (60 chars per line * (3 bytes / 4 chars of base64)). I'll add a constant, good point.

PR #1047, opened 2020-02-19
WIP: V5

twiss:


src/key/factory.js
```diff
@@ -266,61 +263,67 @@ async function wrapKeyObject(secretKeyPacket, secretSubkeyPackets, options) {
...
-      }
-    }
-  } catch (e) {
-    err.push(e);
-  }
-  if (err.length) {
-    result.err = err;
+  const packetlist = new packet.List();
+  await packetlist.read(data);
+  return new Key(packetlist);
```
chesnokovilya, line 275: It would throw if the key is bad or armored. Open questions if we should catch it?

twiss: I think it's better to throw than to return an error array. Most people don't check it, and are confused when the `keys` array is empty.

tomholub:
Just wanted to say, PRs that remove code are my favorite :smiley: 

image

twiss:
Yes :smile: To be fair, a lot of that is removing the `dist` folder (people can get it from npm or unpkg.com instead), but still despite that we've removed quite some code as well :blush:

PR #1052, opened 2020-03-02
Fix "TypeError: fetch is not a function" in Node.js environment

Umoxfo:
The HKP.lookup method throws "TypeError: fetch is not a function" in Node.js environment.


twiss:
Thanks! :+1:

PR #1059, opened 2020-03-12
readArmored() fails because of CR/LF etc.

pbrunschwig:
I found that key.readArmored() fails for simple keys. The reason is that atob does not like newlines and whitespace, which are still present when trying to perform the conversion. 
 
I assume that this is a problem with all readArmored() functions.

This PR addresses the issue by removing any CR, LF, space and tab characters before doing atob()


twiss:
Hey :wave: What environment are you running this in? And are you sure your `atob` is the native one? Because `atob` should strip whitespace. I did find this:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/atob#Polyfill
> ```js
> // Some browsers' implementation of atob doesn't support whitespaces
> // in the encoded string (notably, IE). This wraps the native atob
> // in a function that strips the whitespaces.
> ```
So you could try running that and see if that fixes it.

Also: https://bugzilla.mozilla.org/show_bug.cgi?id=711180 (fixed in.. Firefox 27 🤷‍♂).

pbrunschwig:
I'm running Thunderbird 52 and 68. You're right, the native atob doesn't care about it. So it probably uses the polyfill. 

It's extremely difficult to load the library into privileged Thunderbird code (for Enigmail). I'm almost certain that it doesn't use any of the native functions :-(

PR #1062, opened 2020-03-19
[v5] Use AEAD protection by default

larabr:


test/general/streaming.js
```diff
@@ -234,6 +234,8 @@ function tests() {
   });
 
   it('Encrypt and decrypt larger message roundtrip', async function() {
+    let aead_protectValue = openpgp.config.aead_protect;
```
chesnokovilya: Hey, In my opinion it would be better to use consistent casing. We use camelCase as default one.

PR #1071, opened 2020-03-27
Support verification of messages with charset != UTF-8

### LINKED ISSUE #1064: Unable to verify messages with charset != UTF-8
Many mails are (still) encoded in various character sets such as Latin1 (ISO-8859-15) or BIG5 (Japanese). I haven't found  a way to verify such messages with OpenPGP.js. 

I roughly do the following:

```javascript
  var msg = await openpgp.cleartext.readArmored(msgData);
  var verified = await openpgp.verify({
        message: msg,
        publicKeys: pubKeys
    });
```

This works fine with UTF-8 encoded data, but not with anything else. GnuPG verifies the message successfully.

Attached is a sample message that is encoded in Latin1, together with the key needed to verify the message. I created a ZIP such that the encoding wouldn't change during submission.

[msg.zip](https://github.com/openpgpjs/openpgpjs/files/4355888/msg.zip)

larabr:
Fixes #1064:
Enable verification of messages with any charset that encodes line endings like ASCII (CR=0x0D and LF=0x0A). This includes UTF-8, Latin-1, Big5...


src/cleartext.js
```diff
@@ -43,7 +43,7 @@ export function CleartextMessage(text, signature) {
     return new CleartextMessage(text, signature);
   }
   // normalize EOL to canonical form <CR><LF>
-  this.text = util.canonicalizeEOL(util.removeTrailingSpaces(text));
+  this.text = normalizeEOL(util.removeTrailingSpaces(text), true);
```
twiss: Instead of doing two different things with one function, could you do something like
```suggestion
  this.text = normalizeEOL(util.removeTrailingSpaces(text)).replace(/[\r\n]/g, '\r\n');
```
or even inline `normalizeEOL` in both places (since it will then be very short)?

PR #1072, opened 2020-03-30
[v5] Remove worker

larabr:


tomholub:
Will the worker be developed in a separate repo?

Or is it getting abandoned?

twiss:
We might set up an openpgpjs-worker-example as you suggested, however, to reduce copying of data, it's probably better to send the worker sort of "application-level" tasks (e.g. "send an email", not "encrypt some text"); where the Worker might even communicate with the server, etc.

So it might therefore make more sense to set up an "openpgpjs-application-example", which would include a worker among other things. Then again, such an example might not be that useful because the application-level tasks are, you know, application-specific. In any case, I'm not quite sure yet what we should do.

For what it's worth, v5 will still take a while to be released (probably in the summer or later), so we still have some time to think about this. (We would like to use v5 keys in the new version, but this is blocked on support in GnuPG, which will be in their next release.)

(P.S.: In a browser extension, the "background page" is already running off the main thread, which is the main purpose of using a worker. Have you tried not using a worker at all? It may in fact be faster.)

tomholub:
> (P.S.: In a browser extension, the "background page" is already running off the main thread, which is the main purpose of using a worker. Have you tried not using a worker at all? It may in fact be faster.)

Thanks for the hint. I have the background script itself further delegate things to a worker. I don't want to block the background script because it may have other things to do concurrently.

I understand your overall reasoning, it's too inefficient to have the worker integrated tightly with the internals of OpenPGP.js, harder to cache, pass around objects etc and probably more difficult for development, testing and so on.

I'll think about moving to application-level workers, it might work just as well, or better.

Not sure an example will be that useful, as you say, it's app specific.

Thanks!

pbrunschwig:
> We might set up an openpgpjs-worker-example as you suggested, however, to reduce copying of data, it's probably better to send the worker sort of "application-level" tasks (e.g. "send an email", not "encrypt some text"); where the Worker might even communicate with the server, etc.

I agree that this would make sense implementation-wise. However, for the purpose of demonstrating how to implement a worker, it would probably make sense to create a much smaller worker that just does the basic tasks, maybe with just a few comments where things like sending mails etc would be placed. For example in Enigmail, I support different OpenPGP implementations that need to implement a given API. As you can imagine, sending an email would not be part of such a low-level function.

PR #1073, opened 2020-04-02
[v5] Drop support for \r as EOL

larabr:
`\r` was used as EOL by Mac OS 9. With this change we will only support `\r\n` and `\n` as EOL.

### DIFF
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -43,7 +43,7 @@ export function CleartextMessage(text, signature) {
     return new CleartextMessage(text, signature);
   }
   // normalize EOL to canonical form <CR><LF>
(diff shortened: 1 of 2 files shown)


src/cleartext.js
```diff
@@ -43,7 +43,7 @@ export function CleartextMessage(text, signature) {
     return new CleartextMessage(text, signature);
   }
   // normalize EOL to canonical form <CR><LF>
-  this.text = util.removeTrailingSpaces(text).replace(/\r\n/g, '\n').replace(/[\r\n]/g, '\r\n');
+  this.text = util.removeTrailingSpaces(text).replace(/\r\n/g, '\n').replace(/\n/g, '\r\n');
```
twiss, line 46:
```suggestion
  this.text = util.removeTrailingSpaces(text).replace(/\r?\n/g, '\r\n');
```

PR #1085, opened 2020-04-20
add algorithm=uncompressed handler

chesnokovilya:


src/packet/compressed.js
```diff
@@ -132,6 +132,12 @@ export default Compressed;
 
 const nodeZlib = util.getNodeZlib();
 
+function uncompressed() {
+  return function (data) {
+    return data;
+  };
```
twiss, line 138: You don't have to make a wrapper function for this, you can call the inner function `uncompressed` and do something like `uncompressed: uncompressed` further down.

PR #1095, opened 2020-05-06
[v5] Switch build system to rollup

twiss:
This also:
- Removes the `dist` folder from the main repository in favor of npm and unpkg.com, and building on `npm install` when fetching from GitHub.
- Switches to the minified builds by default.
- Adds sourcemaps to the minified builds.
- Prepares to add the `docs` folder to the main repository during release (instead of keeping it in a separate branch), so that the docs will automatically be kept up to date during every release.
- Removes the compat build and compat polyfills (for IE11 and old Safari). If needed, they will have to be added back by the application's build system.
- Remove Node.js 8 support and Travis test (it should still work, but we can no longer test it
because rollup requires Node.js 10+).
- Uses dynamic `import()` in the lightweight build instead of dynamically loading a `<script>` tag.
- Removes the non-module lightweight build.
- And a bunch of other stuff, see commit messages.


src/packet/marker.js
```diff
@@ -32,31 +34,32 @@ import enums from '../enums';
  *
  * Such a packet MUST be ignored when received.
  * @memberof module:packet
- * @constructor
  */
-function Marker() {
-  this.tag = enums.packet.marker;
-}
+class MarkerPacket {
```
chesnokovilya, line 38: Why rename class?

src/packet/all_packets.js
```diff
@@ -4,75 +4,73 @@
  * @module packet/all_packets
  */
 
-import * as packets from './all_packets.js'; // re-import module to parse packets from tag
-
 export {
-  /** @see module:packet.Compressed */
-  default as Compressed
-} from './compressed.js';
+  /** @see CompressedDataPacket */
```
chesnokovilya, line 8: Those long name are kinda ugly in my opinion, may be just use CompressedData or Compressed?

src/packet/marker.js
twiss: All packet classes are renamed like this, the reason is so that you can do
```js
import { MarkerPacket } from 'openpgp';
```
and only import the packets you need, instead of having to do
```js
import { packet } from 'openpgp';
// do something with packet.Marker
```
This helps with tree shaking.

src/message.js
```diff
@@ -36,494 +36,610 @@ import config from './config';
...
+ * Class that represents an OpenPGP message.
  * Can be an encrypted message, signed message, compressed message or literal message
- * @param  {module:packet.List} packetlist The packets that form this message
  * See {@link https://tools.ietf.org/html/rfc4880#section-11.3}
  */
-
-export function Message(packetlist) {
-  if (!(this instanceof Message)) {
-    return new Message(packetlist);
+export class Message {
```
chesnokovilya, line 60: In my opinion it would be improvement to split Message file to Message class and other functions. That way one could use `export default class` here.

src/packet/all_packets.js
twiss: ["Compressed Data Packet"](https://tools.ietf.org/html/rfc4880#section-5.6) is the official name of the packet, this way all the names of the classes are consistent with the names of the packets.

src/message.js
twiss: Yeah, that could make sense :+1: I guess to be consistent, we'd have to do the same for files like `src/signature.js` though, which are already very small. I'll look into it.

package.json
```diff
@@ -13,76 +13,67 @@
...
-    "grunt-jsdoc": "^2.2.1",
-    "grunt-mocha-test": "^0.13.3",
-    "grunt-simple-nyc": "^3.0.1",
-    "grunt-terser": "^0.1.0",
-    "grunt-text-replace": "~0.4.0",
-    "gruntify-eslint": "^4.0.0",
+    "eslint-plugin-import": "^2.8.0",
+    "hash.js": "^1.1.3",
+    "http-server": "^0.12.3",
+    "jsdoc": "github:twiss/jsdoc#0f1816eb4553856647b4ca9561b9307b11ec4f9e",
```
chesnokovilya, line 63: is using fork really required?

src/key/factory.js
```diff
@@ -270,8 +270,8 @@ async function wrapKeyObject(secretKeyPacket, secretSubkeyPackets, options) {
  * @static
  */
 export async function read(data) {
-  const packetlist = new packet.List();
-  await packetlist.read(data);
+  const packetlist = new PacketList();
+  await packetlist.read(data, helper.allowedKeyPackets);
```
chesnokovilya, line 274: do we need `allowedKeyPackets` here? Maybe instead of helper it is better to place them in Key class.

src/key/key.js
```diff
@@ -26,826 +26,836 @@
...
-  if (!this.keyPacket || !this.users.length) {
-    throw new Error('Invalid key: need at least key and user ID packet');
+ * @borrows PublicKeyPacket#getKeyId as Key#getKeyId
+ * @borrows PublicKeyPacket#getFingerprint as Key#getFingerprint
+ * @borrows PublicKeyPacket#hasSameFingerprintAs as Key#hasSameFingerprintAs
+ * @borrows PublicKeyPacket#getAlgorithmInfo as Key#getAlgorithmInfo
+ * @borrows PublicKeyPacket#getCreationTime as Key#getCreationTime
+ * @borrows PublicKeyPacket#isDecrypted as Key#isDecrypted
+ */
+class Key {
```
chesnokovilya, line 52: we could use `export default class Key` here.

rollup.config.js
```diff
@@ -0,0 +1,117 @@
...
+      commonjs(),
+      replace({
+        'OpenPGP.js VERSION': `OpenPGP.js ${pkg.version}`,
+      })
+    ]
+  },
+  {
+    input: 'src/index.js',
+    output: [
+      { dir: 'dist/lightweight', entryFileNames: 'openpgp.mjs', chunkFileNames: '[name].mjs', format: 'es', banner, intro },
```
chesnokovilya, line 75: Maybe we could create only minified builds by default?

src/packet/all_packets.js
chesnokovilya: Yeah, but this is not documentation, it is common to left out common name if classes belong to a same folder. For example Sequoia PGP https://gitlab.com/sequoia-pgp/sequoia/-/blob/master/openpgp/src/packet/compressed_data.rs

package.json
twiss: The original copies the entire source code into the docs directory. I wanted to make it smaller. See https://github.com/jsdoc/jsdoc/pull/1770. I'll move the fork from `twiss` to `openpgp` before merging.

src/key/factory.js
twiss: Both `key.js` and `factory.js` need it, which is why I put it in `helper.js`.

src/key/key.js
twiss: Yeah. bc6ddc6078464cf44d3b361af0c4c08ca6ee6a6d just mechanically changed prototype based classes to ES6 classes, and didn't change anything else. We can make things prettier later.

rollup.config.js
twiss: The non-minified builds are useful for seeing what the build actually does / looks like.

PR #1097, opened 2020-05-11
Also create issuer fingerprint subpacket for v4 keys.

Mento:
Do not limit creation of signatures with issuer fingerprint subpacket to v5 keys.

This allows a recipient using GnuPG to get the key via auto-key-retrieve from a key server (if the key is already on the server of course).
If the issuer fingerprint subpacket is missing, GnuPG would not automatically download the key.


twiss:
Hey :wave: Interesting. Thanks for the PR! :+1:

Small nitpick: we only support v4 and v5 keys, so I think you can remove the if statement and comment entirely.

twiss:
Thanks!

PR #1100, opened 2020-05-17
Fix keyId documentation so that typescript works

Gaff:
Would it be ok to fix this comment so that typescript can pickup the correct type for the getKeyId function?

Thanks!

Downstream issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/44736


twiss:
Yes, thanks for the PR! :+1: Merged with some small modifications to the new comments :)

PR #1103, opened 2020-05-29
await getReader()

drzraf:
From experiments, `getReader()` may not be immediately ready (at least in case a lot of input data is to be compression).
In these case, the first `await reader.read() / readBytes()` would return undefined unless we explicitely waited for the reader.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ Where the value can be any of:
 
     // Either pipe the above stream somewhere, pass it to another function,
     // or read it manually as follows:
-    const reader = openpgp.stream.getReader(encrypted);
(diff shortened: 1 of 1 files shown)


twiss:
Ehm, that's really strange. Do you have a code example of this happening?

In any case, `getReader` doesn't return a Promise, so it shouldn't need `await` - if it does, it's broken and should probably be fixed somewhere else.

drzraf:
Sorry, the origin of my problem was obviously distinct.
I was firing `readByte()` before `{ openpgp.encrypt() & stream.getReader() }` could actually complete.
(https://github.com/flowjs/flow.js/pull/296)

PR #1112, opened 2020-06-25
Handle message.fromStream when cloning & restoring Messages

### LINKED ISSUE #1109: Streamed Sign & Encrypt hangs in worker due to undefined message.fromStream
When signing and encrypting a binary message constructed from a stream, while using the worker, the call to `openpgp.encrypt` never finishes (tested with OpenPGP.js v4.10.4).

While debugging the issue, I found that encryption only (without signing) works fine, the issue only happens with a worker enabled and also there is no issue without using streams.

- ✔️Without worker
- ✔Encrypt only with binary stream & worker
- ✔️Encrypt & sign with binary array & worker
- ✔Encrypt only with binary array & worker
- ❌Encrypt & sign with binary stream & worker

I think, that the issue is located in [line 330 of openpgp.js](https://github.com/openpgpjs/openpgpjs/blob/v4.10.4/src/openpgp.js#L330). During debugging I found, that `message.fromStream` is undefined while the `encrypt` executes in the worker. Probably the same issue can also be found in [line 327](https://github.com/openpgpjs/openpgpjs/blob/v4.10.4/src/openpgp.js#L327).

I found that replacing `message.fromStream` line 330 with `streaming` resolves the issue for me. Since in [line 334](https://github.com/openpgpjs/openpgpjs/blob/v4.10.4/src/openpgp.js#L334) also `streaming` is used instead of `message.fromStream`, I suspect that for the other lines it also has to be changed to `streaming`.

I made a proof of concept here: 

https://codepen.io/cpupower/project/full/AOgMpv

cpupower:
This should fix #1109. 
Previously, signing & encrypting a message created from a stream would hang in the worker, because `Message.fromStream` is undefined in the worker. This is because the `fromStream` property is not properly cloned/restored when passing options to a worker.

The proposed solution changes the `clonePackets` function to wrap the `Message` packets in an object (similar to how it is done for `CleartextMessage`), while including the `fromStream` property. The `packetlistCloneToMessage` is modified accordingly, such that the property is restored afterwards.

See #1109 for more details.

Changes:

- Include fromStream property when cloning a Message
- Restore fromStream property in packetlistCloneToMessage

### DIFF
--- a/src/packet/clone.js
+++ b/src/packet/clone.js
@@ -63,7 +63,7 @@ export function clonePackets(options) {
   if (options.message) {
     //could be either a Message or CleartextMessage object
     if (options.message instanceof Message) {
-      options.message = options.message.packets;
+      options.message = { packets: options.message.packets, fromStream: options.message.fromStream };
     } else if (options.message instanceof CleartextMessage) {
       options.message = { text: options.message.text, signature: options.message.signature.packets };
     }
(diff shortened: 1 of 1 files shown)


twiss:
Thanks for the PR! :+1:

PR #1115, opened 2020-06-29
Add support for advanced wkd links

YarmoM:
As per the spec[0], web key directory implementations must first try the advanced URL and fall back to the basic URL. This small  PR aims to implement this mechanism in openpgpjs.

There must be more than one way of allowing a fetch statement to fall back to a different URL. This is the cleanest way I could find and hope it's up to the standards of the main devs. Open to suggestions!

[0] https://tools.ietf.org/html/draft-koch-openpgp-webkey-service-10


twiss:
Hey @YarmoM, thanks for your PR!

I think this function could potentially be cleaner if it used `await` instead of `then()`. Then(:upside_down_face:) you could just `return` out of the function on failure instead of having to check `if (publicKey)` repeatedly. However, I think this is somewhat out of the scope of this PR, and I will merge it as is. I'd be grateful if you want to make another PR, but it's not required ^.^

PR #1119, opened 2020-07-02
[v5] BigInt Support

larabr:
Use the native BigInt library when available instead of bn.js

### DIFF
--- a/src/biginteger/bn.interface.js
+++ b/src/biginteger/bn.interface.js
@@ -9,8 +9,13 @@ export default class BigInteger {
   /**
    * Get a BigInteger (input must be big endian for strings and arrays)
    * @param {Number|String|Uint8Array} n value to convert
+   * @throws {Error} on undefined input
    */
   constructor(n) {
+    if (n === undefined) {
+      throw new Error('Invalid BigInteger input');
+    }
+
     this.value = new BN(n);
   }
 
@@ -108,24 +113,6 @@ export default class BigInteger {
   }
 
   /**
-   * BigInteger division in place
-   * @param {BigInteger} x value to divide
-   */
-  idiv(x) {
-    this.value = this.value.div(x.value);
-    return this;
-  }
-
-  /**
-   * BigInteger division
-   * @param {BigInteger} x value to divide
-   * @returns {BigInteger} this / x
-   */
-  div(x) {
-    return this.clone().idiv(x);
-  }
-
-  /**
    * Compute remainder in place
    * This is not the same as mod(m):
    *  the remainder is negative if the value is negative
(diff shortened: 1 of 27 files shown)


src/biginteger/native.interface.js
```diff
@@ -0,0 +1,575 @@
...
+   * @param {Number|String|Uint8Array} n value to convert
+   */
+  constructor(n) {
+    if (n instanceof Uint8Array) {
+      const bytes = n;
+      let hex = '0x0';
+      for (let i = 0; i < n.length; i++) {
+        const hexByte = bytes[i].toString(16);
+        hex += (bytes[i] <= 0xF) ? ('0' + hexByte) : hexByte;
+      }
```
twiss, line 22: This is quite nitpicky, but when constructing long strings, it's usually faster to create an array and then at the end `.join('')` the parts together.

src/crypto/public_key/dsa.js
```diff
@@ -148,25 +148,24 @@ export default {
...
     }
 
     /**
      * Check q is large and probably prime (we mainly want to avoid small factors)
      */
-    const qSize = q.bitLength();
-    if (qSize < 150 || !(await prime.isProbablePrime(q, null, 32))) {
+    const qSize = new BigInteger(q.bitLength());
+    const n150 = new BigInteger(150);
+    if (qSize.lt(n150) || !(await prime.isProbablePrime(q, null, 32))) {
```
twiss, line 168: These don't need to be `BigInteger`s, right? Since both are coming from a `Number`, they're guaranteed to fit in a `Number`

src/crypto/public_key/elgamal.js
```diff
@@ -76,29 +68,29 @@ export default {
...
     if (g.lte(one) || g.gte(p)) {
       return false;
     }
 
     // Expect p-1 to be large
-    const pSize = p.subn(1).bitLength();
-    if (pSize < 1023) {
+    const pSize = new BigInteger(p.bitLength());
+    const n1023 = new BigInteger(1023);
+    if (pSize.lt(n1023)) {
```
twiss, line 85: Idem

src/crypto/pkcs1.js
```diff
@@ -76,50 +75,46 @@ async function getPkcs1Padding(length) {
...
   let i = 2;
-  while (EM.charCodeAt(i) !== 0 && i < EM.length) {
+  while (encoded[i] !== 0 && i < encoded.length) {
     i++;
   }
   const psLen = i - 2;
-  const separator = EM.charCodeAt(i++);
-  if (firstOct === 0 && secondOct === 2 && psLen >= 8 && separator === 0) {
-    return EM.substr(i);
+  const separator = encoded[i++];
```
larabr, line 115: Check that this does not through if `i = length`
Also add invalid padding tests

src/biginteger/native.interface.js
```diff
@@ -0,0 +1,441 @@
+/* eslint-disable new-cap */
```
larabr, line 1: Can remove

src/biginteger/bn.interface.js
```diff
@@ -0,0 +1,366 @@
...
+    return this.clone().imul(x);
+  }
+
+  /**
+   * Compute remainder in place
+   * This is not the same as mod(m):
+   *  the remainder is negative if the value is negative
+   * @param {BigInteger} m divisor
+   */
+  iremainder(m) {
```
larabr, line 121: Can remove

src/biginteger/bn.interface.js
```diff
@@ -0,0 +1,366 @@
...
+   */
+  mod(m) {
+    return this.clone().imod(m);
+  }
+
+  /**
+   * Exponentiate this to the power of x, in place
+   * @param {BigInteger} x exponent
+   */
+  iexp(x) {
```
larabr, line 160: can remove

PR #1120, opened 2020-07-06
Fix comments in sym and pub key encrypted key packets

danrr:
The comments for the symmetrically encrypted session key packets did not reflect what was going on in the class. It looks like a copy-paste job that wasn't updated at the time.

### DIFF
--- a/src/packet/sym_encrypted_session_key.js
+++ b/src/packet/sym_encrypted_session_key.js
@@ -30,20 +30,17 @@ import enums from '../enums';
 import util from '../util';
 
 /**
- * Public-Key Encrypted Session Key Packets (Tag 1)
+ * Symmetric-Key Encrypted Session Key Packets (Tag 3)
  *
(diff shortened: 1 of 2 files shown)


src/packet/sym_encrypted_session_key.js
```diff
@@ -105,6 +97,11 @@ class SymEncryptedSessionKeyPacket {
     }
   }
 
+  /**
+   * Create a string representation of a tag 3 packet
```
twiss, line 101:
```suggestion
   * Create a binary representation of a tag 3 packet
```
I know you copied this from somewhere that has it wrong too, feel free to fix it there too, if you feel like

danrr: Yeah, I copied it without really thinking if it made sense in the original use. Fixed in both places.

twiss:
:+1: Thanks!

PR #1123, opened 2020-07-14
Improve handling of multiple WKD URLs

YarmoM:
I am a bit unsure how exactly @twiss envisioned in #1115 replacing `then()` by `await` to get a cleaner result. Hope this is it :)

If not, do let me know and I'll amend the PR.


twiss:
Hey :wave: Yes, this is a good start :D Thanks!

I was imagining that the code could then be deduplicated a bit, for example as follows:

```js
let response = await fetch(urlAdvanced);
if (response.status !== 200) {
  response = await fetch(urlDirect);
  if (response.status !== 200) {
    return;
  }
}
const rawBytes = new Uint8Array(await response.arrayBuffer());
// ...
```

YarmoM:
That's a very neat solution! Here you go.

twiss:
Great, thanks :blush:

PR #1125, opened 2020-07-17
Prevent CORS errors from stopping code execution

YarmoM:
This PR prevents the fetch statements from stopping the execution of further code when a CORS error is thrown. The issue was that when a CORS related error prevented the advanced URL from fetching, it would also no longer try the direct URL.

### DIFF
--- a/src/wkd.js
+++ b/src/wkd.js
@@ -60,10 +60,16 @@ WKD.prototype.lookup = async function(options) {
   const urlAdvanced = `https://openpgpkey.${domain}/.well-known/openpgpkey/${domain}/hu/${localEncoded}`;
   const urlDirect = `https://${domain}/.well-known/openpgpkey/hu/${localEncoded}`;
 
-  let response = await fetch(urlAdvanced);
-  if (response.status !== 200) {
-    response = await fetch(urlDirect);
-    if (response.status !== 200) {
+  let response = await fetch(urlAdvanced)
+  .catch(function(err) {
+    return null;
+  });
+  if (!response || response.status !== 200) {
+    response = await fetch(urlDirect)
+    .catch(function(err) {
+      return null;
+    });
+    if (!response || response.status !== 200) {
       return;
     }
   }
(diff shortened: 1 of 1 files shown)


src/wkd.js
```diff
@@ -60,10 +60,16 @@ WKD.prototype.lookup = async function(options) {
...
+  });
+  if (!response || response.status !== 200) {
+    response = await fetch(urlDirect).catch(function(err) {
+      console.warn(err);
+      return null;
+    });
+    if (!response || response.status !== 200) {
       return;
     }
   }
```
twiss, line 75: Thanks for the PR!
Rather than returning null, I would prefer to still throw the error. Since callers have needed to try/catch for this already anyway, I think we could take this opportunity to also throw for 404s, without breaking compatibility too much.
```suggestion
  let response;
  try {
    response = await fetch(urlAdvanced);
    if (response.status !== 200) {
      throw new Error('Advanced WKD lookup failed: ' + response.statusText);
    }
  } catch (err) {
    util.print_debug_error(err);
    response = await fetch(urlDirect);
    if (response.status !== 200) {
      throw new Error('Direct WKD lookup failed: ' + response.statusText);
    }
  }
```
By the way, a full implementation of WKD should first lookup the policy file (https://tools.ietf.org/html/draft-koch-openpgp-webkey-service-09#section-4.5) for both the advanced and direct method, and decide based on that which method to use. That way, if the policy file for the advanced method returns 200, but an individual lookup returns 404, we _shouldn't_ fall back to the direct method (because we know that the advanced method is supported, but just the individual user doesn't exist), and just throw the 404 error immediately. If neither the advanced nor direct policy file exist, we shouldn't try to lookup at all, but throw immediately. This way, the browser can cache the policy files, and we don't have to keep hitting the server over and over when doing multiple lookups (because we already know that WKD isn't supported). But anyway, this doesn't have to be done in this PR.

YarmoM: Indeed, the policy file checking was a thing I was going to add in a PR soon but thanks for the clear outline on how to best handle the different situations. To be expected soon :)

twiss: Great, thanks :blush:

PR #1131, opened 2020-07-24
Add key.makeDummy()

larabr:
Add function to convert a decrypted private key into a gnu-dummy key

### DIFF
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -33,7 +33,6 @@ import util from '../util';
 import User from './user';
 import SubKey from './subkey';
 import * as helper from './helper';
-import S2K from '../type/s2k';
 
 /**
  * @class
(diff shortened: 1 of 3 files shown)


src/key/key.js
```diff
@@ -498,6 +500,30 @@ Key.prototype.clearPrivateParams = function () {
...
+    throw new Error("Key is not decrypted");
+  }
+  this.primaryKey.clearPrivateParams();
+  this.primaryKey.isEncrypted = false;
+  this.primaryKey.s2k = new S2K();
+  this.primaryKey.s2k.algorithm = 0;
+  this.primaryKey.s2k.c = 0;
+  this.primaryKey.s2k.type = 'gnu-dummy';
+};
+
```
twiss, line 526: I think this should be a method on `packet.SecretKey.prototype` (so that, in theory, it can also be used on subkeys), and be called `makeDummy`. From `Key.prototype.toDummy` I would expect to make a clone of the key, for consistency with `toPublic`, however we don't necessarily have to add that in this PR (also, it doesn't make clear that it only makes the primary key dummy, so maybe the name is still not ideal, cf. GNU `--export-secret-subkeys`).

Issue #1126, opened 2020-07-21
Session key decryption failed.


Chnovo28:
Session key decryption failed. from Version 4.10.5.
in V 4.10.4  working perfectly.

twiss:
Hey :wave: Do you have an example message and private key that you're unable to decrypt with? Or, if you're unable to post that, the public key that corresponds to that private key?

myrtleTree33:
Hi, another user here of your library.  I can confirm this is failing as of `v4.10.7`.  It works well in `v4.10.4`.

I am unable to disclose the keys, however traced the error to line #35703 in `openpgp.js`.

In `Message.decryptSessionKeys()`, there seems to be a condition that checks for the length of `keypackets`.  If the array is empty, condition falls through, leading to the default behavior of throwing an exception `Session key decryption failed`.

I think it may be looking at some public keys during decryption, when it should be just for private keys, though I am not sure if this is the case.

Can anyone help patch this? :P

gilach:
Same issue here. This needs to be addressed.

larabr:
Hi all,
in v4.10.5 we added some logic that verifies the binding signatures of decryption keys (https://github.com/openpgpjs/openpgpjs/blob/master/src/key/key.js#L351). This is needed to help guarantee the integrity of the keys.

Could you please check whether the decryption keys that you are using have been revoked? Or whether they are missing binding signatures?

twiss:
@gilach and @myrtleTree33, could you please post a message that has this issue, and the public key it was encrypted with? We would like to know whether the above PR will fix this before merging it, as we suspect there are still other issues.

vdabar:
Hi all, I have the exact same issue. I tracked down that it fails to meet this check:
https://github.com/openpgpjs/openpgpjs/blob/aa89893773c2ff97e85846ad5b0a184a89dc1688/src/key/helper.js#L356
which is called from here:
https://github.com/openpgpjs/openpgpjs/blob/aa89893773c2ff97e85846ad5b0a184a89dc1688/src/key/key.js#L336
and
where` signature.keyFlags[0] = 3 `
As I see it's only allowed to have `encrypt_communication` and  `encrypt_storage` flags, is there any reason behind it?

twiss:
[...]

PR #1135, opened 2020-07-30
Allow decryption with revoked keys

larabr:
Ignore revocation status when filtering keys for decryption.

See #1126

### DIFF
--- a/src/packet/public_key_encrypted_session_key.js
+++ b/src/packet/public_key_encrypted_session_key.js
@@ -130,8 +130,7 @@ PublicKeyEncryptedSessionKey.prototype.decrypt = async function (key) {
   const algo = enums.write(enums.publicKey, this.publicKeyAlgorithm);
   const expectedAlgo = enums.write(enums.publicKey, key.algorithm);
   // check that session key algo matches the secret key algo
-  // considering rsa_encrypt and rsa_encrypt_sign as equal
-  if ((algo > 2 || expectedAlgo > 2) && algo !== expectedAlgo) {
+  if (algo !== expectedAlgo) {
     throw new Error('Decryption error');
   }
   const decoded = await crypto.publicKeyDecrypt(algo, key.params, this.encrypted, key.getFingerprintBytes());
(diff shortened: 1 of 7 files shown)


test/general/openpgp.js
```diff
@@ -2428,6 +2428,41 @@ describe('OpenPGP.js public api tests', function() {
...
+        const encrypted = await openpgp.encrypt({
+          message: openpgp.message.fromText(plaintext),
+          publicKeys: pubKeyDE
+        });
+        privKeyDE.subKeys[0] = await privKeyDE.subKeys[0].revoke(privKeyDE.primaryKey);
+        const decOpt = {
+          message: await openpgp.message.readArmored(encrypted.data),
+          privateKeys: privKeyDE
+        };
+        await expect(openpgp.decrypt(decOpt)).to.not.be.rejected;
```
twiss, line 2444: This is a bit redundant with just calling the function. If you want to have an `expect` in there, I would write something like:
```suggestion
        const decrypted = await openpgp.decrypt(decOpt);
        expect(decrypted.data).to.equal(plaintext);
```

src/packet/public_key_encrypted_session_key.js
```diff
@@ -128,6 +128,12 @@ PublicKeyEncryptedSessionKey.prototype.encrypt = async function (key) {
  */
 PublicKeyEncryptedSessionKey.prototype.decrypt = async function (key) {
   const algo = enums.write(enums.publicKey, this.publicKeyAlgorithm);
+  const expectedAlgo = enums.write(enums.publicKey, key.algorithm);
+  // check that session key algo matches the secret key algo
+  // considering rsa_encrypt and rsa_encrypt_sign as equal
+  if ((algo > 2 || expectedAlgo > 2) && algo !== expectedAlgo) {
```
twiss, line 134: I don't think you have to consider rsa_encrypt and rsa_encrypt_sign as equal. ProtonMail/crypto has this: https://github.com/ProtonMail/crypto/blob/1fa7f403fb9c3aa7b47443e1e3b39c80d6eedcb6/openpgp/packet/encrypted_key.go#L91-L93 and it hasn't caused any issues that I know of.

larabr: We have two tests that fail without this check - I think older keys might give issues in this sense

twiss: Maybe the tests should be fixed instead?
https://github.com/openpgpjs/openpgpjs/search?q=enc.publicKeyAlgorithm+%3D+%27rsa_encrypt%27

Issue #1136, opened 2020-08-04
Unknown notations with the "critical" bit set SHOULD invalidate the signature


teythoon:
Unknown notations with the "critical" bit set SHOULD invalidate
the signature.  This ensures that we can introduce notations that
constrain the validity of the signature, and have implementations
that do not yet support the new constraint err on the side of
caution and reject the signature rather than accepting it
disobeying the constraint.

https://tools.ietf.org/html/rfc4880#section-5.2.3.16
https://tests.sequoia-pgp.org/#Detached_signature_with_Subpackets

Reproducer:

The following signature over the data `huhu` contains an unknown
notation with the critical bit set:

```
-----BEGIN PGP SIGNATURE-----

wsEfBAABCABJBYJfKDH0K5QAAAAAAB0ABXVua25vd25AdGVzdHMuc2VxdW9pYS1w
Z3Aub3JndmFsdWUWIQTRpm4aI7GCyZgPeIz7/MgqAV5zMAAKCRD7/MgqAV5zMLBK
C/9Vdte8yq8QFp7Bo3mWWXiUDJ2vOdEpJs9PB12hN97wewjFKWvKcWRHofBrK+0v
P4/sixMjqSXB75iz/lzCbPjkmgmaGKN6Smq6Pj5kIdpWIdtfjNoa+OEn1emrxpY1
dT0Pe7r8w3Y/JmfK1ZoFfiq5YBP+rsjmAXf6BQ9RH+kMUsOgOVCjqzI5vPibTD91
Aefm7XS35Wlj/O/zFXR9tiZyE/dKIujohB3bsV6sAmYzPxZVuZCYaj+6afRlAuWg
IYs+aFmldMg4p1t1Ab/bRHbBxIlzhtbNE6IyOfc17mgOcjQzVJBc/EaxD7S3KjU2
4aFmOaLyWDrvAJu1Ror2b+zekEiq90CnCtAUiAq6/qbWlGCGX4zFKWyHousDJcEB
1n0xDkFUhndisMZanaU3S1hwzawMnpHi1RHI2a/etPPk0Ryrzo2rFi8ZdAHLWY6o
4ymAWpppg47NhCwuYh/EG5G1P4ccxksRj1GX2r5HxH8ZeAQWWdviINErgrUONWHy
bwM=
=0x2S
-----END PGP SIGNATURE-----
```

```
-----BEGIN PGP PUBLIC KEY BLOCK-----
[...]

PR #1140, opened 2020-08-13
signature: Add test case for notations

wiktor-k:
This is a tests case for a bug reported in issue #1136:
non-human-readable notations with critical bit set should trigger an
error when encountered but they did not.

This test will fail until #983 is merged that adds proper support for
non-human-readable notations and fixes the underlying issue.


twiss:
:+1: Thanks!

PR #1141, opened 2020-08-13
[v5] Store named key params

### LINKED ISSUE #1130: Improve MPI parameter parsing
MPIs do not always encode parameters properly, as they drop leading zeros.  In particular, this gives issues with curve p521, which ends up being one byte shorter than expected by webcrypto whenever the leading bit is zero.
With other ciphers and curves, the length mismatch is extremely rare, but we should always ensure that the parameters have the expected length. 

As a temporary solution, `parseParams` is called before using the ECC parameters (e.g. https://github.com/openpgpjs/openpgpjs/blob/00c5f386894098dad7acf93c5c58937a53ffbfc9/src/crypto/signature.js#L118). However, it would be best to pad the parameters when reading/parsing the keys, rather than when using them.
Another advantage of doing this, is that inside the key we could already store the parameters by name and as Uint8Arrays rather than MPIs.

See https://github.com/openpgpjs/openpgpjs/pull/1116#discussion_r447772899 for additional context and ideas.

larabr:
Fix #1130:
- do not keep params in MPI form, but convert them to Uint8Array when generating/parsing the key
- store params by name in an object, instead of using an array'
- uniform low-level crypto functions to accept and always return Uint8Arrays instead of BigIntegers

### DIFF
--- a/src/crypto/crypto.js
+++ b/src/crypto/crypto.js
@@ -77,14 +77,11 @@ export default {
       }
       case enums.publicKey.elgamal: {
         const { p, g, y } = publicParams;
-        data = new type_mpi(await pkcs1.eme.encode(data, p.length));
-        const m = data.toUint8Array();// TODO i would do this as part of encryption
-        const res = await publicKey.elgamal.encrypt(m, p, g, y);
+        const res = await publicKey.elgamal.encrypt(data, p, g, y);
         return constructParams(types, [res.c1, res.c2]);
       }
       case enums.publicKey.ecdh: {
         const { oid, Q, kdfParams } = publicParams;
-        data = new type_mpi(pkcs5.encode(data));
         const { publicKey: V, wrappedKey: C } = await publicKey.elliptic.ecdh.encrypt(
           oid, kdfParams, data, Q, fingerprint);
         return constructParams(types, [V, C]);
@@ -103,8 +100,8 @@ export default {
    * @param {Array<module:type/mpi|
                    module:type/ecdh_symkey>}
                                             data_params encrypted session key parameters
-   * @param {String}                        fingerprint Recipient fingerprint
-   * @returns {String}                          String containing the decrypted data
+   * @param {Uint8Array}                    fingerprint Recipient fingerprint
+   * @returns {Uint8Array}                  decrypted data
    * @async
    */
   publicKeyDecrypt: async function(algo, publicKeyParams, privateKeyParams, data_params, fingerprint) {
@@ -121,17 +118,15 @@ export default {
         const c2 = data_params[1].toUint8Array();
         const p = publicKeyParams.p;
         const x = privateKeyParams.x;
-        const result = new type_mpi(await publicKey.elgamal.decrypt(c1, c2, p, x));
-        return pkcs1.eme.decode(result.toUint8Array('be', p.byteLength())); // TODO do as part of crypto functions
+        return publicKey.elgamal.decrypt(c1, c2, p, x);
       }
       case enums.publicKey.ecdh: {
         const { oid, Q, kdfParams } = publicKeyParams;
         const { d } = privateKeyParams;
         const V = data_params[0].toUint8Array();
         const C = data_params[1].data;
-        const result = new type_mpi(await publicKey.elliptic.ecdh.decrypt(
-          oid, kdfParams, V, C, Q, d, fingerprint));
-        return pkcs5.decode(result.toUint8Array());
+        return publicKey.elliptic.ecdh.decrypt(
+          oid, kdfParams, V, C, Q, d, fingerprint);
       }
       default:
         throw new Error('Invalid public key encryption algorithm.');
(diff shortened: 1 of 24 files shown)


src/crypto/crypto.js
```diff
@@ -152,83 +151,141 @@ export default {
...
+      default:
+        throw new Error('Invalid public key encryption algorithm.');
+    }
+  },
+
+  /**
+   * Parse private key material in binary form to get the key parameters
+   * @param {module:enums.publicKey} algo  The key algorithm
+   * @param {Uint8Array}             bytes The key material to parse
+   * @param {Object}                 publicParams (ECC only) public params, needed to format some private params
```
larabr, line 222: Maybe we should also pad RSA and DSA keys out of principle -  I don't think it's possible that they have a leading 0 byte, see [this study on RSA](https://www.usenix.org/system/files/conference/usenixsecurity16/sec16_paper_svenda.pdf)

twiss: Yeah, the way we generate them they have at most a leading 0 bit I think, not a leading 0 byte. Regarding padding, I don't think there are many restrictions on the length of these parameters - I guess usually `p` and `q` are `n.length / 2`, but I don't think this is really required.

src/crypto/crypto.js
```diff
@@ -264,33 +372,50 @@ export default {
...
       case enums.publicKey.rsaEncryptSign:
       case enums.publicKey.rsaSign: {
         return publicKey.rsa.generate(bits, "10001").then(function(keyObject) {
-          return constructParams(
-            types, [keyObject.n, keyObject.e, keyObject.d, keyObject.p, keyObject.q, keyObject.u]
-          );
+          const { n, e, d, p, q, u } = keyObject;
+          const params = { n, e, d, p, q, u };
+          paramNames.forEach(name => { params[name] = params[name].toArrayLike(Uint8Array); });
+          return params;
```
twiss, line 384: I would modify `publicKey.rsa.generate` to return `Uint8Array`s instead, and then make this a one-liner function call, if possible. And similarly below.

src/crypto/crypto.js
```diff
@@ -152,83 +151,141 @@ export default {
...
+
+  /**
+   * Convert public params to MPI and serializes them in the proper order
+   * @param {module:enums.publicKey}  algo The public key algorithm
+   * @param {Object}                  params The private key parameters indexed by name
+   * @returns {Uint8Array}  The array containing the MPIs
+   */
+  serializePublicKeyParams: function(algo, params) {
+    const paramNames = this.getPublicKeyParamNames(algo);
+    return serializeKeyParams(params, paramNames);
```
twiss, line 286: Rather than abstracting this away, I would make an explicit switch statement here, with something like:
```js
switch (algo) {
  case rsa:
    const { n, e } = params;
    return util.concatUint8Array([n, e].map(util.uint8ArrayToMpi));
  // ...
}
```
I know it's longer, but then you can get rid of `getPublicKeyParamNames`, and it's more flexible for symmetric keys, for example (they don't have to be MPIs).
If we want to be a bit adventurous, you could also do:
```js
switch (algo) {
  case rsa:
  case dsa:
  case elgamal:
    return util.concatUint8Array(Object.values(params).map(util.uint8ArrayToMpi));
  // ...
}
```
Since Object order is nowadays defined in the ECMAscript spec. Even in old browsers, this should be perfectly safe if we have no numeric keys etc.

src/crypto/crypto.js
```diff
@@ -152,83 +151,141 @@ export default {
...
-        return [type_mpi, type_mpi, type_mpi, type_mpi];
+      case enums.publicKey.elgamal: {
+        const params = {};
+        const paramNames = this.getPublicKeyParamNames(algo);
+        paramNames.forEach(name => {
+          const param = util.readMPI(bytes.subarray(read)); read += param.length + 2;
+          params[name] = param;
+        });
+        return { read: read, params };
+      }
```
twiss, line 188: And then here, I would split up the cases and parse each param explicitly; I know it's more repetitive but then it's more consistent with the code below, and once we switch to streaming (and can have a `parseMPI` function that consumes the stream), it won't be too bad.
(And similarly for the private params.)

src/packet/packetlist.js
```diff
@@ -134,6 +134,7 @@ class PacketList extends Array {
     packet.packets = packet.packets || new PacketList();
 
     super.push(packet);
+    return this;
```
twiss, line 137: Eventually, I think we should remove this `push` function and just default to the normal `Array.prototype.push` behavior. Since that one returns the new length; I wouldn't `return this` here, as it's inconsistent with that behavior.

src/crypto/crypto.js
larabr: The ECC generation functions return BN as well, I would keep that consistent, with BigInteger the conversion to Uint8Array will look nicer anyway

PR #1148, opened 2020-08-26
Add config.allow_insecure_decryption_with_signing_keys

larabr:
Follow up to #1116 , #1126 : add option to decrypt messages using a sign-only RSA key.
This is needed as openpgpjs < v0.4.0 did not properly check the key flags when selecting an encryption key, and could end up using a sign-only key. (See fix: https://github.com/openpgpjs/openpgpjs/commit/57849437d9af82b93c8fc55ddb2c98f65c36c286)

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -121,11 +121,11 @@ export default {
    */
   revocations_expire:       false,
   /**
+   * Allow decryption using RSA keys without `encrypt` flag.
+   * This setting is potentially insecure, but it is needed to get around an old openpgpjs bug
+   * where key flags were ignored when selecting a key for encryption.
    * @memberof module:config
    * @property {Boolean} allow_insecure_decryption_with_signing_keys
-   *  If true, RSA sign-only keys can be used for decryption
-   *  This is potentially insecure, but it is needed to get around an old openpgpjs bug
-   *  where sign-only keys were selected for encryption
    */
   allow_insecure_decryption_with_signing_keys: false,
 
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -120,6 +120,14 @@ export default {
    * @property {Boolean} revocations_expire If true, expired revocation signatures are ignored
    */
   revocations_expire:       false,
+  /**
+   * @memberof module:config
+   * @property {Boolean} allow_insecure_decryption_with_signing_keys
+   *  If true, RSA sign-only keys can be used for decryption
+   *  This is potentially insecure, but it is needed to get around an old openpgpjs bug
+   *  where sign-only keys were selected for encryption
+   */
+  allow_insecure_decryption_with_signing_keys: false,
 
   /**
    * @memberof module:config
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -367,7 +367,7 @@ export function isValidDecryptionKeyPacket(signature) {
     throw new Error('Signature not verified');
   }
 
-  if (config.allow_insecure_decryption_with_signing_keys === true) {
+  if (config.allow_insecure_decryption_with_signing_keys) {
     // This is only relevant for RSA keys, all other signing ciphers cannot decrypt
     return true;
   }
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -366,6 +366,12 @@ export function isValidDecryptionKeyPacket(signature) {
   if (!signature.verified) { // Sanity check
     throw new Error('Signature not verified');
   }
+
+  if (config.allow_insecure_decryption_with_signing_keys === true) {
+    // This is only relevant for RSA keys, all other signing ciphers cannot decrypt
+    return true;
+  }
+
   return !signature.keyFlags ||
     (signature.keyFlags[0] & enums.keyFlags.encrypt_communication) !== 0 ||
     (signature.keyFlags[0] & enums.keyFlags.encrypt_storage) !== 0;
--- a/test/general/key.js
+++ b/test/general/key.js
@@ -1916,6 +1916,57 @@ vqBGKJzmO5q3cECw
 =X9kJ
 -----END PGP PRIVATE KEY BLOCK-----`;
 
+
+const rsaSignOnly = `-----BEGIN PGP PRIVATE KEY BLOCK-----
+Comment: different
+
+xcLYBF9Gl+MBCACc09O3gjyO0B1ledGxGFSUpPmhhJzkxKoY1WDX8VlASCHz
+bAA/BytgYBXHTe7N+N3yJ6uiN3DIQ2j5uGWk/h5jyIOsRuzQxJ40n8AdK/71
+SGDCG1X5l1h9vmVTJxkQ3pcOxqRg55EEuJWKN1v7B1hIPxhaM5hgH/7s+PNn
+lQddckQJqYkpm9Hy6EI7f9oHrOtWJWZoCHkWZVld7+9ZVPi34ex5ofYOuvNL
+AIKZCc7lAiUiDJYQ+hIJRoYwLYhjIshpYoHgNeG4snlupNO32BiwDbHFDjeu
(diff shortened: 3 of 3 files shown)


src/config/config.js
```diff
@@ -120,6 +120,14 @@ export default {
    * @property {Boolean} revocations_expire If true, expired revocation signatures are ignored
    */
   revocations_expire:       false,
+  /**
+   * @memberof module:config
+   * @property {Boolean} allow_insecure_decryption_with_signing_keys
+   *  If true, RSA sign-only keys can be used for decryption
```
twiss, line 126:
```suggestion
   *  If true, RSA signing keys can be used for decryption
```
This is very nitpicky, but otherwise people might think you're talking about the "RSA Sign-Only" algorithm (ID 3) as opposed to just RSA keys in general that have `keyFlags = sign`.
Also, even more nitpicky, but please change the double spaces at the beginning to single spaces and add some dots at the end :upside_down_face:

src/key/helper.js
```diff
@@ -366,6 +366,12 @@ export function isValidDecryptionKeyPacket(signature) {
   if (!signature.verified) { // Sanity check
     throw new Error('Signature not verified');
   }
+
+  if (config.allow_insecure_decryption_with_signing_keys === true) {
```
twiss, line 370:
```suggestion
  if (config.allow_insecure_decryption_with_signing_keys) {
```

test/general/key.js
```diff
@@ -2667,6 +2718,26 @@ describe('Key', function() {
     await expect(pubKey.getEncryptionKey()).to.be.rejectedWith('Could not find valid encryption key packet in key c076e634d32b498d');
   });
 
+  it('should not decrypt using a sign-only RSA key, unless explicitly configured', async function () {
+    const allowSigningKeyDecryption = openpgp.config.allow_insecure_decryption_with_signing_keys;
+    const { keys: [key] } = await openpgp.key.readArmored(rsaSignOnly);
+    console.log('off');
```
twiss, line 2724:
```suggestion
```

test/general/key.js
```diff
@@ -2667,6 +2718,26 @@ describe('Key', function() {
...
+    openpgp.config.allow_insecure_decryption_with_signing_keys = false;
+    await expect(openpgp.decrypt({
+      message: await openpgp.message.readArmored(encryptedRsaSignOnly),
+      privateKeys: key
+    })).to.be.rejectedWith(/Session key decryption failed/);
+    console.log('on');
```
twiss, line 2730:
```suggestion
```

test/general/key.js
```diff
@@ -2667,6 +2718,26 @@ describe('Key', function() {
...
+
+    openpgp.config.allow_insecure_decryption_with_signing_keys = true;
+    await expect(openpgp.decrypt({
+      message: await openpgp.message.readArmored(encryptedRsaSignOnly),
+      privateKeys: key
+    })).to.be.fulfilled;
+
+    openpgp.config.allow_insecure_decryption_with_signing_keys = allowSigningKeyDecryption;
```
twiss, line 2738: Please put the rest of this test in a `try` with this line in a `finally`, so that we don't mess up the config option if the test fails

test/general/key.js
```diff
@@ -1916,6 +1916,57 @@ vqBGKJzmO5q3cECw
 =X9kJ
 -----END PGP PRIVATE KEY BLOCK-----`;
 
+
+const rsaSignOnly = `-----BEGIN PGP PRIVATE KEY BLOCK-----
+Comment: different
```
twiss, line 1921: What is the meaning of these comments? :D

src/config/config.js
larabr: But signing keys are not necessarily sign-only, you can have both sign + encrypt flags.

test/general/key.js
larabr: They were automatically generated by openpgpjs, not sure why... I just encrypted the message and it ended up like this, I have seen it before when using keys that were somewhat "off", but haven't looked into it too much.

twiss: I see. I guess it's coming from this: https://github.com/openpgpjs/openpgpjs/blob/cc1bdcbae8d9f7d392d9da3fbc5ded9c42f62302/test/general/openpgp.js#L741
Maybe this is an argument for why we should undo configs in tests :upside_down_face: 
Anyway, please just remove the comments, they're a bit confusing

src/config/config.js
twiss: I see. But I guess in the same vein, they could also be authentication subkeys, or have keyFlags = 0... so maybe it's best to just say "non-encryption keys" or "keys without encryption key flags"

test/general/key.js
wiktor-k: > Maybe this is an argument for why we should undo configs in tests upside_down_face

Another idea would be to avoid global config object altogether and provide it explicitly during initialization (`const openpgp = require('openpgp')(config);`). That'd of course be quite a big change but worth mentioning as each test would have a clean slate by design :)

twiss: Yeah. Even more than that, sometimes you want to set a config option for one message/decryption only, which is a bit difficult right now with the async model. So I think it would be even nicer to add `config` options/parameters to all of the top-level functions. Then you could also do:
```js
try {
  result = openpgp.decrypt({ message });
} catch (e) {
  warn("Insecure!");
  result = openpgp.decrypt({ message, config: { allow_insecure_decryption_with_signing_keys: true } });
}
```
Or make it depend on the date of the email, et cetera.
And then there would also be no statefulness in the tests, potentially allowing us to run them in parallel, etc.

wiktor-k: Great idea! This reminds me of the [policy system](https://docs.sequoia-pgp.org/sequoia_openpgp/policy/index.html) we use at Sequoia. Running test in parallel would be a nice side-effect too...

For top level functions this could be hacked together with a tiny try-finally around the actual code:

```js
function decrypt(..., config) {
  const oldConfig = openpgp.config;
  openpgp.config = Object.assign({}, openpgp.config, config);
  try {
    // actual code...
  } finally {
    openpgp.config = oldConfig;
  }
}
```

twiss: > For top level functions this could be hacked together with a tiny try-finally around the actual code:

This wouldn't work if you're running multiple encryptions/decryptions in parallel, since another call might overwrite the config before it's finished. So we'd probably have to pass these config parameters throughout the codebase, unfortunately.

wiktor-k: Good point. Thanks for taking the time to discuss internals @twiss!

alan19031216:
Error: Error decrypting message: Session key decryption failed.
(node:23480) UnhandledPromiseRejectionWarning: Error: Error decrypting message: Session key decryption failed.
    at Message.decryptSessionKeys (C:\Users\User\OneDrive\Documents\millenning\node_modules\openpgp\dist\openpgp.js:36311:22)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Message.decrypt (C:\Users\User\OneDrive\Documents\millenning\node_modules\openpgp\dist\openpgp.js:36177:35)
    at async decryptFunction (C:\Users\User\OneDrive\Documents\millenning\routes\API\DBS-api\callback.js:197:23)
(node:23480) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:23480) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

PR #1151, opened 2020-09-05
Fix the inconsistency between the document and the code in hkp.js

chenlhlinux:


twiss:
@chenlhlinux Thanks for your PR!

PR #1156, opened 2020-09-16
Fix minor issues like jsdoc

danrr:


src/crypto/crypto.js
```diff
@@ -92,34 +92,33 @@ export async function publicKeyEncrypt(algo, publicParams, data, fingerprint) {
...
  * @param {Array<module:type/mpi|
-                 module:type/ecdh_symkey>}
-                                          data_params encrypted session key parameters
+                 module:type/ecdh_symkey>}  data_params encrypted session key parameters
  * @param {Uint8Array}                    fingerprint Recipient fingerprint
  * @returns {Uint8Array}                  decrypted data
  * @async
  */
-export async function publicKeyDecrypt(algo, publicKeyParams, privateKeyParams, data_params, fingerprint) {
+export async function publicKeyDecrypt(algo, publicParams, privateParams, data_params, fingerprint) {
```
twiss, line 103: I kind of like having `publicKeyParams` and `privateKeyParams` here because we also have `data_params` (which should probably really be called `messageParams` or `sessionKeyParams`) - those are also public, so it helps to disambiguate `publicKeyParams` I think. If you want to make it consistent with the other function, I would rename the param above to `publicKeyParams`, as well.

src/crypto/crypto.js
```diff
@@ -56,7 +56,7 @@ export function constructParams(types, data) {
  * Encrypts data using specified algorithm and public key parameters.
  * See {@link https://tools.ietf.org/html/rfc4880#section-9.1|RFC 4880 9.1} for public key algorithms.
  * @param {module:enums.publicKey}    algo        Public key algorithm
- * @param {Object}                    pubParams  Algorithm-specific public key parameters
+ * @param {Object}                    publicParams  Algorithm-specific public key parameters
```
twiss, line 59: This is very minor, but since that's the name of this PR: please update the spacing of the surrounding lines so that they line up (even though I know it was already kind of crooked :see_no_evil:). And similarly in the next function :blush:

danrr: I lined up the jsdoc in the whole file

src/crypto/crypto.js
danrr: Makes sense, I'll leave this for now

PR #1157, opened 2020-09-17
Named session key params

danrr:
Changed session key parameter handling to mirror key parameters.
Parameters are stored as an object rather than an array. MPIs are
always stored as UInt8Arrays

### DIFF
--- a/src/crypto/crypto.js
+++ b/src/crypto/crypto.js
@@ -25,7 +25,6 @@
  * @requires crypto/random
  * @requires type/ecdh_symkey
  * @requires type/kdf_params
- * @requires type/mpi
  * @requires type/oid
  * @requires enums
  * @requires util
@@ -35,22 +34,12 @@
 import publicKey from './public_key';
 import * as cipher from './cipher';
 import { getRandomBytes } from './random';
-import type_ecdh_symkey from '../type/ecdh_symkey';
+import ECDHSymkey from '../type/ecdh_symkey';
 import KDFParams from '../type/kdf_params';
-import type_mpi from '../type/mpi';
 import enums from '../enums';
 import util from '../util';
 import OID from '../type/oid';
 import { Curve } from './public_key/elliptic/curves';
-
-export function constructParams(types, data) {
-  return types.map(function(type, i) {
-    if (data && data[i]) {
-      return new type(data[i]);
-    }
-    return new type();
-  });
-}
 
 /**
  * Encrypts data using specified algorithm and public key parameters.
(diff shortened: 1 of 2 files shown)


src/crypto/crypto.js
```diff
@@ -94,35 +81,32 @@ export async function publicKeyEncrypt(algo, publicParams, data, fingerprint) {
...
       const p = publicKeyParams.p;
       const x = privateKeyParams.x;
       return publicKey.elgamal.decrypt(c1, c2, p, x);
     }
     case enums.publicKey.ecdh: {
       const { oid, Q, kdfParams } = publicKeyParams;
       const { d } = privateKeyParams;
-      const V = data_params[0].toUint8Array();
-      const C = data_params[1].data;
+      const { V, C } = sessionKeyParams;
```
danrr, line 107: to make this very confusing, these two lines could be:
```
      const { V, C: { data: C } } = sessionKeyParams;
      return publicKey.elliptic.ecdh.decrypt(
        oid, kdfParams, V, C.data, Q, d, fingerprint);
```

src/crypto/crypto.js
```diff
@@ -35,22 +34,13 @@
...
 import { Curve } from './public_key/elliptic/curves';
 
-export function constructParams(types, data) {
-  return types.map(function(type, i) {
-    if (data && data[i]) {
-      return new type(data[i]);
-    }
-    return new type();
-  });
-}
```
twiss, line 44:
```suggestion
```

src/crypto/crypto.js
```diff
@@ -94,35 +81,32 @@ export async function publicKeyEncrypt(algo, publicParams, data, fingerprint) {
...
-      const c = data_params[0].toUint8Array();
+      const { c } = sessionKeyParams;
       const { n, e } = publicKeyParams;
       const { d, p, q, u } = privateKeyParams;
       return publicKey.rsa.decrypt(c, n, e, d, p, q, u);
     }
     case enums.publicKey.elgamal: {
-      const c1 = data_params[0].toUint8Array();
-      const c2 = data_params[1].toUint8Array();
+      const { c1,c2 } = sessionKeyParams;
```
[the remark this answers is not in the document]
danrr: Fixed

src/crypto/crypto.js
danrr: Fixed

src/crypto/crypto.js
twiss: I like it better as it is. Alternatively, you could store it as a `Uint8Array` directly, and get rid of the `ECDHSymkey` type entirely.

danrr: ECDHSymkey needs a prefix of the length and the type abstracts that away with reading and writing.

```
write() {
    return util.concatUint8Array([new Uint8Array([this.data.length]), this.data]);
  }
```

I could parse the length in this function and add a special case to `serializeKeyParams` for ECDH. It currently treats all Uint8Array as MPI and uses the type's write method otherwise.

```
orderedParams = Object.keys(params).map(name => {
        const param = params[name];
        return util.isUint8Array(param) ? util.uint8ArrayToMpi(param) : param.write();
      });
```

PR #1158, opened 2020-09-21
Named signature params

danrr:


src/crypto/crypto.js
```diff
@@ -247,7 +247,7 @@ export function parseEncSessionKeyParams(algo, bytes) {
  * @param {Object}                  params  The key parameters indexed by name
  * @returns {Uint8Array}                    The array containing the MPIs
  */
-export function serializeKeyParams(algo, params) {
+export function serializeAlgorithmSpecificFields(algo, params) {
```
twiss, line 250: Perhaps `serializeParams`? Bit less wordy.

src/packet/signature.js
```diff
@@ -133,15 +131,27 @@ class SignaturePacket {
     this.signedHashValue = bytes.subarray(i, i + 2);
     i += 2;
 
-    this.signature = bytes.subarray(i, bytes.length);
+    this.signature = crypto.signature.parseSignatureParams(this.publicKeyAlgorithm, bytes.subarray(i, bytes.length));
```
twiss, line 134: I would rename this property to `this.params`, if it's not too much work. `signature.signature` is a bit confusing (even though I know it was already the case).

src/packet/signature.js
```diff
@@ -133,15 +131,27 @@ class SignaturePacket {
...
+  }
+
+  /**
+   * @returns {Uint8Array | ReadableStream<Uint8Array>}
+   */
+  write_signature() {
+    if (util.isPromise(this.signature)) {
```
twiss, line 141: Perhaps `this.params instanceof Promise`? Then you don't need to add a helper function just for this.

danrr: This is a corner case but with some libraries/polyfill promises aren't `Promise`s. Checking if something is a `thenable` is less prone to errors even when non-standard Promises exist.

src/crypto/crypto.js
danrr: I was thinking about using `params`, I just thought this was more precise. Although, the objects have `param` keys rather than `AlgorithmSpecificFields` keys, so yeah, `params` would be consistent with that...

Changed it.

src/packet/signature.js
danrr: done

src/crypto/signature.js
```diff
@@ -9,49 +9,89 @@
 import publicKey from './public_key';
 import enums from '../enums';
 import util from '../util';
+import crypto from '.';
```
twiss, line 12: Do you need this import?

danrr: nope, left over from an intermediate commit

src/packet/signature.js
twiss: Yeah, but.. this Promise is coming from calling an `async` function without `await`, so it can only be a built-in Promise. If someone transpiles the code for an environment where Promises don't exist, then `Promise` should also point to the polyfilled `Promise` implementation.

danrr: Removed the util.

There's a non-0 percent chance of this happening though: https://xkcd.com/1172/ :D

PR #1160, opened 2020-09-22
Re-enable linting

danrr:


.travis.yml
```diff
@@ -1,13 +1,12 @@
-sudo: false
```
danrr, line 1: The sudo keyword was deprecated and the build works fine without it. 

https://changelog.travis-ci.com/linux-builds-run-on-vms-by-default-77106

.travis.yml
```diff
@@ -1,13 +1,12 @@
...
 language: node_js
-cache:
-  directories:
-    - node_modules
+cache: npm
```
danrr, line 2: Travis has smart npm caching which adapts to the commands that are used in the pipeline; it appears it is now the default so the cache declaration could even be removed without any downside. 

https://docs.travis-ci.com/user/caching/#npm-cache

Issue #561, opened 2017-07-13
Get TypeScript type definitions up to date and working again


LucasHill:
Sorry if this is not desired here or just the wrong place, but I wanted to bring up an issue over at the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) project. Openpgpjs has type definitions there but they are [not working](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/16573). I am using openpgpjs in my own TypeScript project and not having the type definitions is a bummer. I would be willing to look into it myself but I am pretty new at TypeScript. 

The [current implementation](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/openpgp/index.d.ts) is also not in the correct format I believe as there are no exports from the index.d.ts.

bartbutler:
Hi Lucas,

Unfortunately, I also have very little experience with Typescript. If you'd like to submit a fix for this I'd be happy to merge it.

CB-Giepa:
I also need the updated Typings for the libary, so I am currently putting some work into that.
Unfortunately I have some trouble to understand the Documentation. The return values of the functions are for me the most tricky part.

For example the function encrypt(...) in the openpgp module returns a "encrypted (and optional signed message)". So that is another message object than from the function encryptSessionKey?

Hopefully someone can help me to handle with this.
Greetings.

bartbutler:
Hi @CB-Giepa,

So you should think of an OpenPGP message as sort of a container. Most encrypted messages will have an encrypted data packet and one or more key packets for the recipients. encryptSessionKey's Message object _just_ has key packets. The data packet is then added by the rest of 'encrypt' function. encryptSessionKey is exposed as well as encrypt because it can be useful to manipulate the (small) key packets separately from the (large) data packet.

ffflorian:
Hi @CB-Giepa,
is there any update on the progress? I'd love to use OpenPGP in TypeScript :slightly_smiling_face:. Dow you need any help?

BrunoScheufler:
Sadly this issue seems a bit abandoned, is there any progress of creating updated type definitions?

bartbutler:
It would be great to have someone from the community take this on now that OpenPGP.js 3 is out. None of the core devs have much experience with TypeScript.

tomholub:
On behalf of @FlowCrypt I'll send a $100 bounty to the author of the first accepted PR.

errietta:
I got mine merged in!
[...]

PR #1163, opened 2020-09-25
remove asn1 from external deps

### LINKED ISSUE #925: Error: Cannot find module 'openpgp'
Error
---
```
Error: Cannot find module 'openpgp'
     at Function.Module._resolveFilename (module.js:548:15)
     at Function.Module._load (module.js:475:25)
     at Module.require (module.js:597:17)
     at require (internal/module.js:11:18)
     at Module._compile (module.js:653:30)
     at Object.Module._extensions..js (module.js:664:10)
     at Module.load (module.js:566:32)
```

Problem
---
Hello guys,
I'm using babel to transpile my code.
The problem is that openpgp doesn't work when the code is transpiled.

Workaround
---
I found a workaround. I've created a node_modules folder inside the build folder and I've copied these packages inside it:
- node_modules/openpgp
- node_modules/asn1.js
- node_modules/bn.js
- node_modules/minimalistic-assert

With this solution, I can use openpgp.

Can you check that.

Note
---
I've tested without babel and openpgp is working without any issues.
Unfortunately, I've need babel.

Information
---
**OS:** Linux & macOS
**Node version:** 8.16.0 - 10.16.0
**Babel version:** 6.26.0 (old one)

vordimous:
this could resolve #925,

but does resolve [this issue](https://github.com/vercel/ncc/issues/538)


twiss:
Hi :wave: Thanks for the PR.

So - a few points:

- asn1.js is only required in Node.js, not in the browser. Therefore, in OpenPGP.js v4, I would prefer not to add it to the bundle.
- However, in OpenPGP.js v5 (https://github.com/openpgpjs/openpgpjs/pull/1047), we will have a separate bundle for Node.js and for the browser. So there, I would be more open to adding it to the Node.js bundle.
- Even there, however, it would seem to me (but I might be wrong) that the same problem exists for the other dependencies of OpenPGP.js, it's just that they are less noticable. For example, if you call `new openpgp.Keyring()`, I guess it would throw since the `node-localstorage` dependency is missing?
- If that's the case, it would seem a bit suboptimal to me to include all the dependencies in the build, since some of them are sometimes not used.
- Since these dependencies are clearly listed in `package.json`, it seems to me that there's some other issue, either in the build process of openpgp.js, or ghaction-import-gpg, or in ncc, and I feel like this should be fixed differently.
- If you want, you could check whether this issue has been resolved in OpenPGP.js v5, by pointing package.json to `"openpgp": "github:openpgpjs/openpgpjs#74f16f080c9b5a3c9b0a51ffdd869a7d07a797fe"`. (There are however also breaking API changes in v5, see https://github.com/openpgpjs/openpgpjs/wiki/V5-Changelog#Examples.)

vordimous:
It makes sense this conflict would exist when trying to build for node and browser support. I will try to point to V5 and hope that works. 

As far as ncc goes, I would think that the import strategy in general needs to be specific enough for it to know what libs in needs to compile into the single binary. I did try to alter the code a bit to see if that is what was causing ncc not to see "asn1.js" as a dep to include. 

I will report back after trying V5.

vordimous:
Ok @twiss, I can confirm that V5 works to resolve the dependency issue. I was able to package and deploy the ncc compiled node script without having any module references missing.

Do you have a timeline for V5, specifically when the typscript type definitions will be available, or at least a `.d.ts` file?

twiss:
Thanks for checking!

We don't maintain the TypeScript bindings. The types at [@types/openpgp](https://www.npmjs.com/package/@types/openpgp) are maintained by external contributors; see https://github.com/openpgpjs/openpgpjs/issues/561. We would be open to a PR to the v5 branch that adds types, however.

PR #1172, opened 2020-11-05
Fix SecretKey.prototype.makeDummy()

larabr:
Keys converted using `key.makeDummy()` were not serialised correctly as they were treated as unencrypted keys.


tomholub:
I'm not a maintainer here but still I'd like to thank you for these contributions lately :+1:

PR #1175, opened 2020-11-10
[v5] Unexport openpgp.crypto, openpgp.util, and low-level types

twiss:
These exports don't have much to do with OpenPGP, and don't make much sense in a library called OpenPGP.js. These exports also thwart tree-shaking, as they don't allow granular imports - e.g. to import AES, you'd need to import the whole crypto "namespace". If absolutely required, you can `import aes from "openpgp/src/crypto/cipher/aes"` instead, for example (e.g. with [esm](https://www.npmjs.com/package/esm) on Node.js). And finally, removing these imports will allow us to change the internal API more easily, and switch between low-level crypto libraries without breaking anything, etc.


test/crypto/elliptic.js
```diff
@@ -1,4 +1,8 @@
 const openpgp = typeof window !== 'undefined' && window.openpgp ? window.openpgp : require('../..');
+const elliptic_curves = require('../../src/crypto/public_key/elliptic');
+const hashMod = require('../../src/crypto/hash');
+const config = require('../../src/config');
```
larabr, line 4: Why not access opengpg.config?

twiss: Because I'm modifying the config below, and I want it to have effect for the `elliptic_curves` functions.
`require('../../src')` returns a different instance of openpgp than `require('../..')`.

drzraf:
Although I do understand the underlying reasons, I'm a bit annoyed about this removal.
I use openpgp but now have a use for md5 (for the validation of some encrypted blobs). I'd have appreciated that this code (already bundled with openpgp) be available.

twiss:
Yeah, I understand. The "proper" way to do this, however (reuse code between OpenPGP.js and the app) would be to split up these components into separate modules, and then import them from both OpenPGP.js and your app. Of course, we're not there yet, but that's kind of the direction we'd like to move in. Additionally, one result of this might be that we'd use an external library for md5 (if it's a good replacement) rather than having our own implementation, for example. If you want to help with this, and for example scout out if there's a good implementation of md5 (also for your own usage, although, obligatory warning, md5 should not be used for any security-critical applications), that'd be welcome of course.

Similarly for the util functions; perhaps there is already a good ES6 module / library that allows importing individual util functions, if not, perhaps we should make one. Exporting `Uint8Array_to_hex` from OpenPGP.js does not make much sense, after all, and by having a separate module, other libraries / applications that don't need OpenPGP.js can also use those utils.

larabr:
@drzraf as far as the utils are concerned, you can have a look at [pmcrypto](https://github.com/ProtonMail/pmcrypto/blob/main/lib/utils.ts), where we keep a subset of the openpgpjs utils, which may be sufficient for your use case. You can import the util file only in case (`@protontech/pmcrypto/lib/utils`). We also have an md5 implementation there: https://github.com/ProtonMail/pmcrypto/blob/main/lib/crypto/_md5.js .

I have reservations about maintaining a separate repo for the crypto parts, since these are fairly openpgp-specific, and security-wise the implemented algos are not always state-of-the-art; on the contrary, they have been deprecated by e.g. NIST (see RSA with PKCS#1 padding). I believe there are better libs out there, designed with general purpose in mind. Plus, thanks to the development of SubtleCrypto (WebCrypto API) and its integration in Node, the need for platform-specific code is diminishing.

PR #1179, opened 2020-11-14
[v5] Explicit algorithm parameter in openpgp.generateKey

larabr:
- Changes openpgp.generateKey to accept an explicit `type` parameter, instead of inferring its value from the `curve` or `rsaBits` params
- Introduces `config.minRsaBits` to set minimum key size of RSA key generation


src/key/key.js
```diff
@@ -881,7 +881,7 @@ class Key {
       throw new Error("Subkey could not be encrypted here, please encrypt whole key");
     }
     if (util.getWebCryptoAll() && options.rsaBits < 2048) {
-      throw new Error('When using webCrypto rsaBits should be 2048 or 4096, found: ' + options.rsaBits);
+      throw new Error(`rsaBits should be at least 2048, got: ${options.rsaBits}`);
```
larabr, line 884: Either put a RSA size lower bound regardless of webcrypto (config?), or skip this check altogether

twiss: Yeah, we could make a config option for the minimum rsa bits indeed :+1: And then lower it in the tests. And I think we could also remove the check for web crypto in the tests, as web crypto also supports rsa bits < 2048 (nowadays?). Though I think it doesn't necessarily have to happen in this PR

src/key/helper.js
```diff
@@ -334,24 +335,38 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
...
+        options.curve = options.sign ? enums.curve.ed25519 : enums.curve.curve25519;
+      }
+      if (options.sign) {
+        options.algorithm = options.curve === enums.curve.ed25519 ? enums.publicKey.eddsa : enums.publicKey.ecdsa;
+      } else {
+        options.algorithm = enums.publicKey.ecdh;
+      }
+      break;
+    case "rsa":
+    case "rsaEncryptSign":
```
twiss, line 365: Maybe also add `rsaEncrypt` and `rsaSign` here, for the benefit of `addSubkey`?

src/key/helper.js
```diff
@@ -334,24 +335,38 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
...
+      options.algorithm = enums.publicKey.rsaEncryptSign;
+      break;
+    default:
+      throw new Error(`Unrecognized key algorithm ${options.algorithm}`);
```
twiss, line 369:
```suggestion
      throw new Error(`Unsupported key algorithm ${options.algorithm}`);
```
In case they try to add a subkey to a DSA key or so, it's not necessarily unrecognized (although it could be indeed)

test/general/key.js
```diff
@@ -2297,20 +2301,25 @@ function versionSpecificTests() {
...
+    try {
+      const { key } = await openpgp.generateKey(opt);
       expect(key.users.length).to.equal(1);
       expect(key.users[0].userId.userid).to.equal(userId);
       expect(key.users[0].selfCertifications[0].isPrimaryUserID).to.be.true;
       expect(key.getAlgorithmInfo().algorithm).to.equal('rsaEncryptSign');
-      expect(key.getAlgorithmInfo().bits).to.equal(opt.rsaBits);
-      expect(key.getAlgorithmInfo().rsaBits).to.equal(key.getAlgorithmInfo().bits);
+      expect(key.getAlgorithmInfo().rsaBits).to.equal(opt.rsaBits);
+      expect(key.getAlgorithmInfo().rsaBits).to.equal(key.getAlgorithmInfo().rsaBits);
```
larabr, line 2318:
```suggestion
```

test/general/key.js
```diff
@@ -3383,12 +3397,27 @@ VYGdb3eNlV8CfoEC
...
+    it('Add a new rsa subkey to an rsaSign key', async function() {
+      const userId = 'test <a@b.com>';
+      const opt = { algorithm: "rsa", rsaBits, userIds: [userId], subkeys: [] };
+      try {
+        const { key } = await openpgp.generateKey(opt);
+        expect(key.subKeys).to.have.length(0);
+        key.keyPacket.algorithm = "rsaSign";
+        const newKey = await key.addSubkey();
+        expect(newKey.subKeys[0].getAlgorithmInfo().algorithm).to.equal('rsaEncryptSign');
+      } finally {
```
larabr, line 3416:
```suggestion

```

test/general/key.js
```diff
@@ -3365,12 +3374,17 @@ VYGdb3eNlV8CfoEC
   });
 
   describe('addSubkey functionality testing', function() {
-    let rsaBits = 1024;
+    const rsaBits = 512;
```
larabr, line 3377:
```suggestion
    const rsaBits = 1024;
```

src/openpgp.js
```diff
@@ -62,28 +62,30 @@ if (globalThis.ReadableStream) {
...
- * @param  {Number} rsaBits          (optional) number of bits for RSA keys: 2048 or 4096.
- * @param  {Number} keyExpirationTime (optional) The number of seconds after the key creation time that the key expires
- * @param  {String} curve            (optional) elliptic curve for ECC keys:
- *                                              curve25519, p256, p384, p521, secp256k1,
- *                                              brainpoolP256r1, brainpoolP384r1, or brainpoolP512r1.
- * @param  {Date} date               (optional) override the creation date of the key and the key signatures
- * @param  {Array<Object>} subkeys   (optional) options for each subkey, default to main key options. e.g. [{sign: true, passphrase: '123'}]
- *                                              sign parameter defaults to false, and indicates whether the subkey should sign rather than encrypt
+ * Generates a new OpenPGP key pair. Supports RSA and ECC keys. By default, primary and subkeys will be of same type.
+ * @param  {String} algorithm         (optional) The primary key algorithm: ECC (default) or RSA
```
twiss, line 66: (Some of the comments on `openpgp.key.generate` apply here too)

src/openpgp.js
```diff
@@ -62,28 +62,30 @@ if (globalThis.ReadableStream) {
...
  * @static
  */
-export function generateKey({ userIds = [], passphrase = "", rsaBits = null, keyExpirationTime = 0, curve = "curve25519", date = new Date(), subkeys = [{}] }) {
+export function generateKey({ userIds = [], passphrase = "", algorithm = "ecc", rsaBits, curve = "curve25519", keyExpirationTime = 0, date = new Date(), subkeys = [{}] }) {
   userIds = toArray(userIds);
-  curve = rsaBits ? "" : curve;
-  const options = { userIds, passphrase, rsaBits, keyExpirationTime, curve, date, subkeys };
-  if (util.getWebCryptoAll() && rsaBits && rsaBits < 2048) {
-    throw new Error('rsaBits should be 2048 or 4096, found: ' + rsaBits);
+  rsaBits = rsaBits || Math.max(4096, config.rsaBitsMin);
```
twiss, line 84: I wouldn't magically fix this function if `rsaBitsMin > 4096`. In that (kinda weird) scenario they can also pass `rsaBits` manually imo.

src/openpgp.js
```diff
@@ -62,28 +62,30 @@ if (globalThis.ReadableStream) {
...
+  algorithm = algorithm.toLowerCase();
```
twiss, line 85: Also I would only accept lowercase strings. Fewer variations & options that way.

src/config/config.js
```diff
@@ -106,6 +106,11 @@ export default {
    * @property {Boolean} rsaBlinding
    */
   rsaBlinding: true,
+  /**
+   * @memberof module:config
+   * @property {Number} rsaBitsMin Minimum RSA key size allowed for key generation
+   */
+  rsaBitsMin: 2048,
```
twiss, line 113: Maybe `minRsaBits` or `minRSABits`? For consistency with `maxUseridLength` etc.
Or alternatively, `rsaGenerationMinBits`, in case we ever want to add a minimum for imported RSA keys.

src/key/factory.js
```diff
@@ -37,20 +37,15 @@ import { unarmor } from '../encoding/armor';
...
- *                             Assumes already in form of "User Name <username@email.com>"
- *                             If array is used, the first userId is set as primary user Id
- * @param {String}  options.passphrase The passphrase used to encrypt the resulting private key
- * @param {Number} [options.keyExpirationTime=0]
- *                             The number of seconds after the key creation time that the key expires
- * @param  {String} options.curve            (optional) elliptic curve for ECC keys
- * @param  {Date} options.date         Override the creation date of the key and the key signatures
- * @param  {Array<Object>} options.subkeys   (optional) options for each subkey, default to main key options. e.g. [{sign: true, passphrase: '123'}]
+ * By default, primary and subkeys will be of same type.
+ * @param {String}  options.algorithm         The primary key algorithm: ECC (default) or RSA
```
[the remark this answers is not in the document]
larabr: It can actually be also "ecdh", "rsaSign" etc

src/key/factory.js
```diff
@@ -37,20 +37,15 @@ import { unarmor } from '../encoding/armor';
...
+ * @param {Integer} options.rsaBits           Number of bits for RSA keys
+ * @param {String}  options.curve             Elliptic curve for ECC keys
```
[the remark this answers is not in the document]
larabr: There are no defaults here, which is why they aren't marked as "optional" params either. If these are not set, the code eventually throws.

src/key/key.js
```diff
@@ -880,14 +881,19 @@ class Key {
...
     const secretKeyPacket = this.primaryKey;
     if (!secretKeyPacket.isDecrypted()) {
       throw new Error("Key is not decrypted");
     }
     const defaultOptions = secretKeyPacket.getAlgorithmInfo();
+    defaultOptions.rsaBits = Math.max(defaultOptions.bits || 4096, config.rsaBitsMin);
+    defaultOptions.curve = defaultOptions.curve || "curve25519";
+    if (defaultOptions.algorithm === "dsa") {
+      defaultOptions.algorithm = "rsa";
+    }
```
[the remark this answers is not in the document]
larabr: The primary key cannot have ElGamal as algo

src/openpgp.js
larabr: On the other hand, I'd find it weirder if the code were to throw based on certain (otherwise supported) values of `rsaBitsMin`...

src/key/key.js
```diff
@@ -880,14 +881,19 @@ class Key {
...
+    if (options.rsaBits < config.rsaBitsMin) {
+      throw new Error(`rsaBits should be at least ${config.rsaBitsMin}, got: ${options.rsaBits}`);
     }
     const secretKeyPacket = this.primaryKey;
     if (!secretKeyPacket.isDecrypted()) {
       throw new Error("Key is not decrypted");
     }
     const defaultOptions = secretKeyPacket.getAlgorithmInfo();
+    defaultOptions.rsaBits = Math.max(defaultOptions.bits || 4096, config.rsaBitsMin);
+    defaultOptions.curve = defaultOptions.curve || "curve25519";
```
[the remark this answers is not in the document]
larabr: We currently also apply these defaults directly in `generateKey`. We could move them to `sanitizeOptions`, but I don't think that lower-level functions should apply such specific defaults.

src/config/config.js
larabr: If we go with `minRSABits` (which I think makes sense in terms of casing) we should also make the other options and names consistent with this rule before v5 release (e.g. `maxUserIDLength`). Otherwise we should stick to `minRsaBits` imo.

twiss: :+1:
(You changed it to `minRsaBits` everywhere except here I think)

src/key/factory.js
```diff
@@ -37,20 +37,15 @@ import { unarmor } from '../encoding/armor';
...
- * @param {Number} [options.keyExpirationTime=0]
- *                             The number of seconds after the key creation time that the key expires
- * @param  {String} options.curve            (optional) elliptic curve for ECC keys
- * @param  {Date} options.date         Override the creation date of the key and the key signatures
- * @param  {Array<Object>} options.subkeys   (optional) options for each subkey, default to main key options. e.g. [{sign: true, passphrase: '123'}]
+ * By default, primary and subkeys will be of same type.
+ * @param {ecc|rsa} options.type              The primary key algorithm type: ECC or RSA
+ * @param {String}  options.curve             Elliptic curve for ECC keys
+ * @param {Integer} options.rsaBits           Number of bits for RSA keys
+ * @param {Array<String>}  options.userIds    Array of user IDs e.g. [{ name:'Phil Zimmermann', email:'phil@openpgp.org' }]
```
twiss, line 44: The type contradicts the example (the type says string but the example is an object). In fact the type can be `String|Object|Array<String>|Array<Object>`, though perhaps that's a bit unwieldy :upside_down_face:
(Same below)

src/key/helper.js
```diff
@@ -334,24 +335,27 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
...
-      options.algorithm = options.curve === enums.curve.ed25519 ? enums.publicKey.eddsa : enums.publicKey.ecdsa;
-    } else {
-      options.algorithm = enums.publicKey.ecdh;
-    }
-  } else if (options.rsaBits) {
-    options.algorithm = enums.publicKey.rsaEncryptSign;
-  } else {
-    throw new Error('Unrecognized key type');
+  switch (options.type) {
+    case "ecc":
```
twiss, line 339: This is very nitpicky but can we have single quotes? We should standardize on one eventually and we have more single quotes

src/openpgp.js
twiss: I disagree; `minRsaBits` is a config value that controls the minimum value of `rsaBits` that is passed to `generateKey`, it shouldn't also control the default, IMO.

src/key/factory.js
larabr: I think we require an array here, only generateKey takes a single string/object.

.gitignore
```diff
@@ -4,4 +4,4 @@ npm*
 test/lib/
 dist/
 openpgp.store/
-.nyc_output/
+.nyc_output/
```
twiss, line 7: Nit: could you revert this removed newline?

src/key/key.js
```diff
@@ -880,14 +881,17 @@ class Key {
...
+      throw new Error(`rsaBits should be at least ${config.minRsaBits}, got: ${options.rsaBits}`);
     }
     const secretKeyPacket = this.primaryKey;
     if (!secretKeyPacket.isDecrypted()) {
       throw new Error("Key is not decrypted");
     }
     const defaultOptions = secretKeyPacket.getAlgorithmInfo();
+    defaultOptions.type = defaultOptions.curve ? 'ecc' : 'rsa'; // DSA keys default to RSA
+    defaultOptions.rsaBits = defaultOptions.bits || 4096;
+    defaultOptions.curve = defaultOptions.curve || 'curve25519';
```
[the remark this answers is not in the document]
larabr: They are needed if I say I want e.g. an ECC subkey for RSA primary key, without specifying the curve.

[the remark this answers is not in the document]
twiss: Ah yeah, fair enough :+1:

PR #1181, opened 2020-11-16
Fix "TypeError: fetch is not a function" in Node.js environment

YarmoM:
What PR #1052 fixed for HKP lookup, this PR fixes WKD. Uses the exact same code as the other fix for consistency.


twiss (approved):

twiss:
:+1: Thanks!

PR #1183, opened 2020-12-03
[v5] Use \n instead of \r\n for EOL when armoring

larabr:
Armoring now uses `\n` instead of `\r\n` for new lines.

### DIFF
--- a/test/general/streaming.js
+++ b/test/general/streaming.js
@@ -196,7 +196,7 @@ function tests() {
       passwords: ['test'],
     });
     const reader = openpgp.stream.getReader(encrypted);
-    expect(await reader.peekBytes(1024)).to.match(/^-----BEGIN PGP MESSAGE-----\r\n/);
+    expect(await reader.peekBytes(1024)).to.match(/^-----BEGIN PGP MESSAGE-----\n/);
     dataArrived();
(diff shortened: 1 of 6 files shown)


test/general/streaming.js
```diff
@@ -874,7 +874,7 @@ function tests() {
     });
     expect(openpgp.stream.isStream(signed)).to.equal(expectedType);
     const reader = openpgp.stream.getReader(signed);
-    expect((await reader.readBytes(31)).toString('utf8')).to.equal('-----BEGIN PGP SIGNATURE-----\r\n');
+    expect((await reader.readBytes(31)).toString('utf8')).to.equal('-----BEGIN PGP SIGNATURE-----\n\n');
     dataArrived();
     reader.releaseLock();
     await openpgp.stream.cancel(signed, new Error('canceled by test'));
```
larabr, line 880: This test fails in node because the stream does not get cancelled. How is this related to the changes?

twiss: This is a bit of a guess, but you could try reading one fewer byte, and then expect one fewer newline? As the second newline you're currently expecting comes from a different chunk than the previous expected value.

twiss (approved):

PR #1185, opened 2020-12-08
[v5] Fix parsing of short P-521 keys and EdDSA, RSA signatures

larabr:
P-521 ECC keys as well as EdDSA and RSA signatures can be shorter than expected and need to be padded correctly.
Due to a mistake in `util.padToLength`, in some cases we failed to properly parse short P-521 keys and EdDSA signatures.

- Fix P-521 parsing and EdDSA signature parsing
- Pad RSA signatures
- Strip leading zeros when encoding to MPI
- Fix ECDSA validation tests (which were never run)

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -186,7 +186,7 @@ export default {
    */
   padToLength(bytes, length, endianess = 'be') {
     const padded = new Uint8Array(length);
-    const offset = (endianess === 'be') ? 0 : (length - bytes.length);
+    const offset = (endianess === 'le') ? 0 : (length - bytes.length);
     padded.set(bytes, offset);
     return padded;
   },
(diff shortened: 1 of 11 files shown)


src/util.js
```diff
@@ -197,9 +196,12 @@ export default {
...
    */
   uint8ArrayToMpi: function (bin) {
-    const size = (bin.length - 1) * 8 + util.nbits(bin[0]);
+    let i; // index of leading non-zero byte
+    for (i = 0; i < bin.length; i++) if (bin[i] !== 0) break;
+    const stripped = bin.subarray(i);
+    const size = (stripped.length - 1) * 8 + util.nbits(stripped[0]);
     const prefix = Uint8Array.from([(size & 0xFF00) >> 8, size & 0xFF]);
-    return util.concatUint8Array([prefix, bin]);
+    return util.concatUint8Array([prefix, stripped]);
```
twiss, line 204: This code breaks for a `Uint8Array` of all zeros. I would either do
```
for (i = 0; i < bin.length - 1; i++) if (bin[i] !== 0) break;
```
so that it returns `[0, 1, 0]`, or alternatively we should return `[0, 0]` (but that's a bit more annoying). Right now it returns `[255, 249]`.

larabr: Mmm based on the spec I think it should be `[0, 0]`, since they "the length starting from its most significant non-zero bit".
But I wonder whether this case should actually be supported at all?

twiss: Yeah, I agree that `[0, 0]` is probably the correct one; I also agree that we shouldn't spend too much effort to achieve that, since it shouldn't happen, but I also don't think this function should be able to return malformed MPIs. Ideally we would come up with some relatively branchless code that handles all cases correctly. Maybe something like
```js
    for (i = 0; i < bin.length; i++) if (bin[i] !== 0) break;
    const stripped = bin.subarray(i);
    const size = util.nbits(stripped[0]) + stripped.subarray(1).length;
```
And then change `nbits(0)` / `nbits(undefined)` to return `0`?

larabr: I would consider throwing an error if we are given an array of all zeros, because it is really not expected, and actually sign of trouble for any case I can think of.

twiss: Yeah, that also works :+1:

PR #1187, opened 2020-12-14
[v5] Require User IDs to be objects; refactor UserIDPacket

larabr:
- `openpgp.generateKey` now expects userIds in object format (strings are no longer supported)
- Remove `util.parseUserId` and `util.formatUserId`
- Replace `UserIDPacket#format` with `UserIDPacket.fromObject`

### DIFF
--- a/src/packet/userid.js
+++ b/src/packet/userid.js
@@ -19,9 +19,11 @@
  * @requires enums
  * @requires util
  */
+import emailAddresses from 'email-addresses';
 
 import enums from '../enums';
 import util from '../util';
+import config from '../config';
 
 /**
  * Implementation of the User ID Packet (Tag 13)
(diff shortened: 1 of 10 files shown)


src/packet/userid.js
```diff
@@ -74,14 +76,22 @@ class UserIDPacket {
...
 
   /**
-   * Set userid string from object, e.g. { name:'Phil Zimmermann', email:'phil@openpgp.org' }
+   * Set userId from object
+   * @param {Object} userId  object specifying userId name, email and comment
    */
-  format(userid) {
-    if (util.isString(userid)) {
-      userid = util.parseUserId(userid);
+  fromObject(userId) {
```
twiss, line 82: I meant like this:
```suggestion
  static fromObject(userId) {
```
and then construct a `UserIDPacket` and assign to it inside the function, and call `UserIDPacket.fromObject(userId)` elsewhere (similar to `Message.fromText`)

PR #1191, opened 2020-12-15
[v5] Drop inconsistent return types and fix type definitions

larabr:
- Change several internal `encrypt`, `decrypt`, `sign`, `verify` functions to throw on error, without returning anything on success (the returned value was unused anyway).
- Update and fix type definitions

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -16,33 +16,33 @@ export function readKeys(data: Uint8Array): Promise<Key[]>;
 export function generateKey(options: KeyOptions): Promise<KeyPair>;
 export function generateSessionKey(options: { publicKeys: Key[], date?: Date, toUserIds?: UserId[] }): Promise<SessionKey>;
 export function decryptKey(options: { privateKey: Key; passphrase?: string | string[]; }): Promise<Key>;
-export function encryptKey(options: { privateKey: Key; passphrase?: string }): Promise<Key>;
+export function encryptKey(options: { privateKey: Key; passphrase?: string | string[] }): Promise<Key>;
 export function reformatKey(options: { privateKey: Key; userIds?: (string | UserId)[]; passphrase?: string; keyExpirationTime?: number; }): Promise<KeyPair>;
 
 export class Key {
+  constructor(packetlist: PacketList<AnyPacket>);
   public primaryKey: PublicKeyPacket | SecretKeyPacket;
   public subKeys: SubKey[];
   public users: User[];
   public revocationSignatures: SignaturePacket[];
   public keyPacket: PublicKeyPacket | SecretKeyPacket;
-  constructor(packetlist: PacketList<AnyPacket>);
   public armor(): string;
-  public decrypt(passphrase: string | string[], keyId?: Keyid): Promise<boolean>;
-  public encrypt(passphrase: string | string[]): Promise<void>;
-  public getExpirationTime(capability?: 'encrypt' | 'encrypt_sign' | 'sign' | null, keyId?: Keyid | null, userId?: UserId | null): Promise<Date | typeof Infinity | null>; // Returns null if `capabilities` is passed and the key does not have the specified capabilities or is revoked or invalid.
+  public decrypt(passphrase: string | string[], keyId?: Keyid): Promise<void>; // throws on error
+  public encrypt(passphrase: string | string[]): Promise<void>; // throws on error
+  public getExpirationTime(capability?: 'encrypt' | 'encrypt_sign' | 'sign', keyId?: Keyid, userId?: UserId): Promise<Date | typeof Infinity | null>; // Returns null if `capabilities` is passed and the key does not have the specified capabilities or is revoked or invalid.
   public getKeyIds(): Keyid[];
-  public getPrimaryUser(): Promise<PrimaryUser>; // throws on err
+  public getPrimaryUser(): Promise<PrimaryUser>; // throws on error
   public getUserIds(): string[];
   public isPrivate(): boolean;
   public isPublic(): boolean;
   public toPublic(): Key;
   public update(key: Key): void;
-  public verifyPrimaryKey(): Promise<void>; // throws on err
+  public verifyPrimaryKey(): Promise<void>; // throws on error
   public isRevoked(): Promise<boolean>;
   public revoke(reason: { flag?: enums.reasonForRevocation; string?: string; }, date?: Date): Promise<Key>;
   public getRevocationCertificate(): Promise<Stream<string> | string | undefined>;
-  public getEncryptionKey(keyid?: Keyid | null, date?: Date, userId?: UserId | null): Promise<Key | SubKey | null>;
-  public getSigningKey(keyid?: Keyid | null, date?: Date, userId?: UserId | null): Promise<Key | SubKey | null>;
+  public getEncryptionKey(keyid?: Keyid, date?: Date | null, userId?: UserId): Promise<Key | SubKey>;
+  public getSigningKey(keyid?: Keyid, date?: Date | null, userId?: UserId): Promise<Key | SubKey>;
   public getKeys(keyId?: Keyid): (Key | SubKey)[];
   public isDecrypted(): boolean;
   public getFingerprint(): string;
@@ -52,11 +52,10 @@ export class Key {
 }
 
 export class SubKey {
-  public subKey: SecretSubkeyPacket | PublicSubkeyPacket;
-  public keyPacket: SecretKeyPacket;
+  constructor(subKeyPacket: SecretSubkeyPacket | PublicSubkeyPacket);
+  public keyPacket: SecretSubkeyPacket | PublicSubkeyPacket;
   public bindingSignatures: SignaturePacket[];
   public revocationSignatures: SignaturePacket[];
-  constructor(subKeyPacket: SecretSubkeyPacket | PublicSubkeyPacket);
   public verify(primaryKey: PublicKeyPacket | SecretKeyPacket): Promise<enums.keyStatus>;
   public isDecrypted(): boolean;
   public getFingerprint(): string;
@@ -80,7 +79,8 @@ export interface PrimaryUser {
 
 type AlgorithmInfo = {
   algorithm: enums.publicKeyNames;
-  bits: number;
+  bits?: number;
+  curve?: EllipticCurveName;
 };
 
 /* ############## v5 SIG #################### */
@@ -94,7 +94,7 @@ export class Signature {
   public armor(): string;
 }
 
-export interface VerificationResult {
+interface VerificationResult {
   keyid: Keyid;
   verified: Promise<null | boolean>;
   signature: Promise<Signature>;
@@ -113,7 +113,7 @@ export class CleartextMessage {
 
   /** Returns the key IDs of the keys that signed the cleartext message
    */
-  getSigningKeyIds(): Array<Keyid>;
+  getSigningKeyIds(): Keyid[];
 
   /** Get cleartext
    */
@@ -123,7 +123,7 @@ export class CleartextMessage {
    *
    *  @param privateKeys private keys with decrypted secret key data for signing
    */
-  sign(privateKeys: Array<Key>): void;
+  sign(privateKeys: Key[]): void;
 
   /** Verify signatures of cleartext signed message
    *  @param keys array of keys to verify signatures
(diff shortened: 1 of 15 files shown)


src/packet/signature.js
```diff
@@ -672,7 +671,8 @@ class SignaturePacket {
    * @param {String|Object} data data which on the signature applies
    * @param {Boolean} detached (optional) whether to verify a detached signature
    * @param {Boolean} streaming (optional) whether to process data as a stream
-   * @returns {Promise<Boolean>} True if message is verified, else false.
+   * @return true if the signature is valid
```
larabr, line 674: I wanted to also change this one to always return nothing, but it turns out we check its returned value in several tests, as well as in `getLatestValidSignature` -- I would still change it, @twiss what do you think?

twiss: Yeah, for this function the returned boolean used to actually be significant. I think there's a few other places too if you do a global search for `verified ||`. Which - maybe we should fold that into this function - and also the signature expiry checking. E.g. [this](https://github.com/openpgpjs/openpgpjs/blob/2eab8a1ebcd84f913f948c6da845cd66020ec170/src%2Fkey%2Fhelper.js#L252-L269):
```js
    try {
      if (
        (!signature || revocationSignature.issuerKeyId.equals(signature.issuerKeyId)) &&
        !(config.revocations_expire && revocationSignature.isExpired(normDate)) &&
        (revocationSignature.verified || await revocationSignature.verify(key, signatureType, dataToVerify))
      ) {
        revocationKeyIds.push(revocationSignature.issuerKeyId);
      }
    } catch (e) {}
```
should probably just be something like
```js
    if (!signature || revocationSignature.issuerKeyId.equals(signature.issuerKeyId)) {
      try {
        await revocationSignature.verify(key, signatureType, dataToVerify));
        revocationKeyIds.push(revocationSignature.issuerKeyId);
      } catch (e) {}
    }
```
(but with comments intact). Though that probably shouldn't happen in this PR. But anyway yes I think you can drop the return value and change all the calling locations ^.^

larabr: You mean we should ignore the "cached" `.verified` value, and re-verify every time?

twiss: I meant the former. I guess the most concrete case is if you call `.verify` twice with different data, or different keys. Which.. for (detached) message signatures is actually kinda plausible, so yeah maybe we should keep it outside indeed. Though - for this to affect the result you'd have to first verify with the _correct_ message & key, and _then_ with a wrong one. I don't really see a concrete use-case for that, but I guess it's better to be correct.

openpgp.d.ts
```diff
@@ -643,8 +630,9 @@ export interface FullKeyOptions {
   subkeys: KeyOptions[]; // required unlike KeyOptions.subkeys
 }
 
-export interface Keyid {
+export class KeyId {
```
twiss, line 633: This implies that there is an `openpgp.KeyId` class that you can instantiate, which is not the case, so this should remain an interface. Also, the name of the (unexported) class is `Keyid`; it we wanna change this I think we should change the name of the class too, to `KeyID`. But for now I would keep it as is; the name of an interface doesn't really matter that much.

openpgp.d.ts
```diff
@@ -643,8 +630,9 @@ export interface FullKeyOptions {
...
   bytes: string;
+  equals(keyid: KeyId, matchWildcard?: boolean): boolean;
```
twiss, line 635: By far the most commonly used method here should be `toHex`, maybe you want to add that one?

openpgp.d.ts
```diff
@@ -41,14 +41,14 @@ export class Key {
   public isRevoked(): Promise<boolean>;
   public revoke(reason: { flag?: enums.reasonForRevocation; string?: string; }, date?: Date): Promise<Key>;
   public getRevocationCertificate(): Promise<Stream<string> | string | undefined>;
-  public getEncryptionKey(keyid?: Keyid | null, date?: Date, userId?: UserId | null): Promise<Key | SubKey | null>;
-  public getSigningKey(keyid?: Keyid | null, date?: Date, userId?: UserId | null): Promise<Key | SubKey | null>;
-  public getKeys(keyId?: Keyid): (Key | SubKey)[];
+  public getEncryptionKey(keyid?: KeyId, date?: Date | null, userId?: UserId): Promise<Key | SubKey | null>;
+  public getSigningKey(keyid?: KeyId, date?: Date | null, userId?: UserId): Promise<Key | SubKey | null>;
```
twiss, line 45: FWIW, these can't return `null` anymore, they throw if there is no encryption/signing key.

openpgp.d.ts
```diff
@@ -27,10 +27,10 @@ export class Key {
   public keyPacket: PublicKeyPacket | SecretKeyPacket;
   constructor(packetlist: PacketList<AnyPacket>);
   public armor(): string;
-  public decrypt(passphrase: string | string[], keyId?: Keyid): Promise<boolean>;
-  public encrypt(passphrase: string | string[]): Promise<void>;
-  public getExpirationTime(capability?: 'encrypt' | 'encrypt_sign' | 'sign' | null, keyId?: Keyid | null, userId?: UserId | null): Promise<Date | typeof Infinity | null>; // Returns null if `capabilities` is passed and the key does not have the specified capabilities or is revoked or invalid.
-  public getKeyIds(): Keyid[];
+  public decrypt(passphrase: string | string[], keyId?: KeyId): Promise<void>; // throws on err
```
twiss, line 30: Bit nitpicky but bytes aren't that expensive anymore nowadays 🙃 
```suggestion
  public decrypt(passphrase: string | string[], keyId?: KeyId): Promise<void>; // throws on error
```
(Also elsewhere in the file)

openpgp.d.ts
```diff
@@ -441,33 +427,34 @@ export class SignaturePacket extends BasePacket {
...
   public embeddedSignature: null | SignaturePacket;
   public issuerKeyVersion: null | number;
   public issuerFingerprint: null | Uint8Array;
-  public preferredAeadAlgorithms: null | Uint8Array;
+  public preferredAeadAlgorithms: enums.aead[] | null;
   public verified: null | boolean;
   public revoked: null | boolean;
-  public sign(key: SecretKeyPacket | SecretSubkeyPacket, data: Uint8Array): true;
+  public sign(key: SecretKeyPacket, data: Uint8Array, detached?: boolean, streaming?: boolean): void;
+  public verify(key: SecretKeyPacket, signatureType: enums.signature, data: Uint8Array, detached?: boolean, streaming?: boolean): Promise<true>; // throw on error
```
twiss, line 457:
```suggestion
  public verify(key: SecretKeyPacket, signatureType: enums.signature, data: Uint8Array, detached?: boolean, streaming?: boolean): Promise<true>; // throws on error
```

openpgp.d.ts
```diff
@@ -531,7 +518,7 @@ export namespace stream {
 
 /* ############## v5 GENERAL #################### */
 
-export interface UserId { name?: string; email?: string; }
+export interface UserId { name?: string; email?: string; comment?: string }
```
twiss, line 521:
```suggestion
export interface UserId { name?: string; email?: string; comment?: string; }
```

openpgp.d.ts
```diff
@@ -531,11 +529,11 @@ export namespace stream {
 
 /* ############## v5 GENERAL #################### */
 
-export interface UserId { name?: string; email?: string; }
+export interface UserId { name?: string; email?: string; comment?: string; }
 export interface SessionKey { data: Uint8Array; algorithm: string; }
 
 
-interface EncryptOptions {
+export interface EncryptOptions {
```
twiss, line 536: I think this one should be a declaration as well - but it seems like it isn't necessary to explicitly `declare` `interface`s, so I would just revert this one and change all the other `declare interface`s to `interface`

src/key/user.js
```diff
@@ -157,11 +157,12 @@ class User {
 
   /**
    * Verify User. Checks for existence of self signatures, revocation signatures
-   * and validity of self signature. Throws when there are no valid self signatures.
+   * and validity of self signature.
    * @param  {SecretKeyPacket|
    *          PublicKeyPacket} primaryKey The primary key packet
    * @param  {Date}            date       Use the given date instead of the current time
    * @returns {Promise<true>}             Status of user
```
twiss, line 164: It's not super important, but maybe you want to change this one as well?

src/message.js
```diff
@@ -762,7 +762,7 @@ async function createVerificationObject(signature, literalDataList, keys, date =
       )) {
         throw new Error('Signature is expired');
       }
-      return verified;
+      return !!sig.verified;
```
twiss, line 765: I think you can just
```suggestion
      return true;
```
here, rather than relying on the caching property

test/general/key.js
```diff
@@ -2989,17 +2989,13 @@ module.exports = () => describe('Key', function() {
...
-        })
-      ]);
-    });
+
+    await dest.update(source);
+    expect(dest.isPrivate()).to.be.true;
+    await expect(dest.verifyPrimaryKey()).to.be.fulfilled;
+    await expect(source.verifyPrimaryKey()).to.be.fulfilled;
+    await expect(dest.users[0].verify(dest.primaryKey)).to.be.fulfilled;
+    await expect(source.users[0].verify(source.primaryKey)).to.be.fulfilled;
```
twiss, line 2998: I'm still not really a fan of these, I would just await the functions directly, since they throw on failure it should then be clear what the test is checking. (Also elsewhere)

src/message.js
larabr: But that should really be true at this point, I wanted to rely on it to catch any unexpected behaviour

src/key/user.js
larabr: I wouldn't change this for now, because there is `verifyCertificate` that it is somewhat related (it's used in a similar way in `verifyAllCertifications` or `verifyAllUsers`), and that one needs to return `null | true`, hence I think it's best to uniform the two together eventually.

test/general/key.js
larabr: My position is still that since we do not test any other result/side-effect, it's clearer to be explicit about what is being tested. I feel like leaving `await verify(...)` can look as if we just forgot about those lines.

openpgp.d.ts
twiss: Sorry, what I meant was that I think this one probably makes more sense unexported as well

src/message.js
twiss: Yeah, but.. we rely on `verify` to throw in all other places, I think it's a bit confusing not to do the same here

test/general/key.js
tomholub: I think you are both right. One thing to consider is which of the two approaches produces more useful output during failures - in case there is any difference.

openpgp.d.ts
larabr: I think it may be useful to export it and allow users to instantiate and type-check objects before passing them to the function?
Something like:
```
let options : DecryptOptions;
options.message = ...;
```
I have no idea if people use TS in this way though 😄

openpgp.d.ts
```diff
@@ -80,7 +79,8 @@ export interface PrimaryUser {
 
 type AlgorithmInfo = {
   algorithm: enums.publicKeyNames;
-  bits: number;
+  bits?: number;
+  curve?: EllipticCurveName;
```
larabr, line 83: This type is not quite right, because either `bits` or `curve` are always set (but not both).
A union type can fix the issue, but it's probably a bit too verbose for now, since I don't think the current declaration creates any actual issues.

tomholub: The current declaration will mean that the consumer will have to typically bang the value as in `algo.bits!` if they need the number.

```ts
// now
if(algo.algorithm === rsa) {
  algo.bits // number | undefined
  algo.curve // EllipticCurveName | undefined
}

// union
if(algo.algorithm === rsa) {
  algo.bits // type is number, because TS can infer it if the union was written correctly
  algo.curve // compilation error!
}
```

Generally, it is ergonomic for the user of the API if types are written to behave like this. But it really depends. For example in this case, there is also a downside:

```ts
// now, without if
algo.bits // number | undefined

// union, without if
algo.bits // compilation error!

// union without if, need to cast it
(algo as RsaAlgorithmInfo).bits // number
```

For the AlgorithmInfo in particular, maybe it's not quite worth it. But other places may benefit from this.

twiss: Yeah, I think this would be cool, but doesn't necessarily have to happen in this PR. Also, with the current `algorithm: rsaSign | rsaEncrypt | rsaEncryptSign` this type (and check) might be a little overly verbose, maybe it would be nice to have a `type: 'rsa'` property.

PR #1203, opened 2021-01-31
Added Subkey.isPublic/isPrivate

alichry:
Added isPublic/isPrivate methods to the SubKey class, to provide similar behavior by openpgp.key.Key


twiss:
Closed in favor of #1205. (We don't support passing subkeys to `openpgp.encrypt` directly, and we also don't support mixed public/private keys.)

PR #1204, opened 2021-02-03
Added encryption/signing key preference

alichry:
Out of the box, OpenPGP.js selects the latest valid encryption subkey for encryption. I believe it is helpful to allow the user to selects a specific key id from the list of subkeys when encrypting/signing. This PR aims to fulfill this requirement.


alichry:
Closing to select a different branch

PR #1205, opened 2021-02-03
Added encryption/signing key preference

alichry:
Out of the box, OpenPGP.js selects the latest valid encryption subkey for encryption. I believe it is helpful to allow the user to selects a specific key id from the list of subkeys when encrypting/signing. This PR aims to fulfill this requirement.

### DIFF
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -288,13 +288,13 @@ class Key {
   }
 
   /**
-   * Sets the selected signing key id. Pass null to unselect it.
+   * Selects a signing key. Pass null to unselect it.
    *  Selecting a signing key would affect high-level API calls to
    *  openpgp as well as this.getSigningKey(). The latter would
    *  return the selected signing key when the keyId parameter is null.
    * @param {module.type/keyid|null} keyId
    */
-  setSelectedSigningKeyId(keyId) {
+  selectSigningKey(keyId) {
     this.selectedSigningKeyId = keyId;
   }
 
@@ -307,13 +307,13 @@ class Key {
   }
 
     /**
-   * Sets the selected encryption key id. Pass null to unselect it.
+   * Selects an encryption key. Pass null to unselect it.
    *  Selecting an encryption key would affect high-level API calls to
    *  openpgp as well as this.getEncryptionKey(). The latter would
    *  return the selected encryption key when the keyId parameter is null.
    * @param {module.type/keyid|null} keyId
    */
-  setSelectedEncryptionkeyId(keyId) {
+  selectEncryptionKey(keyId) {
     this.selectedEncryptionKeyId = keyId;
   }
 
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -62,6 +62,8 @@ class Key {
     this.directSignatures = [];
     this.users = [];
     this.subKeys = [];
+    this.selectedEncryptionKeyId = null;
+    this.selectedSigningKeyId = null;
     this.packetlist2structure(packetlist);
     if (!this.keyPacket) {
       throw new Error('Invalid key: need at least key packet');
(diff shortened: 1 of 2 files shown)


twiss:
Hey @alichry :wave: Thanks for the PR! I think rather than a stateful approach (where the user selects a subkey in advance and it's stored on the key object) it would be better to have a stateless approach (e.g., adding an optional `encryptionKeyId` parameter to `openpgp.encrypt`, which is passed down to `getEncryptionKey` directly, and similarly for `signingKeyId` - which you might also want in `openpgp.sign`).

alichry:
Hey @twiss 👋
I had the exact thought for handling several key ids for different encryption/signing keys, especially when there's multiple primary keys passed. Given an array of key ids and primary keys, it would be required to cross-reference the key id list with the primary keys' subkeys. Another caveat would be if the user would like to encrypt/sign data with 2 subkeys of the same primary key, this stateful approach does not help! The approach was simple enough to implement :)

For a stateless approach with an array of key ids, and to allow the user to specify sibling subkey ids, I would have to slice the array after each subkey id match. Would that be okay to implement?

twiss:
I would simply take `encryptionKeyIds[i]` as the key id to use for `publicKeys[i]`. (Same as what we do with `toUserIds` and `fromUserIds`.) That's simpler to implement and also makes it more obvious whether to throw if there is no encryption key with that key id in that public key. Then if you really want to, you could pass `{ publicKeys: [key1, key1], encryptionKeyIds: [keyId1, keyId2] }` to encrypt with multiple subkeys of the same key.

By the way, we should probably normalize the naming here a bit, e.g. `encryptionKeys`, `encryptionKeyIds`, `encryptionUserIds`, `signing*`, `decryption*`, `verification*`, etc. If you're interested in doing this, I'd be happy to merge a separate PR for this :blush:

alichry:
Yes, that is simple to implement ^_^ I believe these new parameters `*keyIds` would be optional, and one difficulty for the user is to specify all *KeyIds for all corresponding *Keys. The user might provide several keys, and would only prefer a specific subkey for the first key only, or for the first and third keys only. That can be addressed by specifying `null` on `*KeyIds[I]` to indicate no preference for `*Key[I]`. How about we allow the user to specify an _optional_ keyId along with each key? i.e.

```js
let primaryKey1, primaryKey2;
openpgp.encrypt({
  encryptionKeys: [
    {
      key: primaryKey1,
      keyId: ... 
    },
    // no preference, use any:
    primaryKey2
  ],
  ...
}, ...)
```

I hope it is okay to stretch this conversation a bit further, I am trying to find a sweet spot to implement that is inclined towards the user with respect to the complexity of the implementation and the significant change it might occur. What do you think? If we force the user to do some `null` padding to indicate a preference for non-consecutive keys, we would not have to use the above format. 

I Would love to undertake the change to normalize the naming scheme and proceed with the implementation \\(^-^)\/ Let me know what you think, and which format to use.

twiss:
Hey @alichry :wave: Sorry for the delay. I'm somewhat conflicted; I think from an API perspective your proposal makes sense, but I'm worried that it might complicate the code somewhat. Another (minor) complication is that (for example) `decryptionKeys` won't be able to take a `keyId`, but they _could_ take a `userId` (even though it's currently not possible to pass one, we should probably add this at some point). So this would introduce subtle differences in the type of the `encryptionKeys` and `decryptionKeys` parameters, whereas the absence of a `decryptionKeyIds` parameter is more clear.

> If we force the user to do some null padding to indicate a preference for non-consecutive keys, we would not have to use the above format.

So personally, for this PR, I would go with this (the simplest approach), which is the same as what we do for `userIds`. Then, in a future PR, we can unify the three parameters somehow (either by renaming them or merging them into one parameter, TBD).

twiss:
Whoops, sorry about that. Could you open a new PR to the master branch? (I've merged and deleted the `v5` branch which accidentally closed this PR.)

alichry:
Great 😊, thank you for the help and involvement ✌️ I will proceed with that. Yes, please feel free to close this PR if it's not yet closed :)

Issue #1166, opened 2020-10-19
allow to pass config into functions


tomholub:
I digged this up in a conversation between @twiss and @wiktor-k and I thought it was a great idea, we'd definitely make a good use of this if it was available :) So I'm pulling it out to give it it's own issue (unless there already is one).

-------------------------------

_Yeah. Even more than that, sometimes you want to set a config option for one message/decryption only, which is a bit difficult right now with the async model. So I think it would be even nicer to add `config` options/parameters to all of the top-level functions. Then you could also do:_
```js
try {
  result = openpgp.decrypt({ message });
} catch (e) {
  warn("Insecure!");
  result = openpgp.decrypt({ message, config: { allow_insecure_decryption_with_signing_keys: true } });
}
```
_Or make it depend on the date of the email, et cetera.
And then there would also be no statefulness in the tests, potentially allowing us to run them in parallel, etc._

_Originally posted by @twiss in https://github.com/openpgpjs/openpgpjs/pull/1148#discussion_r480181580_

tomholub:
Another possibility is something like - this way you don't have to change all of the signatures, but hard to say which is a larger change:

```ts
const laxOpenpgp = openpgp.configure({ config: { allow_insecure_decryption_with_signing_keys: true }});

// ...

try {
  result = openpgp.decrypt({ message });
} catch (e) {
  warn("Insecure!");
  result = laxOpenpgp.decrypt({ message });
}
```

twiss:
Thanks for making an issue for this, @tomholub. I would like to do this in v5.

I would prefer the former option, because v5 allows you to import individual functions, so you'd have:

```js
import { decrypt } from 'openpgp';
try {
  result = decrypt({ message });
} catch (e) {
  warn("Insecure!");
  result = decrypt({ message, config: { allowInsecureDecryptionWithSigningKeys: true } });
}
```

I don't really see a way to make this work with the second option.

tomholub:
Got it - sounds good :+1:

PR #1236, opened 2021-02-12
Consolidate read* functions

twiss:
Make all `read*` functions accept an options object, so that we can add config options to them later (for #1166). This is necessary so that we can remove the global `openpgp.config`, which doesn't work that well when importing individual functions.

Furthermore, merge `readMessage` and `readArmoredMessage` into one function, et cetera.

### DIFF
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -157,14 +157,17 @@ export class CleartextMessage {
 
 
 /**
- * reads an OpenPGP cleartext signed message and returns a CleartextMessage object
- * @param {String | ReadableStream<String>} armoredText text to be parsed
+ * Reads an OpenPGP cleartext signed message and returns a CleartextMessage object
+ * @param {String | ReadableStream<String>} cleartextMessage text to be parsed
  * @returns {module:cleartext.CleartextMessage} new cleartext message object
  * @async
  * @static
  */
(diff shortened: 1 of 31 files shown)


src/cleartext.js
```diff
@@ -155,14 +155,17 @@ export class CleartextMessage {
 
 
 /**
- * reads an OpenPGP cleartext signed message and returns a CleartextMessage object
- * @param {String | ReadableStream<String>} armoredText text to be parsed
+ * Reads an OpenPGP cleartext signed message and returns a CleartextMessage object
+ * @param {String | ReadableStream<String>} cleartextMessage text to be parsed
```
larabr, line 159: Can this be a stream? If so the TS definition should be updated, same for `readKeys`.

twiss: Technically it can be, by virtue of the underlying `unarmor` function accepting a stream. But I don't think there's really any reason to do this, as it doesn't add anything (it reads the entire stream immediately).

larabr: I think it's confusing to have mismatching docs and TS though

twiss: Yeah, true. Anyway, it's not really the concern of this PR, we can make it consistent later 😌

PR #1238, opened 2021-02-14
Added encryption/signing subkey preference

alichry:
Supersedes https://github.com/openpgpjs/openpgpjs/pull/1205

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -424,14 +424,14 @@ export class Message {
    * Sign the message (the literal data packet of the message)
    * @param  {Array<module:key.Key>} privateKeys      private keys with decrypted secret key data for signing
    * @param  {Signature} signature                    (optional) any existing detached signature to add to the message
+   * @param  {Array<module:type/keyid>} signingKeyIds (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
    * @param  {Date} date                              (optional) override the creation time of the signature
    * @param  {Array} userIds                          (optional) user IDs to sign with, e.g. [{ name:'Steve Sender', email:'steve@openpgp.org' }]
    * @param  {Boolean} streaming                      (optional) whether to process data as a stream
-   * @param  {Array<module:type/keyid>} signingKeyIds (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
    * @returns {Promise<Message>}                      new message with signed content
    * @async
    */
-  async sign(privateKeys = [], signature = null, date = new Date(), userIds = [], streaming = false, signingKeyIds = []) {
+  async sign(privateKeys = [], signature = null, signingKeyIds = [], date = new Date(), userIds = [], streaming = false) {
     const packetlist = new PacketList();
 
     const literalDataPacket = this.packets.findPacket(enums.packet.literalData);
@@ -481,7 +481,7 @@ export class Message {
     });
 
     packetlist.push(literalDataPacket);
-    packetlist.concat(await createSignaturePackets(literalDataPacket, privateKeys, signature, date, userIds, false, streaming, signingKeyIds));
+    packetlist.concat(await createSignaturePackets(literalDataPacket, privateKeys, signature, signingKeyIds, date, userIds, false, streaming));
 
     return new Message(packetlist);
   }
@@ -510,19 +510,19 @@ export class Message {
    * Create a detached signature for the message (the literal data packet of the message)
    * @param  {Array<module:key.Key>}               privateKeys private keys with decrypted secret key data for signing
    * @param  {Signature} signature                 (optional) any existing detached signature
+   * @param  {Array<module:type/keyid>} signingKeyIds (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
    * @param  {Date} date                           (optional) override the creation time of the signature
    * @param  {Array} userIds                       (optional) user IDs to sign with, e.g. [{ name:'Steve Sender', email:'steve@openpgp.org' }]
    * @param  {Boolean} streaming                   (optional) whether to process data as a stream
-   * @param  {Array<module:type/keyid>} signingKeyIds (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
    * @returns {Promise<module:signature.Signature>} new detached signature of message content
    * @async
    */
-  async signDetached(privateKeys = [], signature = null, date = new Date(), userIds = [], streaming = false, signingKeyIds = []) {
+  async signDetached(privateKeys = [], signature = null, signingKeyIds = [], date = new Date(), userIds = [], streaming = false) {
     const literalDataPacket = this.packets.findPacket(enums.packet.literalData);
     if (!literalDataPacket) {
       throw new Error('No literal data packet to sign.');
     }
-    return new Signature(await createSignaturePackets(literalDataPacket, privateKeys, signature, date, userIds, true, streaming, signingKeyIds));
+    return new Signature(await createSignaturePackets(literalDataPacket, privateKeys, signature, signingKeyIds, date, userIds, true, streaming));
   }
 
   /**
@@ -694,15 +694,15 @@ export class Message {
  * @param  {LiteralDataPacket}                 literalDataPacket the literal data packet to sign
  * @param  {Array<module:key.Key>}             privateKeys private keys with decrypted secret key data for signing
  * @param  {Signature} signature               (optional) any existing detached signature to append
+ * @param  {Array<module:type/keyid>} signingKeyIds (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
  * @param  {Date} date                         (optional) override the creationtime of the signature
  * @param  {Array} userIds                     (optional) user IDs to sign with, e.g. [{ name:'Steve Sender', email:'steve@openpgp.org' }]
  * @param  {Boolean} detached                  (optional) whether to create detached signature packets
  * @param  {Boolean} streaming                 (optional) whether to process data as a stream
- * @param  {Array<module:type/keyid>} signingKeyIds (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
  * @returns {Promise<module:PacketList>} list of signature packets
  * @async
  */
-export async function createSignaturePackets(literalDataPacket, privateKeys, signature = null, date = new Date(), userIds = [], detached = false, streaming = false, signingKeyIds = []) {
+export async function createSignaturePackets(literalDataPacket, privateKeys, signature = null, signingKeyIds = [], date = new Date(), userIds = [], detached = false, streaming = false) {
   const packetlist = new PacketList();
 
   // If data packet was created from Uint8Array, use binary, otherwise use text
--- a/src/message.js
+++ b/src/message.js
@@ -330,7 +330,7 @@ export class Message {
 
     const { data: sessionKeyData, algorithm, aeadAlgorithm } = sessionKey;
 
-    const msg = await Message.encryptSessionKey(sessionKeyData, algorithm, aeadAlgorithm, keys, passwords, wildcard, date, userIds, encryptionKeyIds);
+    const msg = await Message.encryptSessionKey(sessionKeyData, algorithm, aeadAlgorithm, keys, passwords, wildcard, encryptionKeyIds, date, userIds);
 
     let symEncryptedPacket;
     if (aeadAlgorithm) {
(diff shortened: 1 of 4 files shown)


test/general/openpgp.js
```diff
@@ -2588,6 +2703,118 @@ amnR6g==
...
+        let d;
+        let m;
+        let pList;
+        let kIds;
+        let sIds;
+        // select every tuple of every 2 pair from each set.
+        for (let u = 0; u < encryptionKeyIds.length; u++) {
+          for (let v = 0; v < encryptionKeyIds.length; v++) {
+            for (let w = 0; w < signingKeyIds.length; w++) {
+              for (let z = 0; z < signingKeyIds.length; z++) {
```
twiss, line 2789: It looks like this test failed because it timed out, because it took longer than 10 seconds, probably because of the high number of combinations here. One way to fix this would be to try all the combinations in parallel (e.g. with `await Promise.all(...)`).

But, to be perfectly honest, I have the feeling that testing all the combinations probably doesn't really add much - testing the default is already done in other tests, for example, and I don't think there's a big chance that one combination would work and another wouldn't. The main thing that I feel this test adds is being able to encrypt (and sign) with multiple subkeys of the same key.

So I would personally reduce this test to that: pass the three encryption key IDs, and the three signing key IDs, and then afterwards, check that they match. That should simplify the test and make it a lot faster.

src/message.js
```diff
@@ -304,17 +304,18 @@ export class Message {
...
+   * @param  {Boolean} wildcard                           (optional) use a key ID of 0 instead of the public key IDs
+   * @param  {Date} date                                  (optional) override the creation date of the literal package
+   * @param  {Array<Object>} userIds                      (optional) user IDs to encrypt for, e.g. [{ name:'Robert Receiver', email:'robert@openpgp.org' }]
+   * @param  {Boolean} streaming                          (optional) whether to process data as a stream
+   * @param  {Array<module:type/keyid>} encryptionKeyIds  (optional) array of key IDs to use for encryption. Each encryptionKeyIds[i] corresponds to publicKeys[i]
    * @returns {Promise<Message>}                   new message with encrypted content
    * @async
    */
-  async encrypt(keys, passwords, sessionKey, wildcard = false, date = new Date(), userIds = [], streaming) {
+  async encrypt(keys, passwords, sessionKey, wildcard = false, date = new Date(), userIds = [], streaming, encryptionKeyIds = []) {
```
twiss, line 318: This is very nitpicky, but could you make the order:
```suggestion
  async encrypt(keys, passwords, sessionKey, wildcard = false, encryptionKeyIds = [], date = new Date(), userIds = [], streaming) {
```
(for symmetry with the existing `getEncryptionKey` function, which takes `keyId, date, userId`)?

(And similarly in the other functions.)

src/openpgp.js
```diff
@@ -240,24 +240,26 @@ export async function encryptKey({ privateKey, passphrase }) {
...
+ * @param  {Signature} signature                        (optional) a detached signature to add to the encrypted message
+ * @param  {Boolean} wildcard                           (optional) use a key ID of 0 instead of the public key IDs
+ * @param  {Date} date                                  (optional) override the creation date of the message signature
+ * @param  {Array<Object>} fromUserIds                  (optional) array of user IDs to sign with, one per key in `privateKeys`, e.g. [{ name:'Steve Sender', email:'steve@openpgp.org' }]
+ * @param  {Array<Object>} toUserIds                    (optional) array of user IDs to encrypt for, one per key in `publicKeys`, e.g. [{ name:'Robert Receiver', email:'robert@openpgp.org' }]
+ * @param  {Array<module:type/keyid>} encryptionKeyIds  (optional) array of key IDs to use for encryption. Each encryptionKeyIds[i] corresponds to publicKeys[i]
+ * @param  {Array<module:type/keyid>} signingKeyIds     (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
```
twiss, line 257: And then also here, I would order them `signingKeyIds`, `encryptionKeyIds`, `date`, `fromUserIds`, `toUserIds`, even though it looks a bit weird, to be consistent with e.g. `getEncryptionKey`. (And idem below.)

src/message.js
```diff
@@ -462,7 +465,10 @@ export class Message {
       if (privateKey.isPublic()) {
         throw new Error('Need private key for signing');
       }
-      const signingKey = await privateKey.getSigningKey(undefined, date, userIds);
+      const signingKeyId = privateKeys.length - 1 - i >= 0 ?
+        signingKeyIds[privateKeys.length - 1 - i] :
+        signingKeyIds[0];
+      const signingKey = await privateKey.getSigningKey(signingKeyId, date, userIds);
```
twiss, line 471: I'm confused, shouldn't this just be `signingKeyIds[i]`?

src/message.js
```diff
@@ -351,23 +352,24 @@ export class Message {
...
    */
-  static async encryptSessionKey(sessionKey, algorithm, aeadAlgorithm, publicKeys, passwords, wildcard = false, date = new Date(), userIds = []) {
+  static async encryptSessionKey(sessionKey, algorithm, aeadAlgorithm, publicKeys, passwords, wildcard = false, date = new Date(), userIds = [], encryptionKeyIds = []) {
     const packetlist = new PacketList();
 
     if (publicKeys) {
-      const results = await Promise.all(publicKeys.map(async function(publicKey) {
-        const encryptionKey = await publicKey.getEncryptionKey(undefined, date, userIds);
+      const results = await Promise.all(publicKeys.map(async function(publicKey, index) {
+        const encryptionKey = await publicKey.getEncryptionKey(encryptionKeyIds[index], date, userIds);
```
twiss, line 372: This is also very nitpicky but I would just call this variable `i`, for symmetry with `sign` below.
```suggestion
      const results = await Promise.all(publicKeys.map(async function(publicKey, i) {
        const encryptionKey = await publicKey.getEncryptionKey(encryptionKeyIds[i], date, userIds);
```

twiss (commented):
Hey @alichry, thanks a lot for the PR! I've left some minor comments below :blush:

test/general/openpgp.js
alichry: I am often a fan of naive testing 😆 I went with this all 2-pair testing to ensure the correct order is always being executed, rather than defining a single test case that might pass because it was lucky enough to mask any logical error when it comes to selecting the corresponding signing/key id. Yes, I do believe this amount of tests are unnecessary, I will replace it with the above.

src/message.js
alichry: The array of `privateKeys` is reversed before the loop is executed, so we have to get the opposite index or also reverse the signingKeyIds array

alichry:
Thank you @twiss for the review ✌️ I will reorder the parameters locally and run the tests.

src/message.js
twiss: Ah, I see. Yeah, I'd be tempted to do something like
```js
signingKeyIds.length = privateKeys.length;
signingKeyIds.reverse();
```
beforehand, that way the symmetry between the two arrays is a bit more obvious.

src/openpgp.js
alichry: I reordered the parameters in `openpgp.encrypt` and `Message.encrypt`. Reordering the parameters in a similar fashion in `Message.sign` or `Message.encryptSessionKey` would result in failing unit tests. Refactoring must be expensive, thoughts 🤔?

src/message.js
alichry: Would you suggest leaving a comment above it to aid the reader?

twiss: I think it's fine, given that the other `reverse` is right below it, the connection should be obvious enough. If you want to make it even more clear, you could do
```js
privateKeys = Array.from(privateKeys).reverse();
signingKeyIds = Array.from(signingKeyIds);
signingKeyIds.length = privateKeys.length;
signingKeyIds.reverse();
```
(also to avoid mutating the original parameter) - and then use those variables after.

src/openpgp.js
twiss: Oh, yeah, I guess there are some tests that call `message.sign` etc directly, maybe? I think it's still fine to update these functions, and just pass `undefined` in the correct spot in the tests to make them pass. These functions are considered part of the low-level API, so it's not a gigantic deal to break them, and v5 will be a breaking release, anyway 😌

Internally within OpenPGP.js, I think there are only a few places that call `Message.sign` or `Message.encryptSessionKey`, so I think it should be manageable, but lmk if I missed something.

test/general/openpgp.js
```diff
@@ -2588,6 +2703,100 @@ amnR6g==
...
+      it('Encrypt and sign with specific encryption/signing key ids', async function () {
+        const primaryKey = await getPrimaryKey();
+        const latestEncryptionKeyId = (await primaryKey.getEncryptionKey()).getKeyId();
+        const latestSigningKeyId = (await primaryKey.getSigningKey()).getKeyId();
+        const plaintextMessage = openpgp.Message.fromText("Hello World\n");
+
+        const checkEncryptedPackets = (encryptionKeyIds, pKESKList) => {
+          pKESKList.forEach(({ publicKeyId }, i) => {
+            if (!encryptionKeyIds[i]) {
+              expect(publicKeyId.equals(latestEncryptionKeyId)).to.be.true;
```
larabr, line 2764: Is this line ever reached? Same for `checkSignatures`. Perhaps a 4th `primaryKey` is missing as input of encryption private & publicKeys?

alichry: It's not reached, I'll bring it back, but the heavy test already passed successfully when the *keyId was not specified for a public/private key

src/openpgp.js
alichry: Sounds simple enough :)

src/message.js
alichry: I would not prefer to go with that path as it's an additional unnecessary step (although it's a very cheap call). Okay to proceed ?

twiss: Ah, sure :+1:
Though then still, I think this should default to `undefined`, rather than `signingKeyIds[0]`, right? Or simply
```suggestion
      const signingKeyId = signingKeyIds[privateKeys.length - 1 - i];
      const signingKey = await privateKey.getSigningKey(signingKeyId, date, userIds);
```
should also work?

test/general/openpgp.js
twiss: Alternatively, I think you can also just remove these (now unused) lines, as it's indeed already tested elsewhere

src/message.js
alichry: Yes, this will work. This must have slipped through when I was running the test and forgot to build beforehand ✌️

test/general/openpgp.js
alichry: Okay, I removed it

src/openpgp.js
alichry: Sorry for the delay! I reordered the parameters in `message.sign`, `message.signDetached`, `createEncryptionPackets` and `openpgp.sign`

src/message.js
```diff
@@ -424,14 +424,14 @@ export class Message {
...
+   * @param  {Array<module:type/keyid>} signingKeyIds (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
    * @param  {Date} date                              (optional) override the creation time of the signature
    * @param  {Array} userIds                          (optional) user IDs to sign with, e.g. [{ name:'Steve Sender', email:'steve@openpgp.org' }]
    * @param  {Boolean} streaming                      (optional) whether to process data as a stream
-   * @param  {Array<module:type/keyid>} signingKeyIds (optional) array of key IDs to use for signing. Each signingKeyIds[i] corresponds to privateKeys[i]
    * @returns {Promise<Message>}                      new message with signed content
    * @async
    */
-  async sign(privateKeys = [], signature = null, date = new Date(), userIds = [], streaming = false, signingKeyIds = []) {
+  async sign(privateKeys = [], signature = null, signingKeyIds = [], date = new Date(), userIds = [], streaming = false) {
```
larabr, line 434: `CleartextMessage.sign` and `signDetached` need to be changed as well I think @twiss? We should add some more tests probably, since the issue was not detected.

src/openpgp.js
twiss: Thanks a lot!

src/message.js
twiss: Yes, you're right. @alrichry Sorry for the endless stream of comments, but would you mind adding this parameter there, as well? These methods are called when passing a `CleartextMessage.fromText('test')` to `openpgp.sign`, for example. After that I think it should be ready to merge :blush:

alichry: You got it @twiss 😎, no worries, it's all for quality assurance :)

twiss: Thanks a lot!

src/message.js
```diff
@@ -351,23 +352,24 @@ export class Message {
...
+   * @param  {Array<String>} passwords                    (optional) for message encryption
+   * @param  {Boolean} wildcard                           (optional) use a key ID of 0 instead of the public key IDs
+   * @param  {Date} date                                  (optional) override the date
+   * @param  {Array} userIds                              (optional) user IDs to encrypt for, e.g. [{ name:'Robert Receiver', email:'robert@openpgp.org' }]
+   * @param  {Array<module:type/keyid>} encryptionKeyIds  (optional) array of key IDs to use for encryption. Each encryptionKeyIds[i] corresponds to publicKeys[i]
+   * @returns {Promise<Message>}                          new message with encrypted content
    * @async
    */
-  static async encryptSessionKey(sessionKey, algorithm, aeadAlgorithm, publicKeys, passwords, wildcard = false, date = new Date(), userIds = []) {
+  static async encryptSessionKey(sessionKey, algorithm, aeadAlgorithm, publicKeys, passwords, wildcard = false, date = new Date(), userIds = [], encryptionKeyIds = []) {
```
twiss, line 367: Sorry, very last comment (I hope :sweat_smile:): could you reorder these too? And maybe you also want to add the parameter to `openpgp.encryptSessionKey` (in openpgp.js), for consistency?

alichry: Whoops, sorry must have missed that 😅Thank you for your thorough and impeccable review ✌️

twiss:
Thanks again for the PR! :blush:

PR #1241, opened 2021-02-18
Pass configuration to top-level functions

### LINKED ISSUE #1166: allow to pass config into functions

larabr:
Fix #1166: refactor functions to take the configuration as parameter. `openpgp.config` is now used as default configuration in top-level functions: if no configuration is specified, then `openpgp.config` is applied.
This change avoids concurrency-related issues when `openpgp.config` is modified and read by different async functions.

Other changes:
- remove `config.rsaBlinding`: blinding is now always applied to RSA decryption.
- remove `config.debug`: debugging mode can be enabled by setting `process.env.NODE_ENV = 'development'`
- remove `config.useNative`: native crypto is always used when available

TODO:
- [x] remove config.useNative
- [x] add config tests for top-level functions
- [x] all functions in the public API should default to `openpgp.config`
- [x] update docs
- [x] update TS

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -9,15 +9,15 @@
 
 /* ############## v5 KEY #################### */
 
-export function readKey(options: { armoredKey: string }): Promise<Key>;
-export function readKey(options: { binaryKey: Uint8Array }): Promise<Key>;
-export function readKeys(options: { armoredKeys: string }): Promise<Key[]>;
-export function readKeys(options: { binaryKeys: Uint8Array }): Promise<Key[]>;
+export function readKey(options: { armoredKey: string, config?: PartialConfig }): Promise<Key>;
+export function readKey(options: { binaryKey: Uint8Array, config?: PartialConfig }): Promise<Key>;
+export function readKeys(options: { armoredKeys: string, config?: PartialConfig }): Promise<Key[]>;
+export function readKeys(options: { binaryKeys: Uint8Array, config?: PartialConfig }): Promise<Key[]>;
 export function generateKey(options: KeyOptions): Promise<KeyPair>;
-export function generateSessionKey(options: { publicKeys: Key[], date?: Date, toUserIds?: UserID[] }): Promise<SessionKey>;
-export function decryptKey(options: { privateKey: Key; passphrase?: string | string[]; }): Promise<Key>;
-export function encryptKey(options: { privateKey: Key; passphrase?: string | string[]; }): Promise<Key>;
-export function reformatKey(options: { privateKey: Key; userIds?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; }): Promise<KeyPair>;
+export function generateSessionKey(options: { publicKeys: Key[], date?: Date, toUserIds?: UserID[], config?: PartialConfig }): Promise<SessionKey>;
+export function decryptKey(options: { privateKey: Key; passphrase?: string | string[]; config?: PartialConfig }): Promise<Key>;
+export function encryptKey(options: { privateKey: Key; passphrase?: string | string[]; config?: PartialConfig }): Promise<Key>;
+export function reformatKey(options: { privateKey: Key; userIds?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; config?: PartialConfig }): Promise<KeyPair>;
 
 export class Key {
   constructor(packetlist: PacketList<AnyPacket>);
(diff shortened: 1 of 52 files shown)


src/packet/packetlist.js
```diff
@@ -42,7 +42,7 @@ class PacketList extends Array {
               const packet = packets.newPacketFromTag(tag, allowedPackets);
               packet.packets = new PacketList();
               packet.fromStream = util.isStream(parsed.packet);
-              await packet.read(parsed.packet, streaming);
+              await packet.read(parsed.packet, parsed.tag === enums.packet.compressedData ? streaming : config);
```
larabr, line 45: @twiss to avoid this selection here, it's probably best to add a dummy `streaming` parameter to all `read` functions, even though it's currently used by just one function...?

twiss: Maybe, as a temporary measure, we could pass
```suggestion
              await packet.read(parsed.packet, config, streaming);
```
here, instead of the other way around? Even though it's a bit ugly and normally I'm a stickler for consistency, but maybe in the future we can get rid of the `streaming` parameter from `CompressedDataPacket#read` (this packet is a bit of an anomaly since it decompresses already inside the `read` function, but maybe we could change that at some point), and then it will just be `packet, config` for all of them.

larabr: TBH I also don't like having `read` depend on the configuration -- the only two packets that need that are UserId (for `config.maxUseridLength`) and Signatures (for `config.knownNotations`), so we could also consider setting the relevant preferences in the constructors, even though they are not really part of the packet state either.

twiss: Myeah. For Signature packets, we could fix this by moving the `knownNotations` checking to `verify` instead of during parsing - that might be better anyway. And maybe we could do something similar for `maxUseridLength`. But tbh I would clean this up in a different PR, and not make any functional changes other than to the config here.

larabr: Ok 👍 I think it would be reasonable to apply `config.maxUseridLength` to any new UserID instance, not just the one parsed through `read`.

openpgp.d.ts
```diff
@@ -297,31 +298,33 @@ export class Message<T extends MaybeStream<Data>> {
...
+  revocationsExpire: boolean;
+  tolerant: boolean;
+  versionString: string;
+  commentString: string;
+  keyserver: string;
+  nodeStore: string;
+  allowInsecureDecryptionWithSigningKeys: boolean;
+  v5Keys: boolean;
+}
+export var config: Config;
```
larabr, line 323: I think it'd make sense to rename `openpgp.config` to `openpgp.defaultConfig` given the changes.

twiss: I think it might be confusing, as `openpgp.defaultConfig` would in turn also have a default :s And `openpgp.config` is still something that you can change, not just a description of the defaults, so I would keep it as is. I think it's pretty obvious that the configs passed to the functions are more specific and override these ones.

twiss (commented):
LGTM :+1: Small nitpicks below

src/packet/packetlist.js
```diff
@@ -30,7 +30,7 @@ class PacketList extends Array {
    * Reads a stream of binary data and interprets it as a list of packets.
    * @param {Uint8Array | ReadableStream<Uint8Array>} bytes A Uint8Array of bytes.
    */
-  async read(bytes, allowedPackets, streaming) {
+  async read(bytes, allowedPackets, streaming, config = defaultConfig) {
```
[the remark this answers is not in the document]
larabr: In most cases, `streaming` is still preceded by optional parameters, so when passing down the `config` to lower-level functions, those values still need to be specified, hence I don't think the patch would be substantially cleaner/simpler in the end (or what did you mean?).

About the current `packet.read` change where `streaming` is last, I'd really consider that to be temporary, given the corresponding discussion in https://github.com/openpgpjs/openpgpjs/pull/1241#discussion_r578686693 .

I also prefer to keep `config` as last param, first to avoid breaking (potentially most) existing code that uses the public API, but also because I see it as the param that has the most marginal effect.

src/cleartext.js
```diff
@@ -69,11 +70,12 @@ export class CleartextMessage {
...
    * @param  {Date} date                       (optional) The creation time of the signature that should be created
    * @param  {Array} userIds                   (optional) user IDs to sign with, e.g. [{ name:'Steve Sender', email:'steve@openpgp.org' }]
+   * @param  {Object} config                   (optional) full configuration, defaults to openpgp.config
    * @returns {Promise<module:cleartext.CleartextMessage>} new cleartext message with signed content
    * @async
    */
-  async sign(privateKeys, signature = null, signingKeyIds = [], date = new Date(), userIds = []) {
-    return new CleartextMessage(this.text, await this.signDetached(privateKeys, signature, signingKeyIds, date, userIds));
+  async sign(privateKeys, signature = null, signingKeyIds = [], date = new Date(), userIds = [], streaming, config = defaultConfig) {
+    return new CleartextMessage(this.text, await this.signDetached(privateKeys, signature, signingKeyIds, date, userIds, undefined, config));
```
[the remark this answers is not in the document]
larabr: Mmm how about keeping different function signatures and remove the `streaming` param only in `CleartextMessage`? After all, in `openpgp.sign` we already check whether the input is cleartext or not, so we could also do different function calls. The way it is right now, where the function call is common but the `streaming` input is just discarded for `CleartextMessage` is not-so-easy to follow. Plus, a `Message` and `CleartextMessage` are fairly different objects anyway, so I don't think a (fully) compatible interface is warranted.

I also prefer to keep `config` as last param everywhere, as I mentioned below.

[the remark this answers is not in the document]
twiss: Fair enough, yeah that also works :+1:

src/util.js
```diff
@@ -28,6 +28,7 @@
 import stream from 'web-stream-tools';
 import { getBigInteger } from './biginteger';
 
+// eslint-disable-line no-process-env
```
twiss, line 31: If you want it on a separate line I think you have to do
```suggestion
// eslint-disable-next-line no-process-env
```

PR #1244, opened 2021-02-20
Add convenient methods to Key and UserID Packet w/ typedefs

### LINKED ISSUE #1243: Convenient methods to query the key for capabilities and other information
When working with keys, users might find it convenient to use built-in functions in the OpenPGPjs library for some use cases:

1. Determine the key capabilities by inspecting the direct or binding signature. This method would return either the flag in the signature, or an array of `enum.keyFlag` for each set bit.
2. Get the latest direct or binding signature directly to make it easy for users trying to call low-level methods of OpenPGPjs.
3. Methods to indicate whether a key can be used for encryption or signing, e.g. `canEncrypt` and `canSign` regardless of the validity of the key. 
4. Get the primary key's user ids in the form of `{ name, email, comment }`. i.e. returning the userid packets directly, if data corruption might be an issue since the objects are returned, we can create a new object.

I have written some workarounds to achieve the above, but I believe it would be helpful to have some of those methods provided natively. Now, I would like to say that I am not an expert in OpenPGP(js)?, so please bear with me if some of the implications I made are incorrect 😊

I would be happy to submit a PR to fulfill a subset of these convenient functions.

Related issue: https://github.com/FlowCrypt/flowcrypt-browser/issues/3444

alichry:
Fixes #1243 
- [x] userid.getName()
- [x] userid.getEmail()
- [x] userid.getComment()
- [x] Key.getLatestValidSignature(subkey, date)
- [x] typedef: Keyid.fromId(hex)
- [x] Key.getDirectKeySignature() - Is this helpful and appropriate?
- [x] Exported crypto (openpgp.crypto) to make it available similar to v4. I need this to access `generateParamters`
- [x] SubKey.getCapabilities()

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -49,6 +49,8 @@ export class Key {
   public getCreationTime(): Date;
   public getAlgorithmInfo(): AlgorithmInfo;
   public getKeyId(): Keyid;
+  public getLatestValidSignature(subKey: SubKey, date?: Date): Promise<SignaturePacket>;
+  public getDirectKeySignature(): SignaturePacket;
 }
 
 export class SubKey {
@@ -419,6 +421,9 @@ export class OnePassSignaturePacket extends BasePacket {
 export class UserIDPacket extends BasePacket {
   public tag: enums.packet.userID;
   public userid: string;
+  public getName(): string | undefined;
+  public getComment(): string | undefined;
+  public getEmail(): string | undefined;
   static fromObject(userId: UserID): UserIDPacket;
 }
 
@@ -634,6 +639,7 @@ interface KeyOptions {
 declare class Keyid {
   bytes: string;
   equals(keyid: Keyid, matchWildcard?: boolean): boolean;
+  static fromId(hex: string): Keyid;
   toHex(): string;
 }
 
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -44,6 +44,7 @@ export class Key {
   public getEncryptionKey(keyid?: Keyid, date?: Date | null, userId?: UserID): Promise<Key | SubKey>;
   public getSigningKey(keyid?: Keyid, date?: Date | null, userId?: UserID): Promise<Key | SubKey>;
   public getKeys(keyId?: Keyid): (Key | SubKey)[];
+  public getSubkeys(keyId?: Keyid): SubKey[];
   public isDecrypted(): boolean;
   public getFingerprint(): string;
   public getCreationTime(): Date;
@@ -51,6 +52,7 @@ export class Key {
   public getKeyId(): Keyid;
   public getLatestValidSignature(subKey: SubKey, date?: Date): Promise<SignaturePacket>;
   public getDirectKeySignature(): SignaturePacket;
+  public getSubkeyCapabilities(subKey: SubKey, date?: Date): Promise<enums.keyFlags>
 }
 
 export class SubKey {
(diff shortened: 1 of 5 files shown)


twiss:
Hey @alichry :wave: Thanks for the PR. However:
- `crypto` was unexported deliberately. The reasoning was explained here: https://github.com/openpgpjs/openpgpjs/pull/1175
- Regarding `userid.getName()` and co, I'm not sure that this is better than `userid.name`. The downside of the latter is that it is mutable, and users of the library might expect this to have some effect (even though it doesn't), however, this remains true when adding the function without removing the property. When using TypeScript, obviously we can nudge people to use `getName` instead of `name` (by not exporting typings for the latter), but we could also add a `readonly name: string` type. I would honestly prefer that.
- Similarly for `getCapabilities`; I'm not sure whether this is better than a readonly property that they can binary-and with a constant.
- Re. `getLatestValidSignature`, `getDirectKeySignature`; these are very low-level functions, I would rather solve whatever use-case these serve at a higher level. For example, preferences in direct-key signatures are not taken into account by OpenPGP.js at all at the moment, we should probably do so (if this is important for some cases), before exporting low-level functions that deal with them.

The TypeScript definitions are of course useful and welcome :+1: And I think we could do more there to highlight how we expect the API to be used 😌

alichry:
Hola @twiss ✌️

Yes, I agree that updating the typescript definitions is helpful. I was driven to write `get(Name|Email|Comment)` due to missing typedefs. We can replace it with read-only members, it is appropriate and yields less code-lines to maintain. For providing similar functionality of `getSubkeyCapabilities`, how would we access the read-only property `keyFlags` of the subkey without using `getLatestValidSignature` ? 

Regarding `crypto`, it does make sense from an API point of view, but personally I thought that by using OpenPGP.js's RSA routines, it would be best compared to other libraries. Can we export it with a notice for no-BC guarantee in any minor or major release?

twiss:
> For providing similar functionality of getSubkeyCapabilities, how would we access the read-only property keyFlags of the subkey without using getLatestValidSignature ?

Yeah, that's a fair point. Maybe we could return the binding signature from this function: https://github.com/openpgpjs/openpgpjs/blob/e1307b88d0b96ff7656d00310e0ccd5f05b25dcc/src/key/subkey.js#L74 which already does most of the work. That way, we could also simplify both calling functions (`getSigningKey` and `getEncryptionKey`).

That being said, there is already a high-level way to check whether a key can sign or encrypt:

```js
const canSign = await key.getSigningKey(subkey.getKeyId(), null).then(() => true, () => false);
const canEncrypt = await key.getEncryptionKey(subkey.getKeyId(), null).then(() => true, () => false);
```

> I thought that by using OpenPGP.js's RSA routines, it would be best compared to other libraries

This isn't really the case. The variant of RSA that OpenPGP.js uses (RSASSA-PKCS1-v1_5) is determined by the OpenPGP specification, but isn't necessarily the best. If you don't need compatibility with OpenPGP, it would be better to use RSA-OAEP, or not use RSA at all.

> Can we export it with a notice for no-BC guarantee in any minor or major release?

No, sorry. The other reasons in the linked PR still apply, such as thwarting tree-shaking. In general, we want to move away from "namespace-type" exports, as they aren't a very good fit for ES6 modules. (If anything, it should be a separate module, but I don't think it's a good candidate either for the reasons mentioned above.)

alichry:
I see. Let me close this one and start over with only typescript definitions and returning `bindingSignature` in `Subkey.verify`

twiss:
👍 Thanks! 🙏

PR #1246, opened 2021-02-23
Update README.md

jmetev1:


twiss:
https://github.com/openpgpjs/openpgpjs/discussions/1232#discussioncomment-357519

PR #1259, opened 2021-03-03
Check critical notations during signature verification instead of parsing

larabr:
Changes:
- Signature parsing functions won't throw if a critical unknown notation is encountered, regardless of `config.tolerant`.
- `signature.verify` will always throw instead. Hence, verifying a message with a signature that includes unknown critical notations will now return an invalid signature, rather than no signature.

### DIFF
--- a/src/packet/signature.js
+++ b/src/packet/signature.js
@@ -434,14 +434,10 @@ class SignaturePacket {
         const name = util.uint8ArrayToStr(bytes.subarray(mypos, mypos + m));
         const value = bytes.subarray(mypos + m, mypos + m + n);
 
-        this.rawNotations.push({ name, humanReadable, value });
+        this.rawNotations.push({ name, humanReadable, value, critical });
 
         if (humanReadable) {
           this.notations[name] = util.uint8ArrayToStr(value);
-        }
-
-        if (critical && (config.knownNotations.indexOf(name) === -1)) {
-          throw new Error("Unknown critical notation: " + name);
         }
         break;
       }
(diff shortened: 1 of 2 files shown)


src/packet/signature.js
```diff
@@ -674,6 +669,12 @@ class SignaturePacket {
       throw new Error('Public key algorithm used to sign signature does not match issuer key algorithm.');
     }
 
+    this.rawNotations.forEach(({ name, critical }) => {
+      if (critical && (config.knownNotations.indexOf(name) < 0)) {
+        throw new Error(`Unknown critical notation: ${name}`);
+      }
+    });
```
twiss, line 676: Most of the similar checks are at the end of this function. That way, we throw the "most serious" error first. I would do the same here (so that we don't throw this error, which sounds relatively mild, instead of "Signature verification failed", for example).

test/general/signature.js
```diff
@@ -1046,36 +1020,18 @@ bwM=
...
-      expect(e.message).to.equal('Unknown critical notation: test@example.com');
-    } finally {
-      openpgp.config.tolerant = true;
-    }
-    // fail the test if execution does not throw an exception
-    expect(testFailed).to.be.true;
+    const message = await openpgp.readMessage({ armoredMessage: signature_with_critical_notation });
+    const key = await openpgp.readKey({ armoredKey: pub_key_arm2 });
+    const { signatures: [sig] } = await openpgp.verify({ message, publicKeys: key });
+    expect(sig.error).to.match(/Unknown critical notation: test@example.com/);
```
twiss, line 1026: Just to be extra cautious, maybe add
```suggestion
    expect(sig.valid).to.be.false;
    expect(sig.error).to.match(/Unknown critical notation: test@example.com/);
```

twiss (commented):
LGTM :+1: Small nits below

src/packet/signature.js
larabr: Right, however it seems a waste to do signature verification if we know it's still invalid?

twiss: I agree, it's inefficient, but it just means that we will be showing an error slower :) I don't think it's a super important case to optimize.

PR #1261, opened 2021-03-03
Rename `config.ignoreMdcError`, drop `config.integrityProtect` and allow V4 keys to be AEAD-encrypted

larabr:
Changes:
- rename `config.ignoreMdcError` to `config.allowUnauthenticatedMessages`
- remove `config.integrityProtect`: it is no longer possible to create messages without integrity protection
- `config.aeadProtect` is now used when selecting the encryption mechanism of keys (previously, only v5 keys were encrypted using AEAD)

### DIFF
--- a/src/key/factory.js
+++ b/src/key/factory.js
@@ -185,11 +185,9 @@ async function wrapKeyObject(secretKeyPacket, secretSubkeyPackets, options, conf
     signaturePacket.features = [0];
     signaturePacket.features[0] |= enums.features.modificationDetection;
     if (config.aeadProtect) {
-      signaturePacket.features || (signaturePacket.features = [0]);
       signaturePacket.features[0] |= enums.features.aead;
     }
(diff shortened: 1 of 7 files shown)


src/key/factory.js
```diff
@@ -181,10 +181,9 @@ async function wrapKeyObject(secretKeyPacket, secretSubkeyPackets, options, conf
...
     }
-    if (config.integrityProtect) {
-      signaturePacket.features = [0];
-      signaturePacket.features[0] |= enums.features.modificationDetection;
-    }
+    // integrity protection always enabled
+    signaturePacket.features = [0];
+    signaturePacket.features[0] |= enums.features.modificationDetection;
     if (config.aeadProtect) {
       signaturePacket.features || (signaturePacket.features = [0]);
```
twiss, line 188: This is very nitpicky but you can remove this line now (also the one in the next `if`)

PR #1262, opened 2021-03-04
Simplify algorithm preference selection and uniform config names

larabr:
Changes:
- rename `config.compression` into `config.preferredCompressionAlgorithm`
- rename `config.encryptionCipher` into `config.preferredCipherAlgorithm`
- rename `config.preferHashAlgorithm` into `config.preferredHashAlgorithm`
- rename `config.aeadMode` into `config.preferredAEADAlgorithm`
- when encrypting to public keys, the compression/aead/cipher algorithm is selected by:
  * taking the preferred algorithm specified in config, if it is supported by all recipients
  * otherwise, taking the "MUST implement" algorithm specified by the OpenPGP spec
- when encrypting to passphrases only (no public keys), the preferred algorithms from `config` are always used
- EdDSA signing with hash weaker than sha256 is no longer allowed (https://tools.ietf.org/id/draft-ietf-openpgp-rfc4880bis-10.html#section-15-7.2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,20 +278,20 @@ It's possible to change that behaviour by enabling compression through the confi
     const encrypted = await openpgp.encrypt({
         message,
         passwords: ['secret stuff'], // multiple passwords possible
-        config: { preferredCompressionAlgorithm: openpgp.enums.compression.zip } // compress the data with zip
+        config: { preferredCompressionAlgorithm: openpgp.enums.compression.zlib } // compress the data with zlib
     });
 })();
 ```
 
 or by changing the default global configuration:
 ```js
-openpgp.config.preferredCompressionAlgorithm = openpgp.enums.compression.zip
+openpgp.config.preferredCompressionAlgorithm = openpgp.enums.compression.zlib
 ```
 
 Where the value can be any of:
  * `openpgp.enums.compression.zip`
  * `openpgp.enums.compression.zlib`
- * `openpgp.enums.compression.uncompressed`
+ * `openpgp.enums.compression.uncompressed` (default)
 
 
 
--- a/README.md
+++ b/README.md
@@ -79,9 +79,9 @@ library to convert back and forth between them.
   You can change the AEAD mode by setting one of the following options:
 
   ```
-  openpgp.config.preferredAEADAlgorithm = openpgp.enums.aead.eax // Default, native
-  openpgp.config.preferredAEADAlgorithm = openpgp.enums.aead.ocb // Non-native
-  openpgp.config.preferredAEADAlgorithm = openpgp.enums.aead.experimentalGcm // **Non-standard**, fastest
+  openpgp.config.preferredAeadAlgorithm = openpgp.enums.aead.eax // Default, native
+  openpgp.config.preferredAeadAlgorithm = openpgp.enums.aead.ocb // Non-native
+  openpgp.config.preferredAeadAlgorithm = openpgp.enums.aead.experimentalGcm // **Non-standard**, fastest
   ```
 
 * For environments that don't provide native crypto, the library falls back to [asm.js](https://caniuse.com/#feat=asmjs) implementations of AES, SHA-1, and SHA-256.
--- a/README.md
+++ b/README.md
@@ -79,9 +79,9 @@ library to convert back and forth between them.
   You can change the AEAD mode by setting one of the following options:
 
   ```
-  openpgp.config.aeadMode = openpgp.enums.aead.eax // Default, native
-  openpgp.config.aeadMode = openpgp.enums.aead.ocb // Non-native
-  openpgp.config.aeadMode = openpgp.enums.aead.experimentalGcm // **Non-standard**, fastest
+  openpgp.config.preferredAEADAlgorithm = openpgp.enums.aead.eax // Default, native
+  openpgp.config.preferredAEADAlgorithm = openpgp.enums.aead.ocb // Non-native
+  openpgp.config.preferredAEADAlgorithm = openpgp.enums.aead.experimentalGcm // **Non-standard**, fastest
   ```
 
 * For environments that don't provide native crypto, the library falls back to [asm.js](https://caniuse.com/#feat=asmjs) implementations of AES, SHA-1, and SHA-256.
@@ -169,7 +169,7 @@ Here are some examples of how to use OpenPGP.js v5. For more elaborate examples 
 
 #### Encrypt and decrypt *Uint8Array* data with a password
 
-Encryption will use the algorithm specified in config.encryptionCipher (defaults to aes256), and decryption will use the algorithm used for encryption.
+Encryption will use the algorithm specified in config.preferredCipherAlgorithm (defaults to aes256), and decryption will use the algorithm used for encryption.
 
 ```js
 (async () => {
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ Either set the `compression` parameter in the options object when calling `encry
     const encrypted = await openpgp.encrypt({
         message,
         passwords: ['secret stuff'], // multiple passwords possible
-        compression: openpgp.enums.compression.zip // compress the data with zip
+        preferredCompressionAlgorithm: openpgp.enums.compression.zip // compress the data with zip
     });
 })();
 ```
@@ -287,7 +287,7 @@ Either set the `compression` parameter in the options object when calling `encry
 Or, override the config to enable compression:
 
 ```js
-openpgp.config.compression = openpgp.enums.compression.zlib;
+openpgp.config.preferredCompressionAlgorithm = openpgp.enums.compression.zlib;
 ```
 
 Where the value can be any of:
--- a/README.md
+++ b/README.md
@@ -267,11 +267,10 @@ Encrypt with multiple public keys:
 })();
 ```
 
-#### Encrypt with compression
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -279,15 +279,15 @@ Either set the `compression` parameter in the options object when calling `encry
     const encrypted = await openpgp.encrypt({
         message,
         passwords: ['secret stuff'], // multiple passwords possible
-        compression: openpgp.enums.compression.zip // compress the data with zip
+        preferredCompressionAlgorithm: openpgp.enums.compression.zip // compress the data with zip
```
twiss, line 282: This should be inside the config

src/key/factory.js
```diff
@@ -134,17 +134,8 @@ async function wrapKeyObject(secretKeyPacket, secretSubkeyPackets, options, conf
...
-        const configIndex = algos.indexOf(configAlgo);
-        if (configIndex >= 1) { // If it is included and not in first place,
-          algos.splice(configIndex, 1); // remove it.
-        }
-        if (configIndex !== 0) { // If it was included and not in first place, or wasn't included,
-          algos.unshift(configAlgo); // add it to the front.
-        }
-      }
-      return algos;
+    function createdPreferredAlgos(algos, preferredAlgo) {
```
twiss, line 137: I know this is my typo, but 😅
```suggestion
    function createPreferredAlgos(algos, preferredAlgo) {
```

src/key/factory.js
```diff
@@ -161,23 +152,23 @@ async function wrapKeyObject(secretKeyPacket, secretSubkeyPackets, options, conf
...
       enums.hash.sha512
-    ], config.preferHashAlgorithm);
+    ], config.preferredHashAlgorithm);
     signaturePacket.preferredCompressionAlgorithms = createdPreferredAlgos([
+      enums.compression.uncompressed,
       enums.compression.zlib,
-      enums.compression.zip,
-      enums.compression.uncompressed
-    ], config.compression);
+      enums.compression.zip
```
twiss, line 170: This is very nitpicky, but if someone sets `preferredCompressionAlgorithm = zlib`, this would result in `zlib, uncompressed, zip`, which is slightly strange, so I would actually leave the order here as-is, and leave it up to `createPreferredAlgos` to hoist `uncompressed` to the front, if needed.

src/key/helper.js
```diff
@@ -159,43 +159,40 @@ export async function getPreferredHashAlgo(key, keyPacket, date = new Date(), us
...
-          algo !== enums[type].idea && // not implemented
-          enums.read(enums[type], algo) && // known algorithm
-          count === keys.length && // available for all keys
-          prio > prefAlgo.prio) {
-        prefAlgo = prioMap[algo];
-      }
-    } catch (e) {}
-  });
-  return prefAlgo.algo;
+  return (senderAlgoSupport.length > 0 && senderAlgoSupport.every(Boolean)) ? preferredSenderAlgo : defaultAlgo;
```
twiss, line 195: I think you could just
```suggestion
  return senderAlgoSupport.every(Boolean) ? preferredSenderAlgo : defaultAlgo;
```
and then always call this function, rather than only calling it if there are public keys?

larabr: `[].every(Boolean)` returns true

README.md
```diff
@@ -267,32 +267,32 @@ Encrypt with multiple public keys:
...
+It's possible to change that behaviour by enabling compression through the config, either for the single encryption:
 
 ```js
 (async () => {
     const message = openpgp.Message.fromBinary(new Uint8Array([0x01, 0x02, 0x03])); // or .fromText('string')
     const encrypted = await openpgp.encrypt({
         message,
         passwords: ['secret stuff'], // multiple passwords possible
-        compression: openpgp.enums.compression.zip // compress the data with zip
+        config: { preferredCompressionAlgorithm: openpgp.enums.compression.zip } // compress the data with zip
```
twiss, line 281: I would prefer that we recommend `zlib` here (and below)

README.md
```diff
@@ -267,32 +267,32 @@ Encrypt with multiple public keys:
...
 ```js
-openpgp.config.compression = openpgp.enums.compression.zlib;
+openpgp.config.preferredCompressionAlgorithm = openpgp.enums.compression.zip
 ```
 
 Where the value can be any of:
  * `openpgp.enums.compression.zip`
  * `openpgp.enums.compression.zlib`
+ * `openpgp.enums.compression.uncompressed`
+
```
twiss, line 295:
```suggestion
 * `openpgp.enums.compression.uncompressed` (the default)
```
? :man_shrugging:

src/openpgp.js
```diff
@@ -264,7 +264,10 @@ export function encrypt({ message, publicKeys, privateKeys, passwords, sessionKe
     if (privateKeys.length || signature) { // sign the message only if private keys or signature is specified
       message = await message.sign(privateKeys, signature, signingKeyIds, date, fromUserIds, message.fromStream, config);
     }
-    message = message.compress(config);
+    message = message.compress(
+      await getPreferredAlgo('compression', publicKeys || [], date, toUserIds, config),
```
twiss, line 268: This is very nitpicky but maybe you could make `publicKeys` an optional argument, defaulting to `[]`, rather than including a default here?

src/message.js
```diff
@@ -479,16 +479,18 @@ export class Message {
   }
 
   /**
-   * Compresses the message (the literal and -if signed- signature data packets of the message)
+   * Compresses the message (the literal and -if signed ß- signature data packets of the message)
```
twiss, line 482: I see you're starting to integrate into German, inserting random German ~sentences~ letters in your work :upside_down_face:

src/crypto/public_key/elliptic/eddsa.js
```diff
@@ -24,13 +24,15 @@
...
+import enums from '../../../enums';
+import hash from '../../hash';
 
 nacl.hash = bytes => new Uint8Array(sha512().update(bytes).digest());
 
 /**
  * Sign a message using the provided key
  * @param {module:type/oid} oid - Elliptic curve object identifier
- * @param {module:enums.hash} hash_algo - Hash algorithm used to sign
+ * @param {module:enums.hash} hash_algo - Hash algorithm used to sign (sha256 or stronger)
```
twiss, line 35:
```suggestion
 * @param {module:enums.hash} hash_algo - Hash algorithm used to sign (must be sha256 or stronger)
```

src/message.js
```diff
@@ -479,16 +479,18 @@ export class Message {
   }
 
   /**
-   * Compresses the message (the literal and -if signed- signature data packets of the message)
+   * Compresses the message (the literal and -if signed - signature data packets of the message)
```
twiss, line 482:
```suggestion
   * Compresses the message (the literal and -if signed- signature data packets of the message)
```

src/key/helper.js
```diff
@@ -159,43 +159,41 @@ export async function getPreferredHashAlgo(key, keyPacket, date = new Date(), us
...
+ * @returns {module:enums.symmetric|aead|compression} Preferred algorithm
  * @async
  */
-export async function getPreferredAlgo(type, keys, date = new Date(), userIds = [], config = defaultConfig) {
-  const prefProperty = type === 'symmetric' ? 'preferredSymmetricAlgorithms' : 'preferredAeadAlgorithms';
-  const defaultAlgo = type === 'symmetric' ? enums.symmetric.aes128 : enums.aead.eax;
-  const prioMap = {};
-  await Promise.all(keys.map(async function(key, i) {
+export async function getPreferredAlgo(type, keys = [], date = new Date(), userIds = [], config = defaultConfig) {
+  const defaultAlgo = { // these are all must-implement
```
twiss, line 172:
```suggestion
  const defaultAlgo = { // these are all must-implement in rfc4880bis
```

twiss (commented):
Increasingly nitpicky nitpicks

src/config/config.js
```diff
@@ -24,19 +24,19 @@ import enums from '../enums';
...
    */
-  preferHashAlgorithm: enums.hash.sha256,
+  preferredHashAlgorithm: enums.hash.sha256,
   /**
    * @memberof module:config
-   * @property {Integer} encryptionCipher Default encryption cipher {@link module:enums.symmetric}
+   * @property {Integer} preferredCipherAlgorithm Default encryption cipher {@link module:enums.symmetric}
    */
-  encryptionCipher: enums.symmetric.aes256,
+  preferredCipherAlgorithm: enums.symmetric.aes256,
```
twiss, line 34: Sorry, last nitpick - probably this should be `preferredSymmetricAlgorithm`, for consistency with `preferredSymmetricAlgorithms`?

larabr: I disagree because AEAD is also symmetric, I think the existing 'symmetric' stuff should be renamed instead, however they are not all internal, so perhaps it risks breaking too much?

twiss: Well `preferredSymmetricAlgorithms` is named that way because of the name of the subpacket in the spec.. I don't think we can rename that. And - if you say "symmetric encryption algorithm", people think of AES, not EAX, no?

larabr: Right, it's unfortunate they haven't updated the packet name (yet? :D) -- I don't find "symmetric encryption algorithm" especially descriptive, so not sure what other people would think :sweat_smile: I would still consider renaming the `enums` though, since also the spec do talk about "ciphers" here and there

PR #1265, opened 2021-03-12
Drop support for verification of detached cleartext signatures

larabr:
When verifying cleartext messages, it is no longer possible to pass the signature separately -- this is because a cleartext message in itself should include the signature.
In the past, we allowed to detach-sign a cleartext message, so we have kept supporting verifying detached cleartext signatures up until now.  However, the concept of "detached cleartext signature" does not make sense.

In practice, given a signed cleartext message:
```js
const cleartextMessage = await openpgp.sign({
    message: openpgp.CleartextMessage.fromText('test      '),
    privateKeys
});
```
it should be verified as:
```js
 const verified = await openpgp.verify({
      message: await openpgp.readCleartextMessage({ cleartextMessage }),
      publicKeys
});
```

**With this PR, the following is no longer supported:**
```js 
const verified = await openpgp.verify({
    message: openpgp.CleartextMessage.fromText('test      '),
    publicKeys,
   // cannot pass signature with cleartext message
    signature: (await openpgp.readCleartextMessage({ cleartextMessage })).signature
})
```
if you have a detached signature over some cleartext data (coming from previous openpgpjs versions), you can still verify such signature by passing a `Message` generated from the text data, with trailing spaces removed:
```js
const verified = await openpgp.verify({
      message: openpgp.Message.fromText(removeTrailingSpaces('test      ')),
      publicKeys,
      signature: (await openpgp.readCleartextMessage({ cleartextMessage })).signature
})
```

### DIFF
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -18,6 +18,8 @@ jobs:
       with:
         poll-interval-seconds: 30
         abort-after-seconds: 900
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - uses: actions/setup-node@v1
     - run: npm ci
     - run: npm run build-test
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -14,6 +14,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: softprops/turnstyle@v0.1.5
+      with:
+        poll-interval-seconds: 30
+        abort-after-seconds: 900
     - uses: actions/setup-node@v1
     - run: npm ci
     - run: npm run build-test
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - uses: softprops/turnstyle@v0.1.5
+    - uses: softprops/turnstyle@v1
       with:
         poll-interval-seconds: 30
         abort-after-seconds: 900
@@ -34,6 +34,12 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: softprops/turnstyle@v1
+      with:
+        poll-interval-seconds: 30
+        abort-after-seconds: 900
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - uses: actions/setup-node@v1
     - run: npm ci
     - run: npm run build-test --lightweight
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -66,25 +66,10 @@ export class CleartextMessage {
    * @async
    */
   async sign(privateKeys, signature = null, signingKeyIds = [], date = new Date(), userIds = [], config = defaultConfig) {
-    return new CleartextMessage(this.text, await this.signDetached(privateKeys, signature, signingKeyIds, date, userIds, config));
-  }
-
-  /**
-   * Sign the cleartext message
-   * @param {Array<Key>} privateKeys - private keys with decrypted secret key data for signing
-   * @param {Signature} [signature] - Any existing detached signature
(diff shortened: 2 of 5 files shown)


tomholub:
Hello - is there an issue associated that clarifies this change? This seems like something we have a usecase for, but I may have missed something.

larabr:
> Hello - is there an issue associated that clarifies this change? This seems like something we have a usecase for, but I may have missed something.

I'll add details later

twiss:
For additional context, in OpenPGP.js v3, if you did:

```js
await openpgp.sign({
    data: 'test      ',
    privateKeys,
    detached: true
});
```

then it would internally construct a cleartext message, and detached-sign it. Logically speaking, this doesn't make much sense, since a detached cleartext signature doesn't exist, it's just a detached signature. Practically speaking, it doesn't much matter; the only difference between a cleartext signature and a normal signature is that for the former, we strip trailing spaces from each line (the spec requires this for cleartext signatures, but not for detached signatures - or text signatures in general).

In OpenPGP.js v4, we switched to passing messages instead of data (this was necessary for streaming support, since when passing a stream, it's not obvious whether it contains text or binary data, hence the `message.fromText/fromBinary`), so then you instead had to do:

```js
await openpgp.sign({
    message: openpgp.cleartext.fromText('test      '),
    privateKeys,
    detached: true
});
```

(if you wanted to keep the same behavior of removing trailing spaces - otherwise, `openpgp.message.fromText` would also work).

In OpenPGP.js v5, to keep the same behavior, you'll have to do:

```js
await openpgp.sign({
    message: openpgp.Message.fromText(removeTrailingSpaces('test      ')),
    privateKeys,
    detached: true
});
```

(with `removeTrailingSpaces` copied from [here](https://github.com/openpgpjs/openpgpjs/blob/master/src/util.js#L645-L651), for example).

This is more verbose, but at least now it's obvious what's going on. In principle, we'd prefer people just use `openpgp.Message.fromText` directly - there's no particular reason to remove trailing spaces, except if you need to be backwards compatible with this quirk from OpenPGP.js v3.

The above change has already been done in the v5 branch, this PR does the same for `openpgp.verify`.

twiss:
For posterity: the original example code was unfortunately not quite complete in all cases. I've updated the PR description above with code examples and an explanation of the cases in which it's needed.

For additional details, read on:

Creating a cleartext message also normalized the newlines to `\r\n`. Verifying a text signature does so as well (also on a normal message), but verifying a binary signature does not. For signatures created by OpenPGP.js, this doesn't matter, as OpenPGP.js always creates text signatures for text messages. However, when verifying signatures from other implementations, if a binary signature was created over text with `\r\n` newlines, and the text was then modified to use `\n` newlines (during transport, for example), and the signature would then get verified by OpenPGP.js on the modified text, this would work in v4 but break with the original example code above.

Unfortunately, the behavior prior to v3.0.9 was again different. Prior to that version, OpenPGP.js would also remove trailing spaces in lines with `\r\n` newlines, but from v3.0.9 to v5.3.1 (inclusive), only trailing spaces before `\n` were removed, due to a bug which was fixed in #1548. (In v5, however, this is only relevant for signing, as verifying detached signatures on cleartext messages is no longer supported, as per this PR :))

All of this is mostly only relevant for PGP/MIME signatures, which should have trailing spaces stripped and newlines normalized to `\r\n`. There, the behavior prior to v3.0.9 is thus correct. So, applications verifying PGP/MIME signatures should, in principle, have already had custom code to strip trailing spaces and normalize newlines, in which case all of the above wouldn't cause issues. But, it's possible that, in cases where the application only handled removing trailing spaces but not normalizing newlines, *or* normalized newlines to `\n`, the above would cause bugs. Don't ask me how I know :)

Apologies for the confusion!

(The reason we went to all this trouble, though, and the main advantage of having all this logic spelled out in the application, rather than hidden away in OpenPGP.js, is that this logic cleanly maps from the PGP/MIME specification, [RFC 3156](https://www.rfc-editor.org/rfc/rfc3156.html#page-5), which is separate from the OpenPGP specification, and having this logic in the application makes it easier to verify that the application matches the behavior required for its specific use-case.)

PR #1267, opened 2021-03-15
Type definitions

sinderw:
- Added `key.write()`, returning binary representation of a Key.
- Added type definitions for `key.write()`, `signature.write()`, and `message.write()` to avoid dealing with packets.
- Added type definitions for key certification and verification

Should we hide `key.keyPacket`, `signature.packets` and `message.packets` from the type definitions?

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -25,7 +25,7 @@ export class Key {
   public subKeys: SubKey[];
   public users: User[];
   public revocationSignatures: SignaturePacket[];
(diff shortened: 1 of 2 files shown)


twiss:
Thanks for the PR!


> Should we hide `key.keyPacket`, `signature.packets` and `message.packets` from the type definitions?

Yeah, I think we could mark them as `private` :+1:

sinderw:
Right, I marked them private 👍

twiss:
:+1: Thanks!

PR #1269, opened 2021-03-21
Add in missing readToEnd in streaming example

willfarrell:
Currently the example for `Streaming sign and verify *Uint8Array* data` returns `Error: signature could not be verified`.


twiss:
Hey :wave: Thanks for the PR! Could you remove the first `readToEnd` and restore the second one? Then it should also work, and then the verification is also streaming.

willfarrell:
Merged upstream doc changes and added in suggested changes.

twiss:
Thanks! :+1:

PR #1275, opened 2021-03-26
Add `expectSigned` option to `openpgp.decrypt` and `openpgp.verify`

### LINKED ISSUE #1274: Add `expectSigned` option to enforce signature verification in openpgp.decrypt
`openpgp.decrypt` does not throw if signatures are missing or do not verify. Instead, one needs to inspect the returned `{ signatures }` object for errors/validity. This behaviour makes sense as a message is not necessarily signed, or the recipient might not have the (correct) public keys to verify the signature.
However, it can result in verification errors being missed. To help avoid such issues, as well as simplify dealing with messages that are expected to be signed, we could add an `expectSigned : Boolean` option that - if set - results in `decrypt` throwing an error and not returning any data if the message signatures are missing or invalid.

larabr:
Fix #1274 , changes: 
- add `expectSigned` option to `openpgp,decrypt` and `openpgp.verify` (defaulting to false)
- if `expectSigned = true`:
  - `openpgp.decrypt` throws immediately if public keys or signatures are missing, or if the decrypted data is not streamed & signatures are invalid
  - `openpgp.verify` throws immediately if signatures are missing, or if the decrypted data is not streamed & signatures are invalid
  - if the data is streamed, and signatures are invalid, reading the data stream will eventually throw

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,8 +296,8 @@ If you expect an encrypted message to be signed with one of the public keys you 
     const { data: decrypted, signatures } = await openpgp.decrypt({
         message,
         privateKeys: privateKey // for decryption
-        expectSign: true,
-        publicKeys: publicKey, // for verification (mandatory with expectSign=true)
+        expectSigned: true,
+        publicKeys: publicKey, // for verification (mandatory with expectSigned=true)
     });
     console.log(decrypted); // 'Hello, World!'
 })();
--- a/README.md
+++ b/README.md
@@ -225,16 +225,17 @@ const openpgp = require('openpgp'); // use as CommonJS, AMD, ES6 module or via w
     const message = await openpgp.readMessage({
         armoredMessage: encrypted // parse armored message
     });
-    const { data: decrypted } = await openpgp.decrypt({
+    const { data: decrypted, signatures } = await openpgp.decrypt({
         message,
         publicKeys: publicKey, // for verification (optional)
         privateKeys: privateKey // for decryption
     });
     console.log(decrypted); // 'Hello, World!'
+    console.log(signatures[0].valid) // signature validity (signed messages only)
 })();
 ```
 
-Encrypt with multiple public keys:
+Encrypt to multiple public keys:
 
 ```js
 (async () => {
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -267,6 +268,41 @@ Encrypt with multiple public keys:
...
+
+    const message = await openpgp.readMessage({
+        armoredMessage: encryptedAndSignedMessage // parse armored message
+    });
+    // decryption will fail if all signatures are invalid or missing
+    const { data: decrypted, signatures } = await openpgp.decrypt({
+        message,
+        privateKeys: privateKey // for decryption
+        expectSign: true,
+        publicKeys: publicKey, // for verification (mandatory with expectSign=true)
```
twiss, line 300:
```suggestion
        expectSigned: true,
        publicKeys: publicKey, // for verification (mandatory with expectSigned=true)
```

src/openpgp.js
```diff
@@ -311,6 +312,21 @@ export function decrypt({ message, privateKeys, passwords, sessionKeys, publicKe
     result.data = format === 'binary' ? decrypted.getLiteralData() : decrypted.getText();
     result.filename = decrypted.getFilename();
     linkStreams(result, message);
+    if (expectSigned) {
+      if (publicKeys.length === 0) {
+        throw new Error('public keys are required to verify message signatures');
```
twiss, line 317: This is very nitpicky but please capitalize these error messages
```suggestion
        throw new Error('Public keys are required to verify message signatures');
```
(also the other ones)

src/openpgp.js
```diff
@@ -311,6 +312,21 @@ export function decrypt({ message, privateKeys, passwords, sessionKeys, publicKe
...
+      }
+      if (result.signatures.length === 0) {
+        throw new Error('message is not signed');
+      }
+      result.data = stream.concat([
+        result.data,
+        stream.fromAsync(() => util.anyPromise(
+          result.signatures.map(sig => sig.verified)
+        ).then(() => {}) // append nothing to the data
+          .catch(() => { throw new Error('invalid message signature'); }))
```
twiss, line 327: I know this was my suggestion but probably it's cleaner to write it like
```suggestion
        stream.fromAsync(async () => {
          await util.anyPromise(result.signatures.map(sig => sig.verified));
        })
```

src/openpgp.js
larabr: But these errors are wrapped and appear after "Error decrypting message:"

twiss: Yeah, but the same thing is true for most other errors in this library, and we capitalize those

twiss (approved):

PR #1276, opened 2021-03-27
Fixed createMessage return type

sinderw:
The `createMessage` function is async (see [`message.js:838`](https://github.com/openpgpjs/openpgpjs/blob/eb496d2018f8f6a8957a5022f1fd3b606aa39c7a/src/message.js)), so its type definition should return a `Promise`


twiss:
Whoops, thanks for catching that!

PR #1285, opened 2021-04-09
Remove `Key.prototype.encrypt()` and `Key.prototype.decrypt()`

larabr:
Changes:
- remove `Key.prototype.encrypt` and `decrypt`, since these methods could leave the key object in an inconsistent state in case of errors, or when used in combination with `key.addSubkey()`.
- `openpgp.generateKey` now requires `options.userIDs`, since otherwise the key is basically unusable.

To encrypt/decrypt a key, the top-level functions `openpgp.encryptKey` and `openpgp.decryptKey` should be used instead: on success, these return a new encrypted/decrypted key object, and do not affect the original key.

Note that the `keyID` parameter is not supported by `encryptKey`/`decryptKey`, since partial key decryption is not recommended. If you still want to decrypt a single subkey or primary key `k`, you can call `k.keyPacket.decrypt(...)`, followed by `k.keyPacket.validate(...)`. Similarly, for encryption do `k.keyPacket.encrypt(...)`.

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -28,8 +28,6 @@ export class Key {
   private keyPacket: PublicKeyPacket | SecretKeyPacket;
   public write(): Uint8Array;
   public armor(config?: Config): string;
-  public decrypt(passphrase: string | string[], keyID?: KeyID, config?: Config): Promise<void>; // throws on error
-  public encrypt(passphrase: string | string[], keyID?: KeyID, config?: Config): Promise<void>; // throws on error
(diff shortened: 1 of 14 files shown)


openpgp.d.ts
```diff
@@ -28,8 +28,6 @@ export class Key {
   private keyPacket: PublicKeyPacket | SecretKeyPacket;
```
larabr, line 28: We need to declare this as `public` if we want to allow for single key/subkey decryption

twiss (approved):

PR #1288, opened 2021-04-20
Updated wrong documentation examples

thomaskokholm:
When looking at the code the old `openpgp.createMessage({ text : "Some text"})` has been replace with `openpgp.Message.fromText("Some text")`  as well as the binary method which is now `Message.fromBinary()`.
I updated the examples in Readme to reflect this.


larabr:
The README documents OpenPGP.js v5, which is not officially released yet. For the documentation of the latest released version (v4.10.10) , see https://github.com/openpgpjs/openpgpjs/tree/v4.10.10#readme .

PR #1289, opened 2021-04-20
Make `PacketList` a valid subtype of `Array` and update `Packet.tag` type accessor

### LINKED ISSUE #907: filterByTag check never throws due to List() truthyness
At [message.js:158](https://github.com/openpgpjs/openpgpjs/blob/master/src/message.js#L158) there is:
`    const symESKeyPacketlist = this.packets.filterByTag(enums.packet.symEncryptedSessionKey);
    if (!symESKeyPacketlist) {
      throw new Error('No symmetrically encrypted session key packet found.');
    }
`
However the implementation [filterByTag](https://github.com/openpgpjs/openpgpjs/blob/master/src/packet/packetlist.js#L154) always returns a `new List()` (which may be empty).

Even when empty, `    if (!symESKeyPacketlist) {` always evaluates false because empty lists are still `truthy`

There may be other places in the code that follows this pattern...

larabr:
Changes:
- Implementation:
  - Remove `PacketList.prototype.concat` and `push` (we solely rely on `Array.push` instead).
  - Fix #907 by correctly handling result of `filterByTag`.
  - Implement `write()` method for `Trust` and `Marker` packets, to make them compatible with the `BasePacket` interface
- Types:
  - Simplified and updated `PacketList` type definitions
  - Fix type accessor for `Packet.tag`, which is `static` since https://github.com/openpgpjs/openpgpjs/pull/1268
  - Prevent passing SubkeyPackets where KeyPackets are expected, and vice versa

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -294,7 +294,7 @@ interface PartialConfig extends Partial<Config> {}
 /* ############## v5 PACKET #################### */
 
 declare abstract class BasePacket {
-  public tag: enums.packet;
+  static tag: enums.packet;
   public read(bytes: Uint8Array): void;
   public write(): Uint8Array;
 }
@@ -319,11 +319,11 @@ declare abstract class BasePublicKeyPacket extends BasePacket {
 }
 
 export class PublicKeyPacket extends BasePublicKeyPacket {
-  public tag: enums.packet.publicKey;
+  static tag: enums.packet.publicKey;
 }
 
 export class PublicSubkeyPacket extends BasePublicKeyPacket {
-  public tag: enums.packet.publicSubkey;
+  static tag: enums.packet.publicSubkey;
 }
 
 declare abstract class BaseSecretKeyPacket extends BasePublicKeyPacket {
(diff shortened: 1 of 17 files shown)


src/packet/packetlist.js
```diff
@@ -186,6 +190,8 @@ class PacketList extends Array {
 
   /**
    * Concatenates packetlist or array of packets
+   * @param {PacketList|Packet} packets - packets to concatenate
+   * @returns {PacketList} this packetlist
```
larabr, line 194: I think we should also consider renaming this `concat` method, or return a new `PacketList` instance (like `Array.concat` does).

openpgp.d.ts
```diff
@@ -336,56 +336,76 @@ declare abstract class BaseSecretKeyPacket extends BasePublicKeyPacket {
 }
 
 export class SecretKeyPacket extends BaseSecretKeyPacket {
-  public tag: enums.packet.secretKey;
+  static readonly tag: enums.packet.secretKey;
 }
 
 export class SecretSubkeyPacket extends BaseSecretKeyPacket {
-  public tag: enums.packet.secretSubkey;
+  static readonly tag: enums.packet.secretSubkey;
```
larabr, line 343: We might want to add a private dummy field `isSubkey()` to the key packets, to prevent passing a subkey when expecting a primary key, and vice versa -- the current classes have compatible structure and TS can't detect the difference.

src/key/key.js
```diff
@@ -150,10 +150,10 @@ class Key {
   toPacketlist() {
     const packetlist = new PacketList();
     packetlist.push(this.keyPacket);
-    packetlist.concat(this.revocationSignatures);
-    packetlist.concat(this.directSignatures);
-    this.users.map(user => packetlist.concat(user.toPacketlist()));
-    this.subKeys.map(subKey => packetlist.concat(subKey.toPacketlist()));
+    packetlist.append(this.revocationSignatures);
```
twiss, line 153: `push` accepts multiple arguments, so you could replace this with
```suggestion
    packetlist.push(...this.revocationSignatures);
```
(and similarly everywhere else). Then you don't need a special method for this

test/security/unsigned_subpackets.js
```diff
@@ -81,7 +81,7 @@ async function makeKeyValid() {
   pusersig.readSubPackets(fake.writeHashedSubPackets(), false);
   // reconstruct the modified key
   const newlist = new PacketList();
-  newlist.concat([pubkey, puser, pusersig]);
+  newlist.push(...[pubkey, puser, pusersig]);
```
twiss, line 84: This is then very nitpicky but you can just do:
```suggestion
  newlist.push(pubkey, puser, pusersig);
```

test/security/subkey_trust.js
```diff
@@ -56,7 +56,7 @@ async function testSubkeyTrust() {
   fakeBindingSignature.keyFlags = [enums.keyFlags.signData];
   await fakeBindingSignature.sign(attackerPrivKey.keyPacket, dataToSign);
   const newList = new PacketList();
-  newList.concat([
+  newList.push(...[
```
twiss, line 59: And similarly here

test/security/unsigned_subpackets.js
larabr: Yes this is fixed in the next PR :upside_down_face:

twiss (approved):

PR #1300, opened 2021-05-06
Unexport `Key` and introduce `PrivateKey` and `PublicKey` classes

### LINKED ISSUE #1293: Introduce `PrivateKey` and `PublicKey` classes
The current `Key` class could be extended to have `PrivateKey` and `PublicKey` subclasses, thus allowing TS to statically check key type. Also, given the changes proposed in https://github.com/openpgpjs/openpgpjs/pull/1205#issuecomment-772888464, having distinct subclasses would allow to easily document where passing a public key is sufficient.

larabr:
Close #1293 , changes: 
- add `PrivateKey` and `PublicKey` classes. A `PrivateKey` can always be passed where a `PublicKey` key is expected, but not vice versa.
- unexport `Key`, and export `PrivateKey` and `PublicKey`. 
- rename `Key.packetlist2structure` to `Key.packetListToStructure`
- change `Key.update` to return a new updated key, rather than modifying the destination one in place
- add `openpgp.readPrivateKey` and `openpgp.readPrivateKeys` to avoid having to downcast the result of `readKey(s)` in TS.

TODO:
- [x] Fix/simplify `Key.update`
- [x] Update function docs, after merging #1299

### DIFF
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -679,12 +679,16 @@ export default Key;
  * Creates a PublicKey or PrivateKey depending on the packetlist in input
  * @param {PacketList} - packets to parse
  * @return {Key} parsed key
+ * @throws if no key packet was found
  */
 export function createKey(packetlist) {
-  switch (packetlist[0].constructor.tag) {
-    case enums.packet.secretKey:
-      return new PrivateKey(packetlist);
-    case enums.packet.publicKey:
-      return new PublicKey(packetlist);
+  for (const packet of packetlist) {
+    switch (packet.constructor.tag) {
+      case enums.packet.secretKey:
+        return new PrivateKey(packetlist);
+      case enums.packet.publicKey:
+        return new PublicKey(packetlist);
+    }
   }
+  throw new Error('No key packet found');
 }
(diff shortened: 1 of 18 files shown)


src/key/key.js
```diff
@@ -906,3 +675,16 @@ class Key {
 
 export default Key;
 
+/**
+ * Creates a PublicKey or PrivateKey depending on the packetlist in input
+ * @param {PacketList} - packets to parse
+ * @return {Key} parsed key
+ */
+export function createKey(packetlist) {
+  switch (packetlist[0].constructor.tag) {
```
twiss, line 684: I don't feel very strongly about this, but technically speaking it's legal for the first packet to be a Marker packet, for example, and the second packet to be the key packet. Currently we would accept that, I think, this would break that (though I don't think it's very likely to happen in the wild, but it might be more robust to check if there's any private key packet in the list, for example).

src/key/key.js
```diff
@@ -906,3 +675,16 @@ class Key {
...
+    case enums.packet.secretKey:
+      return new PrivateKey(packetlist);
+    case enums.packet.publicKey:
+      return new PublicKey(packetlist);
+  }
```
twiss, line 689: Also, it might be good to throw here if neither type of key packet is present.

Issue #880, opened 2019-04-01
refactor async / Promise loop(s)


tomholub:
There's seems to be a bit of silliness in the way the following (and potentially a few more) piece of code is written at https://github.com/openpgpjs/openpgpjs/blob/master/src/key.js#L563

```js
  await Promise.all(users.map(async function (a) {
    return a.user.revoked || a.user.isRevoked(primaryKey, a.selfCertification, null, date);
  }));
```
Firstly, `async` keyword is not needed there (may even affect error propagation as written - unsure).
```js
  await Promise.all(users.map(function (a) {
    return a.user.revoked || a.user.isRevoked(primaryKey, a.selfCertification, null, date);
  }));
```
Secondly, what does it even do? It seems to want to make sure to go through the `isRevoked` (unless already known to be revoked) promisified function before proceeding below. If that's the case, I think the intent could be made a bit clearer with:
```js
  for(const a of users.filter(a => !a.user.revoked)) { // evaluate revocations if any
    await a.user.isRevoked(primaryKey, a.selfCertification, null, date);
  }
```
I'm writing it here instead of making a PR because I'm unsure if there are other motivations. One would think if this is cpu bound and not IO bound, `Promise.all` should not bring a performance boost as opposed to doing it sequentially? (not tested)

twiss:
> If this is cpu bound and not IO bound, Promise.all should not bring a performance boost as opposed to doing it sequentially? (not tested)

Benchmarks are of course always welcome, but the theory behind why this could be faster is that `user.isRevoked` calls `packet.signature.verify`, which could use Web Crypto to hash the data, which could use multiple cores. (To be fair, for the specific code you quoted this is probably unlikely to actually happen, as the default `config.min_bytes_for_web_crypto` is probably too high for it.)

Nevertheless, this code pattern is used in many other places in the codebase, to indicate that we want to evaluate multiple async functions, and that they don't have to wait for each other. I'd argue that that information by itself is valuable to have, to make the intention of the code clear. Also, if we ever somehow implement signature verification in Web Crypto, we get a performance boost for free here.

Finally, it might make this code clearer to define a helper function for this, something like:

```js
  await util.each(users, a => {
    return a.user.revoked || a.user.isRevoked(primaryKey, a.selfCertification, null, date);
  });
```

```js
  util.each = (list, fn) => Promise.all(list.map(fn));
```

This might then make it less explicit that the intention was to execute them concurrently, though.

tomholub:
[...]

PR #1302, opened 2021-05-12
Fix various signature verification issues

### LINKED ISSUE #1129: missing test for expired signatures
This code does not seem to have any equivalent test that triggers this error (unless I overlooked it).

https://github.com/openpgpjs/openpgpjs/blob/e29de76dc1d86a25b3fe6d20c7dc21c78e8c9ec2/src/message.js#L655-L663

Adding a test would help ensure the behavior doesn't regress in the future.

// note: I came to this conclusion by searching `Signature is expired` in the codebase / tests.

### LINKED ISSUE #1137: Signatures without signature creation time subpackets are invalid
The signature creation time subpacket "MUST be present in the hashed area".  Knowing the signature's creation time is crucial to signature verification.

https://tools.ietf.org/html/rfc4880#section-5.2.3.4
https://tests.sequoia-pgp.org/#Detached_signature_with_Subpackets

Reproducer:

The following signature over the data `huhu` contains no signature creation time subpacket:

```
-----BEGIN PGP SIGNATURE-----

wsDtBAABCAAXFiEE0aZuGiOxgsmYD3iM+/zIKgFeczAACgkQ+/zIKgFeczDjiwv+
LFUWJohCYtauaVDHBDHWF+tojls+ducY6uuU6iUTBb1969okh2sjUmvPwIjrVXuk
cfPl616xRqVWolEU9T5sG6MjRlAaG31Oo/FVAVFXZn30ldEtuDss12+/IhES3Wfx
M1jGdJZ1ZMZJpRsNBJXBegEBFKbPleEd66seuuFfvoIUbgsdj7IT/ZlEMlixelWW
igXPVY1C05oPbkC8oo0lVSxwdq6gDvm8a52k3GCtXJELrYGH29C+eDqmyLP1zJOt
NBoZBAqMd9XYVrJtuip436D9pdo5pbg4zCE6uPf2zzx4taK7jGkk6nn7LqVDxvQm
3dAXUnIxw4V9eL3V8SFAKwmouUmHPRbjfnQ70hxYQxDXUcIwu1aYn13QS1s/F/jf
DVRZWaAhNdL9BfHfhEsRVsmjMhe0zwRpaepvXnERbnA/lAUHEmEvgfPFz/2GsAo/
kCNcH9WI6idSzFjuYegECf+ZA1xOCjS9oLTGbSeT7jNfC8dH5+E92qlBLq4Ctt7k
=lMU7
-----END PGP SIGNATURE-----
```

```
-----BEGIN PGP PUBLIC KEY BLOCK-----
Comment: Bob's OpenPGP certificate

mQGNBF2lnPIBDAC5cL9PQoQLTMuhjbYvb4Ncuuo0bfmgPRFywX53jPhoFf4Zg6mv
/seOXpgecTdOcVttfzC8ycIKrt3aQTiwOG/ctaR4Bk/t6ayNFfdUNxHWk4WCKzdz
/56fW2O0F23qIRd8UUJp5IIlN4RDdRCtdhVQIAuzvp2oVy/LaS2kxQoKvph/5pQ/
5whqsyroEWDJoSV0yOb25B/iwk/pLUFoyhDG9bj0kIzDxrEqW+7Ba8nocQlecMF3
X5KMN5kp2zraLv9dlBBpWW43XktjcCZgMy20SouraVma8Je/ECwUWYUiAZxLIlMv
9CurEOtxUw6N3RdOtLmYZS9uEnn5y1UkF88o8Nku890uk6BrewFzJyLAx5wRZ4F0
qV/yq36UWQ0JB/AUGhHVPdFf6pl6eaxBwT5GXvbBUibtf8YI2og5RsgTWtXfU7eb
SGXrl5ZMpbA6mbfhd0R8aPxWfmDWiIOhBufhMCvUHh1sApMKVZnvIff9/0Dca3wb
vLIwa3T4CyshfT0AEQEAAbQhQm9iIEJhYmJhZ2UgPGJvYkBvcGVucGdwLmV4YW1w
bGU+iQHOBBMBCgA4AhsDBQsJCAcCBhUKCQgLAgQWAgMBAh4BAheAFiEE0aZuGiOx
gsmYD3iM+/zIKgFeczAFAl2lnvoACgkQ+/zIKgFeczBvbAv/VNk90a6hG8Od9xTz
XxH5YRFUSGfIA1yjPIVOnKqhMwps2U+sWE3urL+MvjyQRlyRV8oY9IOhQ5Esm6DO
ZYrTnE7qVETm1ajIAP2OFChEc55uH88x/anpPOXOJY7S8jbn3naC9qad75BrZ+3g
9EBUWiy5p8TykP05WSnSxNRt7vFKLfEB4nGkehpwHXOVF0CRNwYle42bg8lpmdXF
DcCZCi+qEbafmTQzkAqyzS3nCh3IAqq6Y0kBuaKLm2tSNUOlZbD+OHYQNZ5Jix7c
ZUzs6Xh4+I55NRWl5smrLq66yOQoFPy9jot/Qxikx/wP3MsAzeGaZSEPc0fHp5G1
6rlGbxQ3vl8/usUV7W+TMEMljgwd5x8POR6HC8EaCDfVnUBCPi/Gv+egLjsIbPJZ
ZEroiE40e6/UoCiQtlpQB5exPJYSd1Q1txCwueih99PHepsDhmUQKiACszNU+RRo
zAYau2VdHqnRJ7QYdxHDiH49jPK4NTMyb/tJh2TiIwcmsIpGuQGNBF2lnPIBDADW
ML9cbGMrp12CtF9b2P6z9TTT74S8iyBOzaSvdGDQY/sUtZXRg21HWamXnn9sSXvI
DEINOQ6A9QxdxoqWdCHrOuW3ofneYXoG+zeKc4dC86wa1TR2q9vW+RMXSO4uImA+
Uzula/6k1DogDf28qhCxMwG/i/m9g1c/0aApuDyKdQ1PXsHHNlgd/Dn6rrd5y2AO
baifV7wIhEJnvqgFXDN2RXGjLeCOHV4Q2WTYPg/S4k1nMXVDwZXrvIsA0YwIMgIT
86Rafp1qKlgPNbiIlC1g9RY/iFaGN2b4Ir6GDohBQSfZW2+LXoPZuVE/wGlQ01rh
827KVZW4lXvqsge+wtnWlszcselGATyzqOK9LdHPdZGzROZYI2e8c+paLNDdVPL6
vdRBUnkCaEkOtl1mr2JpQi5nTU+gTX4IeInC7E+1a9UDF/Y85ybUz8XV8rUnR76U
qVC7KidNepdHbZjjXCt8/Zo+Tec9JNbYNQB/e9ExmDntmlHEsSEQzFwzj8sxH48A
EQEAAYkBtgQYAQoAIBYhBNGmbhojsYLJmA94jPv8yCoBXnMwBQJdpZzyAhsMAAoJ
EPv8yCoBXnMw6f8L/26C34dkjBffTzMj5Bdzm8MtF67OYneJ4TQMw7+41IL4rVcS
KhIhk/3Ud5knaRtP2ef1+5F66h9/RPQOJ5+tvBwhBAcUWSupKnUrdVaZQanYmtSx
cVV2PL9+QEiNN3tzluhaWO//rACxJ+K/ZXQlIzwQVTpNhfGzAaMVV9zpf3u0k14i
tcv6alKY8+rLZvO1wIIeRZLmU0tZDD5HtWDvUV7rIFI1WuoLb+KZgbYn3OWjCPHV
dTrdZ2CqnZbG3SXw6awH9bzRLV9EXkbhIMez0deCVdeo+wFFklh8/5VK2b0vk/+w
qMJxfpa1lHvJLobzOP9fvrswsr92MA2+k901WeISR7qEzcI0Fdg8AyFAExaEK6Vy
jP7SXGLwvfisw34OxuZr3qmx1Sufu4toH3XrB7QJN8XyqqbsGxUCBqWif9RSK4xj
zRTe56iPeiSJJOIciMP9i2ldI+KgLycyeDvGoBj0HCLO3gVaBe4ubVrj5KjhX2PV
NEJd3XZRzaXZE2aAMQ==
=NXei
-----END PGP PUBLIC KEY BLOCK-----
```

### LINKED ISSUE #1138: Expiration of primary key binding signatures is not honored
OpenPGPjs does not honor the expiration time on primary key
binding signatures.

Reproducer:

This is a signature over the string "Hello World :)":

```
-----BEGIN PGP SIGNATURE-----

wsDzBAABCgAGBYJfLAFsACEJEHwvqk35PDeyFiEEHdzhXwkhfO4vOzdgfC+qTfk8
N7KiqwwAts4QGB7v9bABCC2qkTxJhmStC0wQMcHRcjL/qAiVnmasQWmvE9KVsdm3
AaXd8mIx4a37/RRvr9dYrY2eE4uw72cMqPxNja2tvVXkHQvk1oEUqfkvbXs4ypKI
NyeTWjXNOTZEbg0hbm3nMy+Wv7zgB1CEvAsEboLDJlhGqPcD+X8a6CJGrBGUBUrv
KVmZr3U6vEzClz3DBLpoddCQseJRhT4YM1nKmBlZ5quh2LFgTSpajv5OsZheqt9y
EZAPbqmLhDmWRQwGzkWHKceKS7nZ/ox2WK6OS7Ob8ZGZkM64iPo6/EGj5Yc19vQN
AGiIaPEGszBBWlOpHTPhNm0LB0nMWqqaT87oNYwP8CQuuxDb6rKJ2lffCmZH27Lb
UbQZcH8J+0UhpeaiadPZxH5ATJAcenmVtVVMLVOFnm+eIlxzov9ntpgGYt8hLdXB
ITEG9mMgp3TGS9ZzSifMZ8UGtHdp9QdBg8NEVPFzDOMGxpc/Bftav7RRRuPiAER+
7A5CBid5
=aQkm
-----END PGP SIGNATURE-----
```

Made by the subkey of this certificate:

```
-----BEGIN PGP PUBLIC KEY BLOCK-----

xsDNBF2lnPIBDAC5cL9PQoQLTMuhjbYvb4Ncuuo0bfmgPRFywX53jPhoFf4Zg6mv
/seOXpgecTdOcVttfzC8ycIKrt3aQTiwOG/ctaR4Bk/t6ayNFfdUNxHWk4WCKzdz
/56fW2O0F23qIRd8UUJp5IIlN4RDdRCtdhVQIAuzvp2oVy/LaS2kxQoKvph/5pQ/
5whqsyroEWDJoSV0yOb25B/iwk/pLUFoyhDG9bj0kIzDxrEqW+7Ba8nocQlecMF3
X5KMN5kp2zraLv9dlBBpWW43XktjcCZgMy20SouraVma8Je/ECwUWYUiAZxLIlMv
9CurEOtxUw6N3RdOtLmYZS9uEnn5y1UkF88o8Nku890uk6BrewFzJyLAx5wRZ4F0
qV/yq36UWQ0JB/AUGhHVPdFf6pl6eaxBwT5GXvbBUibtf8YI2og5RsgTWtXfU7eb
SGXrl5ZMpbA6mbfhd0R8aPxWfmDWiIOhBufhMCvUHh1sApMKVZnvIff9/0Dca3wb
vLIwa3T4CyshfT0AEQEAAc0hQm9iIEJhYmJhZ2UgPGJvYkBvcGVucGdwLmV4YW1w
bGU+wsEABBMBCgATBYJeO2eVAgsJAxUICgKbAQIeAQAhCRD7/MgqAV5zMBYhBNGm
bhojsYLJmA94jPv8yCoBXnMwKWUMAJ3FKZfJ2mXvh+GFqgymvK4NoKkDRPB0CbUN
aDdG7ZOizQrWXo7Da2MYIZ6eZUDqBKLdhZ5gZfVnisDfu/yeCgpENaKib1MPHpA8
nZQjnPejbBDomNqY8HRzr5jvXNlwywBpjWGtegCKUY9xbSynjbfzIlMrWL4S+Rfl
+bOOQKRyYJWXmECmVyqY8cz2VUYmETjNcwC8VCDUxQnhtcCJ7Aej22hfYwVEPb/J
BsJBPq8WECCiGfJ9Y2y6TF+62KzG9Kfs5hqUeHhQy8V4TSi479ewwL7DH86XmIIK
chSANBS+7iyMtctjNZfmF9zYdGJFvjI/mbBR/lK66E515Inuf75XnL8hqlXuwqvG
ni+i03Aet1DzULZEIio4uIU6ioc1lGO9h7K2Xn4S7QQH1QoISNMWqXibUR0RCGjw
FsEDTt2QwJl8XXxoJCooM7BCcCQo+rMNVUHDjIwrdoQjPld3YZsUQQRcqH6bLuln
cfn5ufl8zTGWKydoj/iTz8KcjZ7w187AzQRdpZzyAQwA1jC/XGxjK6ddgrRfW9j+
s/U00++EvIsgTs2kr3Rg0GP7FLWV0YNtR1mpl55/bEl7yAxCDTkOgPUMXcaKlnQh
6zrlt6H53mF6Bvs3inOHQvOsGtU0dqvb1vkTF0juLiJgPlM7pWv+pNQ6IA39vKoQ
sTMBv4v5vYNXP9GgKbg8inUNT17BxzZYHfw5+q63ectgDm2on1e8CIRCZ76oBVwz
dkVxoy3gjh1eENlk2D4P0uJNZzF1Q8GV67yLANGMCDICE/OkWn6daipYDzW4iJQt
YPUWP4hWhjdm+CK+hg6IQUEn2Vtvi16D2blRP8BpUNNa4fNuylWVuJV76rIHvsLZ
1pbM3LHpRgE8s6jivS3Rz3WRs0TmWCNnvHPqWizQ3VTy+r3UQVJ5AmhJDrZdZq9i
aUIuZ01PoE1+CHiJwuxPtWvVAxf2POcm1M/F1fK1J0e+lKlQuyonTXqXR22Y41wr
fP2aPk3nPSTW2DUAf3vRMZg57ZpRxLEhEMxcM4/LMR+PABEBAAHCwrIEGAEKAAkF
gl8sAVYCmwIB3QkQ+/zIKgFeczDA+qAEGQEKAAwFgl47Z5UFgwB4TOAAIQkQfC+q
Tfk8N7IWIQQd3OFfCSF87i87N2B8L6pN+Tw3st58C/0exp0X2U4LqicSHEOSqHZj
jiysdqIELHGyo5DSPv92UFPp36aqjF9OFgtNNwSa56fmAVCD4+hor/fKARRIeIjF
qdIC5Y/9a4B10NQFJa5lsvB38x/d39LI2kEoglZnqWgdJskROo3vNQF4KlIcm6FH
dn4WI8UkC5oUUcrpZVMSKoacIaxLwqnXT42nIVgYYuqrd/ZagZZjG5WlrTOd5+NI
zi/l0fWProcPHGLjmAh4Thu8i7omtVw1nQaMnq9I77ffg3cPDgXknYrLL+q8xXh/
0mEJyIhnmPwllWCSZuLv9DrD5pOexFfdlwXhf6cLzNpW6QhXD/Tf5KrqIPr9aOv8
9xaEEXWh0vEby2kIsI2++ft+vfdIyxYw/wKqx0awTSnuBV1rG3z1dswX4BfoY66x
Bz3KOVqlz9+mG/FTRQwrgPvR+qgLCHbuotxoGN7fzW+PI75hQG5JQAqhsC9sHjQH
UrI21/VUNwzfw3v5pYsWuFb5bdQ3ASJetICQiMy7IW8WIQTRpm4aI7GCyZgPeIz7
/MgqAV5zMG6/C/wLpPl/9e6Hf5wmXIUwpZNQbNZvpiCcyx9sXsHXaycOQVxn3McZ
nYOUP9/mobl1tIeDQyTNbkxWjU0zzJl8XQsDZerb5098pg+x7oGIL7M1vn5s5JMl
owROourqF88JEtOBxLMxlAM7X4hB48xKQ3Hu9hS1GdnqLKki4MqRGl4l5FUwyGOM
GjyS3TzkfiDJNwQxybQiC9n57ij20ieNyLfuWCMLcNNnZUgZtnF6wCctoq/0ZIWu
a7nvuA/XC2WW9YjEJJiWdy5109pqac+qWiY11HWy/nms4gpMdxVpT0RhrKGWq4o0
M5q3ZElOoeN70UO3OSbU5EVrG7gB1GuwF9mTHUVlV0veSTw0axkta3FGT//XfSpD
lRrCkyLzwq0M+UUHQAuYpAfobDlDdnxxOD2jm5GyTzak3GSVFfjW09QFVO6HlGp5
01/jtzkUiS6nwoHHkfnyn0beZuR8X6KlcrzLB0VFgQFLmkSM9cSOgYhD0PTu9aHb
hW1Hj9AO8lzggBQ=
=Nt+N
-----END PGP PUBLIC KEY BLOCK-----
```

Relevant test: https://tests.sequoia-pgp.org/#Primary_key_binding_signatures

larabr:
- Fix #1137 : throw on signature parsing (e.g. in `openpgp.readSignature`) if the [creation time subpacket](https://datatracker.ietf.org/doc/html/draft-ietf-openpgp-rfc4880bis-10#section-5.2.3.4) is missing from the hashed area.
- Fix #1138: `SignaturePacket.verify` now directly checks for signature expiration and signatures created in the future. This makes it easier to thoroughly check the validity of signatures.  Also:
  - `openpgp.revokeKey` now takes a `date` to check the provided revocation certificate
  - `openpgp.decryptSessionKeys` now takes a `date` to check the validity of the provided private keys
  - whenever a `date` is used internally, the function accepts a `date` param to allow for proper time sync.
- Fix #1129 by adding tests for all of the above.
- Like `openpgp.generateKey`, `openpgp.reformatKey` now also requires `options.userIDs` (follow up to #1285).
- Simplify calling `SubKey.isRevoked/update/getExpirationTime` by adding the `SubKey.mainKey` field to hold the reference of the corresponding `Key`.

Breaking changes in function signatures:
- changes to `date` param:
  - `Key.update(key, config)` -> `update(key, date, config)`
  - `Key.applyRevocationCertificate(revocationCertificate, config)` -> `applyRevocationCertificate(revocationCertificate, date, config)`
  - `Key.signAllUsers(privateKeys, config)` -> `signAllUsers(privateKeys, date, config)`
  - `Key.verifyAllUsers(keys, config)` -> `verifyAllUsers(keys, date, config)`
  - `SignaturePacket.constructor(date)` -> `constructor(void)`
  - `SignaturePacket.sign(key, data, detached)` -> `sign(key, data, date, detached)`
  - `Message.sign(primaryKey, privateKeys, config)` -> `sign(primaryKey, privateKeys, date, config)`
  - `Message.decrypt(privateKeys, passwords, sessionKeys, config)` -> `decrypt(privateKeys, passwords, sessionKeys, date, config)`
  - `Message.decryptSessionKeys(privateKeys, passwords, config)` -> `decryptSessionKeys(privateKeys, passwords, date, config)`
- removed `primaryKey` param:
  - `SubKey.isRevoked(primaryKey, signature, key, date, config)` -> `isRevoked(signature, key, date, config)`
  - `SubKey.update(subKey, primaryKey, date, config)` -> `update(subKey, date, config)`
  - `SubKey.getExpirationTime(primaryKey, date, config)` -> `getExpirationTime(date, config)`


test/general/signature.js
```diff
@@ -1493,28 +1409,6 @@ hkJiXopCSWKSlQInL1devkJJUWJmTmZeugJYlpdLAagQJM0JpsCqIQZwKgAA
     });
   });
 
-  it('Verify test with expired verification public key', async function() {
```
larabr, line 1496: I didn't get what these 2 tests were checking, given in neither case verification is expected to fail...?
I think the tests I added should cover all "expired key" cases anyway.

src/key/helper.js
```diff
@@ -269,10 +266,11 @@ export async function isDataRevoked(primaryKey, signatureType, dataToVerify, rev
         // third-party revocation signatures here. (It could also be revoking a
         // third-party key certification, which should only affect
         // `verifyAllCertifications`.)
-        (!signature || revocationSignature.issuerKeyID.equals(signature.issuerKeyID)) &&
-        !(config.revocationsExpire && revocationSignature.isExpired(normDate))
+        !signature || revocationSignature.issuerKeyID.equals(signature.issuerKeyID)
       ) {
-        revocationSignature.verified || await revocationSignature.verify(key, signatureType, dataToVerify, undefined, config);
+        revocationSignature.verified || await revocationSignature.verify(
```
larabr, line 271: Just noticed that if the revocation signature expires soon after we have verified it once, the signature is "always" considered verified regardless, due to the caching. The same issue is present wherever `.verified` is checked in the rest of the code.
I still think it's a good idea to check `isExpired` as part of `verify`, so to solve the problem, I'd rather check `.verified` inside `verify` as well, and then add a new optional parameter `reverify=true` to determine whether to discard the cached value (something similar was already proposed in #880).

larabr: I just added a test to detect this problem -- it's failing for now, as expected.

twiss: I agree moving the `.verified` check into `verify` makes sense, but I'd prefer not to have a `reverify` parameter, rather I would just consider `.verified` to only mean that the signature was cryptographically valid, but always re-check expiry (it should be very cheap to do so anyway).

src/key/key.js
```diff
@@ -525,10 +528,17 @@ class Key {
...
     // check for valid, unrevoked, unexpired self signature
     const { selfCertification } = await this.getPrimaryUser(date, userID, config);
-    // check for expiration time
+    // check for expiration time in binding signatures
     if (helper.isDataExpired(primaryKey, selfCertification, date)) {
       throw new Error('Primary key is expired');
     }
+    // check for expiration time in direct signatures
+    if (this.directSignatures.length > 0) {
+      const directSignature = await helper.getLatestValidSignature(this.directSignatures, primaryKey, enums.signature.key, { key: primaryKey }, date, config);
```
twiss, line 537: This probably causes the key to be considered invalid if the direct signature is invalid, which might be an issue if someone adds a direct signature signed by a different key to an SKS server, for example.
So probably it would be good to add a try/catch around this line.
(Though I'm a bit confused, I thought this would cause [this test](https://github.com/openpgpjs/openpgpjs/blob/02a1ed2d783b89eee79e529f08d096f32debe911/test/general/key.js#L2751) to fail, but maybe I'm missing something?)

src/message.js
```diff
@@ -721,39 +722,42 @@ async function createVerificationObject(signature, literalDataList, keys, date =
...
+  const signaturePacketPromise = isOnePassSignature ? signature.correspondingSig : signature;
+  const isSignaturePacketAvailable = !util.isPromise(signaturePacketPromise);
   if (!primaryKey) {
     keyError = new Error(`Could not find signing key with key ID ${signature.issuerKeyID.toHex()}`);
   } else {
     try {
-      signingKey = await primaryKey.getSigningKey(signature.issuerKeyID, null, undefined, config);
+      // We need the signature creation time to check whether the key was expired at the time of signing.
+      // If the signature packet is not available yet, we delay the check after verifying the signature
+      signingKey = await primaryKey.getSigningKey(
```
twiss, line 735: Could we, rather than sometimes calling `getSigningKey` twice, simply get the signing key packet from the `getKeys` call above (and store it there), and then always call `getSigningKey` below, to simplify the code (and hopefully not need `isPromise`, etc)?

src/packet/signature.js
```diff
@@ -364,6 +369,9 @@ class SignaturePacket {
     // subpacket type
     switch (type) {
       case 2:
+        if (!trusted) {
+          throw new Error('Unexpected non-hashed signature creation time');
```
twiss, line 373: It's not possible to end up here I think, see the `if` above.

src/packet/signature.js
```diff
@@ -654,19 +662,28 @@ class SignaturePacket {
...
     const publicKeyAlgorithm = enums.write(enums.publicKey, this.publicKeyAlgorithm);
     const hashAlgorithm = enums.write(enums.hash, this.hashAlgorithm);
 
     if (publicKeyAlgorithm !== enums.write(enums.publicKey, key.algorithm)) {
       throw new Error('Public key algorithm used to sign signature does not match issuer key algorithm.');
     }
 
+    const normDate = util.normalizeDate(date);
+    if (normDate && this.created > normDate) {
+      throw new Error('Signature is in the future');
```
twiss, line 681:
```suggestion
      throw new Error('Signature creation time is in the future');
```

src/packet/signature.js
```diff
@@ -660,33 +660,41 @@ class SignaturePacket {
...
+      if (this.hashed) {
+        hash = await this.hashed;
+      } else {
+        toHash = this.toHash(signatureType, data, detached);
+        hash = await this.hash(signatureType, data, toHash);
+      }
+      hash = await stream.readToEnd(hash);
+      if (this.signedHashValue[0] !== hash[0] ||
+          this.signedHashValue[1] !== hash[1]) {
+        throw new Error('Invalid signature: Message digest did not match');
```
twiss, line 686: Usually, this should be wrapped in something like "Could not find valid self-signature:" or "Message signature could not be verified:" (though some of those we still have to add at some point) and then this will be a bit redundant.

src/packet/signature.js
```diff
@@ -654,19 +662,28 @@ class SignaturePacket {
...
 
     if (publicKeyAlgorithm !== enums.write(enums.publicKey, key.algorithm)) {
       throw new Error('Public key algorithm used to sign signature does not match issuer key algorithm.');
     }
 
+    const normDate = util.normalizeDate(date);
+    if (normDate && this.created > normDate) {
+      throw new Error('Signature is in the future');
+    }
+    if (this.isExpired(date)) {
```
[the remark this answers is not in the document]
larabr: It's used by `mergeSignatures`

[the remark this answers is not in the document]
larabr: I changed `isExpired` instead, since it's used by `mergeSignatures` as well

src/packet/signature.js
```diff
@@ -654,41 +651,58 @@ class SignaturePacket {
...
-    );
-    if (!verified) {
-      throw new Error('Signature verification failed');
+      const verified = await crypto.signature.verify(
+        publicKeyAlgorithm, hashAlgorithm, this.params, key.publicParams,
+        toHash, hash
+      );
+      if (!verified) {
+        throw new Error('Signature verification failed');
+      }
```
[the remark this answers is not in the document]
larabr: I'd rather not mark a signature as verified when e.g. we consider the hash algo to be unsafe. I actually think it might be best to reset the flag once we detect that the signature is no longer valid, because we do return the signature packets as part of `signatures` in the top-level functions, and people might end up accessing `signatures[0].signature.verified`

src/key/key.js
```diff
@@ -268,6 +264,9 @@ class Key {
 
     try {
       const primaryUser = await this.getPrimaryUser(date, userID, config);
+      if (date && primaryKey.getCreationTime() > date) {
+        throw new Error('Primary key was created after the given date');
+      }
```
twiss, line 269: This seems very ad-hoc, and should be checked somewhere else, most likely in `verifyPrimaryKey` - in fact, I think it will currently already throw "Primary key is expired" in this situation?

src/key/key.js
```diff
@@ -360,47 +359,77 @@ class Key {
...
       throw new Error('Primary key is expired');
     }
+    // check for expiration time in direct signatures
+    if (this.directSignatures.length > 0) {
+      const directSignature = await helper.getLatestValidSignature(
+        this.directSignatures, primaryKey, enums.signature.key, { key: primaryKey }, date, config
+      ).catch(() => {}); // invalid signatures are discarded, to avoid breaking the key
+
+      if (directSignature && helper.isDataExpired(primaryKey, directSignature, date)) {
+        throw new Error('Primary key is expired through a direct signature');
```
twiss, line 373: This is a bit nitpicky, but this error message seems slightly too specific to me. I don't think the user really cares what kind of signature was used to make the key expire

src/key/key.js
```diff
@@ -360,47 +359,77 @@ class Key {
...
+      return Math.min(...notNull);
+    };
+
+    let primaryKeyExpiry;
+    try {
+      const { selfCertification } = await this.getPrimaryUser(null, userID, config);
+      const keyExpiry = helper.getKeyExpirationTime(this.keyPacket, selfCertification);
+      const sigExpiry = selfCertification.getExpirationTime();
+      // TODO check direct signatures
+      primaryKeyExpiry = toDateOrInfinity(Math.min(keyExpiry, sigExpiry));
```
twiss, line 404: Is this really better than `keyExpiry < sigExpiry ? keyExpiry : sigExpiry`?
And then, for the remainder of this function, I think you could just use numbers internally, and then convert back to `Date` if needed only at the very end. Then, you don't really need the helper functions. (And if you get rid of the early return, you can also use `Math.min` even here.

src/key/key.js
```diff
@@ -360,47 +359,77 @@ class Key {
...
-      const signExpiry = await signKey.getExpirationTime(this.keyPacket, undefined, config);
-      if (signExpiry < expiry) expiry = signExpiry;
+      const signingKeys = (await Promise.all(
+        this.getKeys(keyID).map(key => this.getSigningKey(key.getKeyID(), null, userID, config).catch(() => null))
+      )).filter(Boolean);
+      if (signingKeys.length === 0) return null; // the key cannot sign
+      signExpiry = await getLatestKeyExpiry(signingKeys);
     }
-    return expiry;
+    return toDateOrInfinity(minNotNull(encryptExpiry, signExpiry));
```
twiss, line 432: This is not quite sufficient, I think. Theoretically speaking, you could have a primary key for certification only which expires in 1 day, a signing subkey which expires in 2 days, and an encryption subkey which expires in 3 days. The overall expiry should still be in 1 day. That's kind of why the function was setup as it was, updating the expiry inside the ifs.

src/key/key.js
```diff
@@ -360,47 +359,77 @@ class Key {
...
+      const sigExpiry = selfCertification.getExpirationTime();
+      // TODO check direct signatures
+      primaryKeyExpiry = toDateOrInfinity(Math.min(keyExpiry, sigExpiry));
+    } catch (e) {
+      primaryKeyExpiry = null;
+    }
+    if (!capabilities) return primaryKeyExpiry;
+
+    // loop through subkeys based on requested capabilities and retrieve their expiration times
+    const getLatestKeyExpiry = async keys => (keys.length === 0 ? null : toDateOrInfinity(
```
twiss, line 411: You don't need this check for `keys.length` here, as you already check it below

src/key/key.js
```diff
@@ -360,47 +359,77 @@ class Key {
...
+      Math.max(...await Promise.all(
+        keys.map(async key => (key instanceof SubKey ? key.getExpirationTime(null, config) : primaryKeyExpiry))
```
twiss, line 413: You don't need this `async` here, as there's no `await`

src/key/key.js
```diff
@@ -593,24 +621,25 @@ class Key {
    */
   async signPrimaryUser(privateKeys, date, userID, config = defaultConfig) {
     const { index, user } = await this.getPrimaryUser(date, userID, config);
-    const userSign = await user.sign(this.keyPacket, privateKeys, config);
-    const key = this.clone();
+    const userSign = await user.sign(this.keyPacket, privateKeys, date, config);
+    const key = await this.clone();
```
twiss, line 625: Is this necessary?

src/key/subkey.js
```diff
@@ -87,37 +90,36 @@ class SubKey {
 
   /**
    * Returns the expiration time of the subkey or Infinity if key does not expire
-   * Returns null if the subkey is invalid.
-   * @param  {SecretKeyPacket|
-   *          PublicKeyPacket} primaryKey  The primary key packet
+   * Returns null if the subkey is invalid
```
twiss, line 93: This is very nitpicky but personally I prefer the version with dots, to have proper grammar, so I would prefer adding a dot to the first line rather than removing the second one.

src/key/subkey.js
```diff
@@ -61,21 +65,20 @@ class SubKey {
 
   /**
    * Verify subkey. Checks for revocation signatures, expiration time
-   * and valid binding signature.
-   * @param  {SecretKeyPacket|
-   *          PublicKeyPacket} primaryKey The primary key packet
+   * and valid binding signature
```
twiss, line 68: Also here
```suggestion
   * and valid binding signature.
```

src/message.js
```diff
@@ -697,8 +699,7 @@ export async function createSignaturePackets(literalDataPacket, signingKeys, sig
  * @param {SignaturePacket} signature - Signature packet
  * @param {Array<LiteralDataPacket>} literalDataList - Array of literal data packets
  * @param {Array<PublicKey>} verificationKeys - Array of public keys to verify signatures
- * @param {Date} date - Verify the signature against the given date,
- *                    i.e. check signature creation time < date < expiration time
+ * @param {Date} [date] - Check signature validity with respect to the given date ()
```
twiss, line 702: ()

src/packet/signature.js
```diff
@@ -108,6 +108,9 @@ class SignaturePacket {
 
     // hashed subpackets
     i += this.readSubPackets(bytes.subarray(i, bytes.length), true);
+    if (!this.created) {
+      throw new Error('Invalid signature packet: missing signature creation time in hashed area.');
```
twiss, line 112: This error also feels overly specific to me, I think it could just say
```suggestion
      throw new Error('Missing signature creation time subpacket.');
```
And then in many cases `Could not find valid * signature:` will already be prepended to it.

And then if someone complains that they have a signature with a creation time in the unhashed area, we can point them to https://datatracker.ietf.org/doc/html/draft-ietf-openpgp-crypto-refresh#section-5.2.3.4.

src/packet/signature.js
```diff
@@ -331,30 +339,22 @@ class SignaturePacket {
...
     const critical = bytes[mypos] & 0x80;
     const type = bytes[mypos] & 0x7F;
 
     // GPG puts the Issuer and Signature subpackets in the unhashed area.
-    // Tampering with those invalidates the signature, so we can trust them.
-    // Ignore all other unhashed subpackets.
-    if (!trusted && ![
+    // Tampering with those invalidates the signature, so we still trust them and parse them.
+    // All other unhashed subpackets are not processed further.
+    const allowedUntrustedTypes = new Set([
```
twiss, line 352: `allowedUnhashedSubpackets`?
And, if you're going to the trouble of making a separate variable, maybe you could put it at the top-level of the file, in order not to have to create it again every time this function is called?

src/packet/signature.js
```diff
@@ -363,11 +363,11 @@ class SignaturePacket {
 
     // subpacket type
     switch (type) {
-      case 2:
+      case enums.signatureSubpacket.signatureCreationTime:
```
twiss, line 366: Not a comment, but thanks for this :)

test/general/key.js
```diff
@@ -3251,17 +3245,11 @@ module.exports = () => describe('Key', function() {
...
-
-    destCertification.verified = null;
-    sourceCertification.verified = null;
-    await updated.users[0].verify(updated.keyPacket).then(async () => expect(destCertification.verified).to.be.true);
-    await source.users[0].verify(source.keyPacket).then(async () => expect(sourceCertification.verified).to.be.true);
+    await expect(updated.verifyPrimaryKey()).to.be.fulfilled;
+    await expect(source.verifyPrimaryKey()).to.be.fulfilled;
+
+    await expect(updated.users[0].verify(updated.keyPacket)).to.be.fulfilled;
+    await expect(source.users[0].verify(source.keyPacket)).to.be.fulfilled;
```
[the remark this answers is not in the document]
larabr: In these cases I would keep them, since it's otherwise unclear whether we are verifying to test that verifying works, or whether we need verification to test other side-effects IMO

src/packet/signature.js
larabr: I called them `Untrusted` because we use the `trusted` variable, I would either rename that as well, or leave `Untrusted`, otherwise I find the mix a bit confusing

test/general/key.js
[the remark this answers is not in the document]
twiss: Verification should never have side-effects, so IMO it should always be obvious that it's part of the test

src/packet/signature.js
twiss: OK, fair enough. Yeah, I would rename that as well, because indeed "hashed" and "trusted" doesn't perfectly correspond, exactly because of this variable (so you could also call it `trustedUnhashedSubpackets`, for example).

test/general/key.js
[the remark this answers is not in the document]
larabr: One day I'll convert you

src/key/key.js
larabr: Promise.all is awaited

twiss: I mean the one on line 410, not the one on line 408

src/key/key.js
larabr: Fixed in #1319

twiss (approved):

danhab99:
Hey I just noticed that the typedef for `revokeKey` is not in the typedefs file. Could that be added?

twiss:
See https://github.com/openpgpjs/openpgpjs/pull/1324 :)

PR #1318, opened 2021-06-01
Require keys in `openpgp.sign` and make all top-level functions fully async

larabr:
Changes:
- `openpgp.sign` throws if no signing keys are given (before, the operation would succeed, returning a non-signed literal packet)
- all top-level function errors cause promise rejection (before, some errors were thrown outside of promises)

### DIFF
--- a/src/openpgp.js
+++ b/src/openpgp.js
@@ -61,19 +61,20 @@ export async function generateKey({ userIDs = [], passphrase = "", type = "ecc",
   }
   const options = { userIDs, passphrase, type, rsaBits, curve, keyExpirationTime, date, subkeys };
 
-  return generate(options, config).then(async key => {
+  try {
+    const key = await generate(options, config);
     const revocationCertificate = await key.getRevocationCertificate(date, config);
     key.revocationSignatures = [];
 
     return {
-
-      key: key,
+      key,
       privateKeyArmored: key.armor(config),
       publicKeyArmored: key.toPublic().armor(config),
       revocationCertificate: revocationCertificate
-
     };
-  }).catch(err => onError('Error generating keypair', err));
+  } catch (err) {
+    return onError('Error generating keypair', err);
+  }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/openpgp.js
```diff
@@ -73,7 +73,7 @@ export function generateKey({ userIDs = [], passphrase = "", type = "ecc", rsaBi
       revocationCertificate: revocationCertificate
 
     };
-  }).catch(onError.bind(null, 'Error generating keypair'));
+  }).catch(err => onError('Error generating keypair', err));
```
twiss, line 76: I'd try to be consistent within this file, and personally I prefer the variant with the `try/catch`, so I'd use that in all these functions (and try to get rid of `.then` and `.catch`)

src/openpgp.js
```diff
@@ -253,14 +254,14 @@ export async function encryptKey({ privateKey, passphrase, config }) {
...
-export function encrypt({ message, encryptionKeys, signingKeys, passwords, sessionKey, armor = true, detached = false, signature = null, wildcard = false, signingKeyIDs = [], encryptionKeyIDs = [], date = new Date(), signingUserIDs = [], encryptionUserIDs = [], config }) {
+export async function encrypt({ message, encryptionKeys, signingKeys, passwords, sessionKey, armor = true, detached = false, signature = null, wildcard = false, signingKeyIDs = [], encryptionKeyIDs = [], date = new Date(), signingUserIDs = [], encryptionUserIDs = [], config }) {
   config = { ...defaultConfig, ...config };
   checkMessage(message); encryptionKeys = toArray(encryptionKeys); signingKeys = toArray(signingKeys); passwords = toArray(passwords); signingUserIDs = toArray(signingUserIDs); encryptionUserIDs = toArray(encryptionUserIDs);
   if (detached) {
     throw new Error("detached option has been removed from openpgp.encrypt. Separately call openpgp.sign instead. Don't forget to remove privateKeys option as well.");
   }
 
-  return Promise.resolve().then(async function() {
+  try {
```
twiss, line 264: In general, I'd try to make the `try/catch` as small as possible, e.g. in this case the `try` can be moved down 4 lines.

src/openpgp.js
```diff
@@ -61,19 +61,20 @@ export function generateKey({ userIDs = [], passphrase = "", type = "ecc", rsaBi
...
-      key: key,
+      key,
       privateKeyArmored: key.armor(config),
       publicKeyArmored: key.toPublic().armor(config),
       revocationCertificate: revocationCertificate
-
     };
-  }).catch(onError.bind(null, 'Error generating keypair'));
+  } catch (err) {
+    return onError('Error generating keypair', err);
```
twiss, line 76: It feels weird to me to `return` in a `catch`, maybe we could replace this with
```suggestion
    throw util.wrapError('Error generating keypair', err);
```
\- then it's kind of more obvious what's going on, and we could get rid of `onError` entirely - I don't think the part about `util.printDebugError` is really necessary.
(And similarly everywhere else)

larabr: Yeah I definitely found the `return onError` quite confusing as well 👍

twiss (approved):

PR #1324, opened 2021-06-05
Add typescript binding for revokeKey

gmreburn:
The documentation mentions a global revokeKey function but there is no typescript binding for it yet. https://openpgpjs.org/openpgpjs/global.html#revokeKey

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -22,7 +22,7 @@ export function generateSessionKey(options: { encryptionKeys: PublicKey[], date?
 export function decryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function encryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; config?: PartialConfig }): Promise<KeyPair>;
-export function revokeKey(options: {key: Key, revocationCertificate?: String, reasonForRevocation?: {flag?: enums.reasonForRevocation, string?: String}, config?: PartialConfig}): {Promise.<KeyPair|PublicKey>}
+export function revokeKey(options: {key: Key, revocationCertificate?: String, reasonForRevocation?: {flag?: enums.reasonForRevocation, string?: String}, config?: PartialConfig}): Promise<KeyPair|PublicKey>
 
 export abstract class Key {
   private primaryKey: PublicKeyPacket | SecretKeyPacket;
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -22,7 +22,8 @@ export function generateSessionKey(options: { encryptionKeys: PublicKey[], date?
 export function decryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function encryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; config?: PartialConfig }): Promise<KeyPair>;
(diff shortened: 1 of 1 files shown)


openpgp.d.ts
```diff
@@ -22,6 +22,7 @@ export function generateSessionKey(options: { encryptionKeys: PublicKey[], date?
 export function decryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function encryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; config?: PartialConfig }): Promise<KeyPair>;
+export function revokeKey(options: {key: Key, revocationCertificate?: String, reasonForRevocation?: {flag?: enums.reasonForRevocation, string?: String}, config?: PartialConfig}): Promise<KeyPair|PublicKey>
```
twiss, line 25: Hey :wave: Thanks for the PR! :+1:

A few nitpicks:
- Please try to follow the formatting of the surrounding code (spaces in object literals, simicolons);
- Since the `revocationCertificate` and `reasonForRevocation` parameters are essentially exclusive, could you create two overloads of this function, each with only one of them? That should make it clearer how this function can be used.
- And then finally, if you don't pass a revocation certificate, you have to pass a `PrivateKey`, whereas if you do, you can also pass a `PublicKey` (and the return type depends on whether you do - so you might in fact want three overloads - though perhaps we should just simplify the function at some point).

gmreburn: Thank you for the feedback. I think this can be consolidated to 2 overloads unless I am forgetting something. One overload for when `PrivateKey` is passed and another when `PublicKey` is passed. Let me know what you think of this.

openpgp.d.ts
```diff
@@ -22,6 +22,8 @@ export function generateSessionKey(options: { encryptionKeys: PublicKey[], date?
 export function decryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function encryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; config?: PartialConfig }): Promise<KeyPair>;
+export function revokeKey(options: { key: PrivateKey, reasonForRevocation?: { flag?: enums.reasonForRevocation, string?: String }, config?: PartialConfig }): Promise<PrivateKey>;
+export function revokeKey(options: { key: PublicKey, revocationCertificate?: String, config?: PartialConfig }): Promise<PublicKey>;
```
twiss, line 26: Thanks, and sorry for the delay! However, I think this return type currently doesn't match what the function actually returns - it either returns a `KeyPair`, or a partial one. That being said, I think it's kind of ugly and this type definition would make more sense, and would make the function simpler - so if you want, you could make a separate PR to just return the new key object, rather than a wrapper object with the armored keys etc?

openpgp.d.ts
```diff
@@ -22,6 +22,8 @@ export function generateSessionKey(options: { encryptionKeys: PublicKey[], date?
 export function decryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function encryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
 export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; config?: PartialConfig }): Promise<KeyPair>;
+export function revokeKey(options: { key: PrivateKey, reasonForRevocation?: { flag?: enums.reasonForRevocation, string?: String }, config?: PartialConfig }): Promise<PrivateKey>;
```
larabr, line 25: @gmreburn we've just introduced a `date` parameter for `revokeKey` in #1302 , could you add it to the definitions as well please? See https://github.com/openpgpjs/openpgpjs/pull/1302/files#diff-f0804beb9a26d785ef733ba8a71c41be04aa299329b68eb114798f3189fe5da4R133

openpgp.d.ts
gmreburn: I agree and opened #1349 based on this comment. Let me know if you were thinking something different.

twiss:
#1345 also included and adapted the typings you added for `revokeKey`. In hindsight, it would've been better to merge this PR first and then modify the function after. Apologies for that, and thanks for adding the typings nevertheless!

Issue #1325, opened 2021-06-06
How to get the public key ID/Fingerprint from a detatched signature


danhab99:
```ts
import * as openpgp from "openpgp";

let detatchedSignature = "-----BEGIN PGP SIGNATURE----- insert signature here -----END PGP SIGNATURE-----";
openpgp.readSignature({
  armoredSignature: detatchedSignature,
}).then((sig: openpgp.Signature) => {
  let signingPublicKey = // whaaa??
});

```

According the the [typedefs file](https://github.com/openpgpjs/openpgpjs/blob/master/openpgp.d.ts) the `verify` function needs the public key to test against, that makes sence. The problem is that I don't know what the public key might be. I have a database of public keys that I can query by `keyID` or `fingerprint`, and according to [RFC 4880](https://datatracker.ietf.org/doc/html/rfc4880) it is possible to get the fingerprint. The [typedefs for `openpgp.readSignature()`](https://github.com/openpgpjs/openpgpjs/blob/master/openpgp.d.ts#L109) I can input a detached signature and get the promise of a [`signature`](https://github.com/openpgpjs/openpgpjs/blob/master/openpgp.d.ts#L112), but the object doesn't provide the fingerprint.

So how can I get the fingerprint or ID of the public key that created a signature using OpenPGP.js? Do I need to ask the user for the public key too or can I just get it from the signature?

twiss:
Hey :wave: It's indeed technically possible to get the key IDs or fingerprints from the signature, even though there's no public method for it at the moment. If you want to go about creating one, you could take https://github.com/openpgpjs/openpgpjs/blob/f0280262172ddc54910c056e2c8d8d42be5643c2/src/message.js#L97-L100 as inspiration, and make a similar method in the Signature class. (A PR would be welcome ^.^)

larabr:
Fixed in #1331.

PR #1326, opened 2021-06-08
Added method for getting public key ids from signature

danhab99:
Resolving issue #1325 by adding a method for getting the KeyIDs of all the issuers of a given signature


src/signature.js
```diff
@@ -51,6 +51,15 @@ export class Signature {
   armor(config = defaultConfig) {
     return armor(enums.armor.signature, this.write(), undefined, undefined, undefined, config);
   }
+
+  /**
+   * 
+   * @returns Returns an array of KeyIDs of all of the issuers who created this signature
+   * @returns {Array<KeyID>} The IDs of the issuers
+   */
+  getIssuerIDs() {
```
twiss, line 60: It might be good to be consistent with `Message#getSigningKeyIDs`, and call this function the same thing.

src/signature.js
```diff
@@ -51,6 +51,15 @@ export class Signature {
...
+    return this.packets.map(packet => packet.issuerKeyID)
```
twiss, line 61: Technically speaking, there could be a marker packet here, as well, which wouldn't have an issuer key ID. So it might be good to still do something like `Message#getSigningKeyIDs` does, i.e. get the signature packets here explicitly.

package-lock.json
```diff
@@ -1,172 +1,6375 @@
 {
   "name": "openpgp",
   "version": "5.0.0-3",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
twiss, line 4: Since package.json is not changed in this PR, please don't change package-lock.json, either

README.md
```diff
@@ -525,6 +526,34 @@ Using the private key:
 })();
 ```
 
+#### Getting the IDs of the Issuers of a signature
```
twiss, line 529: To be honest, while I appreciate you writing some documentation for this, I'm not sure if this operation is common enough to warrant a full example in the top-level README. Once this PR is merged and included in a release, the method will automatically be included in https://openpgpjs.org/openpgpjs/Signature.html (generated from the JSDoc), I feel like that should be enough.

src/signature.js
```diff
@@ -51,6 +51,15 @@ export class Signature {
   armor(config = defaultConfig) {
     return armor(enums.armor.signature, this.write(), undefined, undefined, undefined, config);
   }
+
+  /**
+   * 
+   * @returns Returns an array of KeyIDs of all of the issuers who created this signature
+   * @returns {Array<KeyID>} The IDs of the issuers
```
twiss, line 58: Speaking of the JSDoc, this should be something like:
```suggestion
   * Returns an array of KeyIDs of all of the issuers who created this signature
   * @returns {Array<KeyID>} The Key IDs of the signing keys
```

twiss (commented):
Hey @danhab99, thanks for the PR! See some nitpicks below.

README.md
```diff
@@ -5,24 +5,25 @@ OpenPGP.js [![BrowserStack Status](https://automate.browserstack.com/badge.svg?b
 
 **Table of Contents**
 
-- [OpenPGP.js](#openpgpjs)
+- [OpenPGP.js ![BrowserStack Status](https://automate.browserstack.com/public-build/N1l2eHFOanVBMU9wYWxJM3ZnWERnc1lidkt5UkRqa3BralV3SWVhOGpGTT0tLVljSjE4Z3dzVmdiQjl6RWgxb2c3T2c9PQ==--5864052cd523f751b6b907d547ac9c4c5f88c8a3) [![Join the chat on Gitter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/openpgpjs/openpgpjs?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)](#openpgpjs--)
```
twiss, line 8: Thanks! Could you revert these changes to the table of contents as well?

src/signature.js
```diff
@@ -51,6 +51,14 @@ export class Signature {
   armor(config = defaultConfig) {
     return armor(enums.armor.signature, this.write(), undefined, undefined, undefined, config);
   }
+
+  /**
+   * Returns an array of KeyIDs of all of the issuers who created this signature
+   * @returns {Array<KeyID>} The Key IDs of the signing keys
+   */
+  getSigningKeyIDs() {
+    return this.packets.filter(packet => packet.tag !== enums.packet.marker).map(packet => packet.issuerKeyID)
```
twiss, line 60: And then finally, I would prefer to select the signature packets explicitly, rather than the other way around. See `Message#getSigningKeyIDs` for inspiration. (Also, there's a final semicolon missing.)

danhab99: I don't quite understand how `Message#getSigningKeyIDs` relates to my solution.

danhab99: figured it out

PR #1333, opened 2021-06-14
CI: detect unhandled rejections in browser tests

larabr:
Follow up to https://github.com/openpgpjs/openpgpjs/pull/1251 , which added an unhandled rejection listener for Node tests. This PR does the same for browsers.

Also target the Safari release for macOS Big Sur in Browserstack.

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -102,7 +102,7 @@ module.exports = function(config) {
       bs_safari_14: {
         base: 'BrowserStack',
         browser: 'Safari',
-        browser_version: '14',
+        browser_version: '14.0',
         os: 'OS X',
         os_version: 'Big Sur'
       },
--- a/test/unittests.js
+++ b/test/unittests.js
@@ -31,7 +31,7 @@ describe('Unit Tests', function () {
   openpgp.config.s2kIterationCountByte = 0;
 
   if (typeof window !== 'undefined') {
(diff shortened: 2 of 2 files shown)


test/unittests.js
```diff
@@ -31,6 +31,10 @@ describe('Unit Tests', function () {
   openpgp.config.s2kIterationCountByte = 0;
 
   if (typeof window !== 'undefined') {
+    window.addEventListener("unhandledrejection", function (event) {
```
twiss, line 34: Very nitpicky, but:
```suggestion
    window.addEventListener('unhandledrejection', function (event) {
```

test/karma.conf.js
```diff
@@ -102,7 +102,7 @@ module.exports = function(config) {
       bs_safari_14: {
```
twiss, line 102: Sorry, one more nitpick, maybe we should make this
```suggestion
      bs_safari_14_0: {
```
for consistency with 13_1, and then later add 14_1 as well?

larabr: I would just target the latest stable one, since we do the same in Chrome and Firefox. I don't think it'll be fixed to 14.0 for BigSur

twiss: Oki :+1:

twiss (approved):

PR #1336, opened 2021-06-15
Always generate RSA keys of expected bit length

larabr:
- Fix RSA key generation code used when no native crypto library is available (i.e. no NodeCrypto or WebCrypto). Now generated keys are always of exact bit length. This was not guaranteed before, and it was common for keys to be one bit shorter than expected.
- Remove leftover code related to legacy WebCrypto interfaces (IE11 and Safari 10)

### DIFF
--- a/src/crypto/public_key/prime.js
+++ b/src/crypto/public_key/prime.js
@@ -25,7 +25,7 @@ import util from '../../util';
 import { getRandomBigInteger } from '../random';
 
 /**
- * Probabilistic random number generator
+ * Probabilistic prime random number generator
  * @param {Integer} bits - Bit length of the prime
  * @param {BigInteger} e - Optional RSA exponent to check against the prime
  * @param {Integer} k - Optional number of iterations of Miller-Rabin test
--- a/src/crypto/public_key/prime.js
+++ b/src/crypto/public_key/prime.js
@@ -25,7 +25,7 @@ import util from '../../util';
 import { getRandomBigInteger } from '../random';
 
 /**
- * Probabilistic prime random number generator
+ * Generate a probably prime random number
  * @param {Integer} bits - Bit length of the prime
  * @param {BigInteger} e - Optional RSA exponent to check against the prime
  * @param {Integer} k - Optional number of iterations of Miller-Rabin test
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -175,11 +175,7 @@ export async function generate(bits, e) {
 
     // export the generated keys as JsonWebKey (JWK)
     // https://tools.ietf.org/html/draft-ietf-jose-json-web-key-33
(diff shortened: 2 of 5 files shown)


src/crypto/public_key/prime.js
```diff
@@ -25,7 +25,7 @@ import util from '../../util';
 import { getRandomBigInteger } from '../random';
 
 /**
- * Probabilistic random number generator
+ * Probabilistic prime random number generator
```
twiss, line 28:
```suggestion
 * Probabilistic random prime number generator
```

src/crypto/public_key/rsa.js
```diff
@@ -178,41 +163,20 @@ export async function generate(bits, e) {
...
     // https://tools.ietf.org/html/draft-ietf-jose-json-web-key-33
-    let jwk = webCrypto.exportKey('jwk', keyPair.privateKey);
-    jwk = await promisifyIE11Op(jwk, 'Error exporting RSA key pair.');
-
-    // parse raw ArrayBuffer bytes to jwk/json (WebKit/Safari/IE11 quirk)
+    let jwk = await webCrypto.exportKey('jwk', keyPair.privateKey);
+    // parse raw ArrayBuffer bytes to jwk/json (WebKit/Safari quirk)
     if (jwk instanceof ArrayBuffer) {
       jwk = JSON.parse(String.fromCharCode.apply(null, new Uint8Array(jwk)));
     }
```
twiss, line 182: Do we still need this?

src/crypto/public_key/rsa.js
```diff
@@ -261,15 +225,23 @@ export async function generate(bits, e) {
...
+    n = p.mul(q);
+  } while (n.bitLength() !== bits);
+
+  const phi = p.dec().imul(q.dec());
 
   if (q.lt(p)) {
     [p, q] = [q, p];
   }
-  const phi = p.dec().imul(q.dec());
+
```
twiss, line 242: I get that it doesn't matter, but I think it was clearer the other way around - then it's sort of more obvious that flipping p and q doesn't cause any issues.

larabr: I had the opposite impression, namely that doing it afterwards implied the order mattered 😄  and in the WebCrypto code above we also swap the numbers when returning, so it's consistent with that.

twiss: OK, fair enough :+1:

src/crypto/public_key/rsa.js
larabr: I guess not, I cannot find any docs related to it, but Safari 11 works fine without it

twiss: :+1: I guess it was only necessary for `webkitSubtle`, then

twiss (approved):

PR #1342, opened 2021-06-17
Support passing a non-array value to `encryption/signingKeyIDs` in top-level functions

larabr:
- support passing a single Key ID directly to `encryption/signingKeyIDs` in `encrypt`, `sign`, `generateSessionKey` and `encryptSessionKey`, to be consistent with how we treat the other options
- add type definitions for `encryptSessionKey` and `decryptSessionKeys`

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -22,11 +22,11 @@ export function generateKey(options: KeyOptions & { format: 'binary' }): Promise
 export function generateKey(options: KeyOptions & { format: 'object' }): Promise<KeyPair & { revocationCertificate: string }>;
 
 export function generateSessionKey(options: { encryptionKeys: PublicKey[], date?: Date, encryptionUserIDs?: UserID[], config?: PartialConfig }): Promise<SessionKey>;
-export function decryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
-export function encryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
-export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; date?: Date, format?: 'armor', config?: PartialConfig }): Promise<SerializedKeyPair<string> & { revocationCertificate: string }>;
-export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; date?: Date, format: 'binary', config?: PartialConfig }): Promise<SerializedKeyPair<Uint8Array> & { revocationCertificate: string }>;
-export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; date?: Date, format: 'object', config?: PartialConfig }): Promise<KeyPair & { revocationCertificate: string }>;
+export function decryptKey(options: { privateKey: PrivateKey; passphrase?: MaybeArray<string>; config?: PartialConfig }): Promise<PrivateKey>;
+export function encryptKey(options: { privateKey: PrivateKey; passphrase?: MaybeArray<string>; config?: PartialConfig }): Promise<PrivateKey>;
+export function reformatKey(options: { privateKey: PrivateKey; userIDs?: MaybeArray<UserID>; passphrase?: string; keyExpirationTime?: number; date?: Date, format?: 'armor', config?: PartialConfig }): Promise<SerializedKeyPair<string> & { revocationCertificate: string }>;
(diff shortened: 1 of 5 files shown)


openpgp.d.ts
```diff
@@ -17,11 +17,9 @@ export function readPrivateKey(options: { armoredKey: string, config?: PartialCo
...
 export function readPrivateKeys(options: { armoredKeys: string, config?: PartialConfig }): Promise<PrivateKey[]>;
 export function readPrivateKeys(options: { binaryKeys: Uint8Array, config?: PartialConfig }): Promise<PrivateKey[]>;
-export function generateKey(options: KeyOptions): Promise<KeyPair>;
-export function generateSessionKey(options: { encryptionKeys: PublicKey[], date?: Date, encryptionUserIDs?: UserID[], config?: PartialConfig }): Promise<SessionKey>;
-export function decryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
-export function encryptKey(options: { privateKey: PrivateKey; passphrase?: string | string[]; config?: PartialConfig }): Promise<PrivateKey>;
-export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; config?: PartialConfig }): Promise<KeyPair>;
+export function generateKey(options: KeyOptions): Promise<KeyPair>;export function decryptKey(options: { privateKey: PrivateKey; passphrase?: MaybeArray<string>; config?: PartialConfig }): Promise<PrivateKey>;
```
twiss, line 20:
```suggestion
export function generateKey(options: KeyOptions): Promise<KeyPair>;
export function decryptKey(options: { privateKey: PrivateKey; passphrase?: MaybeArray<string>; config?: PartialConfig }): Promise<PrivateKey>;
```

src/openpgp.js
```diff
@@ -287,7 +288,7 @@ export async function encrypt({ message, encryptionKeys, signingKeys, passwords,
 }
 
 /**
- * Decrypts a message with the user's private key, a session key or a password. Either a private key,
+ * Decrypts a message with the user's private key, a session key or a password (multiple options are not supported). Either a private key,
  *   a session key or a password must be specified.
```
twiss, line 292:
```suggestion
 * Decrypts a message with the user's private key, a session key or a password. Either private keys,
 *   session keys or passwords must be specified (passing a combination of these is not supported).
```

src/openpgp.js
```diff
@@ -508,7 +509,7 @@ export async function generateSessionKey({ encryptionKeys, date = new Date(), en
 }
 
 /**
- * Encrypt a symmetric session key with public keys, passwords, or both at once. At least either public keys
```
twiss, line 511: In this case, the old text was correct; `Message.encryptSessionKey` actually supports passing both public keys and passwords.

src/openpgp.js
```diff
@@ -540,15 +541,15 @@ export async function encryptSessionKey({ data, algorithm, aeadAlgorithm, encryp
 }
 
 /**
- * Decrypt symmetric session keys with a private key or password. Either a private key or
+ * Decrypt symmetric session keys with private keys or passwords (not both). Either a private key or
  *   a password must be specified.
```
twiss, line 545:
```suggestion
 * Decrypt symmetric session keys with a private key or password (not both). Either private keys or
 *   passwords must be specified (passing both is not supported).
```
The distinction is somewhat subtle, but it makes slightly more sense to me this way because while you can pass multiple private keys or passwords, it ends up (succesfully) decrypting the message using (at most) only one of them.

src/openpgp.js
```diff
@@ -235,7 +235,7 @@ export async function encryptKey({ privateKey, passphrase, config }) {
 
 
 /**
- * Encrypts message text/data with public keys, passwords or both at once. At least either encryption keys or passwords
```
twiss, line 238: This was also correct as written

src/openpgp.js
larabr: I am not sure the existing text is correct English..? I'd write "at least one of" (I left only "Either..." in an attempt to mean the same, but indeed it can still be ambiguous whether you can pass both 😅 )

src/openpgp.js
larabr: Uhm, I'd leave the singular everywhere then, because we still require only 1 key or password (and we have the singular notation in the `encrypt*` functions)

twiss: I think "Either private keys or passwords must be specified" could be taken to refer to the option names `privateKeys` and `passwords` - though actually now that it's `decryptionKeys` it should probably say that instead. I don't think it necessarily implies that you must pass more than one, since neither does the option name `decryptionKeys`. But yeah, we should probably be a bit consistent across all functions.

src/openpgp.js
twiss: "At least one of" sounds good :+1:

twiss (approved):

PR #1345, opened 2021-06-21
Simplify return value of `generateKey`, `reformatKey` and `revokeKey` and add support for binary output

larabr:
This PR changes the return value of `openpgp.generateKey`, `reformatKey` and `revokeKey` and adds support for returning a key pair in binary form.

Changes:
- A `format` options has been added to all three functions, accepting the values `'armor', 'binary', 'object'` (defaulting to `armor`).
- `generateKey` and `reformatKey` now return an object of the form `{ publicKey, privateKey, revocationCertificate }`, where the type of `publicKey` and `privateKey` depends on `options.format`:
   - if `format: 'armor'` then `privateKey, publicKey` are armored strings
   -  if `format: 'binary'` then `privateKey, publicKey` are `Uint8Array`
   -  if `format: 'object'` then `privateKey, publicKey` are `PrivateKey` and `PublicKey` objects respectively
- `revokeKey` now returns `{ publicKey, privateKey }`, where:
   - if a `PrivateKey` was passed as input,  `privateKey, publicKey` are of the requested format
   - if a `PublicKey` was passed as input, `publicKey` is of the requested format, while `privateKey` is `null` (previously, in this case the `privateKey` field was not defined).

Breaking changes in `generateKey`, `reformatKey` and `revokeKey`:
- In `revokeKey`, if no `format` option is specified in input, the returned `publicKey, privateKey` are armored strings (they used to be objects).
- In `generateKey` and `reformatKey`, the `key` value is no longer returned.
- For all functions, the `publicKeyArmored` and `privateKeyArmored` values are no longer returned.
To get armored keys, either call e.g. `generateKey({ ..., format: 'armor' })` or use the `armor()` method on the object keys: 
```js
generateKey({ ..., format: 'object' }).then({ privateKey, publicKey }) => {
   console.log(privateKey.armor())
}
```

Also, superseeds #1324

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -27,6 +27,12 @@ export function encryptKey(options: { privateKey: PrivateKey; passphrase?: strin
 export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; date?: Date, format?: 'armor', config?: PartialConfig }): Promise<SerializedKeyPair<string>>;
 export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; date?: Date, format: 'binary', config?: PartialConfig }): Promise<SerializedKeyPair<Uint8Array>>;
 export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; date?: Date, format: 'object', config?: PartialConfig }): Promise<KeyPair>;
+export function revokeKey(options: { key: PrivateKey, reasonForRevocation?: ReasonForRevocation, date?: Date, format?: 'armor', config?: PartialConfig }): Promise<string>;
+export function revokeKey(options: { key: PrivateKey, reasonForRevocation?: ReasonForRevocation, date?: Date, format: 'binary', config?: PartialConfig }): Promise<Uint8Array>;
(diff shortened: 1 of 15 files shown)


openpgp.d.ts
```diff
@@ -17,11 +17,25 @@ export function readPrivateKey(options: { armoredKey: string, config?: PartialCo
...
+export function reformatKey(options: { privateKey: PrivateKey; userIDs?: UserID|UserID[]; passphrase?: string; keyExpirationTime?: number; date?: Date, format: 'object', config?: PartialConfig }): Promise<KeyPair & { revocationCertificate: string }>;
+export function revokeKey(options: { key: PrivateKey, reasonForRevocation?: ReasonForRevocation, date?: Date, format?: 'armor', config?: PartialConfig }): Promise<SerializedKeyPair<string>>;
+export function revokeKey(options: { key: PrivateKey, reasonForRevocation?: ReasonForRevocation, date?: Date, format: 'binary', config?: PartialConfig }): Promise<SerializedKeyPair<Uint8Array>>;
+export function revokeKey(options: { key: PrivateKey, reasonForRevocation?: ReasonForRevocation, date?: Date, format: 'object', config?: PartialConfig }): Promise<KeyPair>;
+export function revokeKey(options: { key: PrivateKey, revocationCertificate: string, date?: Date, format?: 'armor', config?: PartialConfig }): Promise<SerializedKeyPair<string>>;
```
larabr, line 33: I would merge the private key definitions and leave both `reasonForRevocation` and `revocationCertificate`, even though they are mutually exclusive, the function description also clarifies that only one is used...

twiss: I would keep them at this point; the type definitions are here to help users understand how the functions should be used; more descriptive is better. Simplifying these doesn't make the actual functions simpler :sweat_smile:

twiss (approved):

PR #1348, opened 2021-06-23
Remove `valid` and `error` from the verification result of `openpgp.verify` and `decrypt`

### LINKED ISSUE #1313: `valid` property missing on type VerificationResult
Docs [here](https://github.com/openpgpjs/openpgpjs/blame/master/README.md#L236) say that you can ask for `.valid` to type `VerificationResult` which is defined [here](https://github.com/openpgpjs/openpgpjs/blob/f0280262172ddc54910c056e2c8d8d42be5643c2/openpgp.d.ts#L119) but it doesn't define `valid` property.

```ts
interface VerificationResult {
  keyID: KeyID;
  verified: Promise<null | boolean>;
  signature: Promise<Signature>;
}
```

image

larabr:
This PR removes the `valid` and `error` properties that were returned by `openpgp.verify` and `decrypt` when verifying a non-streamed message. This change is to make the code more consistent between the streaming and non-streaming cases.

The validity of a signature can be determined by using the (existing) `verified` promise:
```js
const verificationResult = await openpgp.verify({
  message,
  verificationKeys: publicKey
});

const { verified } = verificationResult.signatures[0];
try {
  await verified; // throws on invalid signature
  console.log('signature is valid');
} catch (e) {
  throw new Error('Signature could not be verified: ' + e.message);
}

// The following is no longer possible:
const { valid, error } = verificationResult.signatures[0];
if (valid) {
  console.log('signature is valid')
} else {
  console.log(error)
}
```

Close #1313 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,7 @@ and a subkey for encryption using Curve25519.
         userIDs: [{ name: 'Jon Smith', email: 'jon@example.com' }], // you can pass multiple user IDs
         passphrase: 'super long and hard to guess secret', // protects the private key
         format: 'armor' // output key format, defaults to 'armor' (other options: 'binary' or 'object')
+    });
 
     console.log(privateKey);     // '-----BEGIN PGP PRIVATE KEY BLOCK ... '
     console.log(publicKey);      // '-----BEGIN PGP PUBLIC KEY BLOCK ... '
--- a/README.md
+++ b/README.md
@@ -233,7 +233,13 @@ const openpgp = require('openpgp'); // use as CommonJS, AMD, ES6 module or via w
         decryptionKeys: privateKey
     });
     console.log(decrypted); // 'Hello, World!'
-    console.log(signatures[0].valid) // signature validity (signed messages only)
+    // check signature validity (signed messages only)
+    try {
+        await signatures[0].verified; // throws on invalid signature
+        console.log('signature is valid');
+    } catch (e) {
+        throw new Error('Signature could not be verified: ' + e.message);
+    }
 })();
 ```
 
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -233,7 +233,13 @@ const openpgp = require('openpgp'); // use as CommonJS, AMD, ES6 module or via w
         decryptionKeys: privateKey
     });
     console.log(decrypted); // 'Hello, World!'
-    console.log(signatures[0].valid) // signature validity (signed messages only)
+    // check signature validity (signed messages only)
+    try {
+        await signatures[0].verified; // throws on invalid signature
+        console.log('signature is valid');
```
twiss, line 239: Very nitpicky but, let's try to make the capitalization vaguely consistent (also below)
```suggestion
        console.log('Signature is valid');
```

README.md
```diff
@@ -515,15 +522,16 @@ Using the private key:
...
     });
-    const { valid } = verified.signatures[0];
-    if (valid) {
-        console.log('signed by key id ' + verified.signatures[0].keyID.toHex());
-    } else {
-        throw new Error('signature could not be verified');
+    const { verified, keyID } = verificationResult.signatures[0];
+    try {
+        await verified; // throws on invalid signature
+        console.log('signed by key id ' + keyID.toHex());
```
twiss, line 532: The [also below](https://github.com/openpgpjs/openpgpjs/pull/1348#discussion_r658649826) was significant :upside_down_face:
```suggestion
        console.log('Signed by key id ' + keyID.toHex());
```

README.md
```diff
@@ -558,16 +566,17 @@ Using the private key:
...
```
twiss, line 577: Also here
```suggestion
        console.log('Signed by key id ' + keyID.toHex());
```

README.md
```diff
@@ -603,21 +612,21 @@ Using the private key:
...
     // verification of the data.
 
-    const valid = await verified.signatures[0].verified;
-    if (valid) {
-        console.log('signed by key id ' + verified.signatures[0].keyID.toHex());
-    } else {
-        throw new Error('signature could not be verified');
+    try {
+        await verificationResult.signatures[0].verified; // throws on invalid signature
+        console.log('signed by key id ' + verificationResult.signatures[0].keyID.toHex());
```
twiss, line 627:
```suggestion
        console.log('Signed by key id ' + verificationResult.signatures[0].keyID.toHex());
```

twiss (approved):

PR #1349, opened 2021-06-23
Simplify revoke method return

gmreburn:
The wrapper object is not necessary and performs extra work that may be unnecessary for a client. Passing the revoked key allows the consumer to decide to armor or create public key. See https://github.com/openpgpjs/openpgpjs/pull/1324/files#r650742027 for more detail/inspiration.

### DIFF
--- a/src/encoding/armor.js
+++ b/src/encoding/armor.js
@@ -360,6 +360,7 @@ export function unarmor(input, config = defaultConfig) {
  * @static
  */
 export function armor(messageType, body, partIndex, partTotal, customComment, config = defaultConfig) {
+  config = { ...defaultConfig, ...config };
   let text;
   let hash;
   if (messageType === enums.armor.signed) {
--- a/test/general/config.js
+++ b/test/general/config.js
@@ -136,11 +136,11 @@ vAFM3jjrAQDgJPXsv8PqCrLGDuMa/2r6SgzYd03aw/xt1WM6hgUvhQD+J54Z
       const { key, revocationCertificate } = await openpgp.generateKey({ userIDs });
 
       const opt = { key };
-      const { privateKeyArmored: revKeyArmored } = await openpgp.revokeKey(opt);
+      const revKeyArmored = await openpgp.revokeKey(opt).then(revKey => revKey.armor(opt.config));
       expect(revKeyArmored.indexOf(openpgp.config.commentString) > 0).to.be.false;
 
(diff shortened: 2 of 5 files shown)


src/encoding/armor.js
```diff
@@ -360,6 +360,7 @@ export function unarmor(input, config = defaultConfig) {
  * @static
  */
 export function armor(messageType, body, partIndex, partTotal, customComment, config = defaultConfig) {
+  config = { ...defaultConfig, ...config };
```
gmreburn, line 363: This change was made for [Custom configuration openpgp.revokeKey](https://github.com/openpgpjs/openpgpjs/blob/master/test/general/config.js#L144) test to pass. `defaultConfig` is merged in the `revokeKey` method but `armor` is no longer called from `revoke` method's context. Since `defaultConfig` is not merged in [PrivateKey.armor](https://github.com/openpgpjs/openpgpjs/blob/master/src/key/private_key.js#L76), I thought it might be good to place this merge logic in the armor encoder to avoid something like this (a missed default config merge) in the future.

I mention this because it might be better to place this line in the `PrivateKey.armor` method instead to avoid some unnecessary merging.

test/general/config.js
```diff
@@ -136,11 +136,11 @@ vAFM3jjrAQDgJPXsv8PqCrLGDuMa/2r6SgzYd03aw/xt1WM6hgUvhQD+J54Z
       const { key, revocationCertificate } = await openpgp.generateKey({ userIDs });
 
       const opt = { key };
-      const { privateKeyArmored: revKeyArmored } = await openpgp.revokeKey(opt);
+      const revKeyArmored = await openpgp.revokeKey(opt).then(revKey => revKey.armor(opt.config));
```
gmreburn, line 139: do you prefer this broken into multiple variable/awaits?

```suggestion
      const revKey = await openpgp.revokeKey(opt);
      const revKeyArmored = await revKey.armor(opt.config);
```

gmreburn:
This is an enabler for #1324

twiss:
Hey :wave: After discussing more with @larabr, we came to the conclusion that we should still armor the key by default in `revokeKey`, but add a `format` parameter to change that behavior. That also solves the issues you ran into with `key.armor()`. She implemented that in #1345, which kind of invalidates this PR, so I'll close this one. Sorry, I should've let you know in advance, but didn't know you were still planning to work on this. This was a bit of unfortunate timing, apologies.

PR #1354, opened 2021-06-28
Replace `armor` option with `format` in `openpgp.encrypt`, `sign` and `encryptSessionKey`

larabr:
Breaking changes to top-level functions:
- The `format` option has been added to `openpgp.encrypt`, `sign` and `encryptSessionKey` to select the format of the output message. `format` replaces the existing `armor` option, and accepts three values:
  - if `format: 'armor'` (default), an armored signed/encrypted message is returned (same as `armor: true`).
  - if `format: 'binary'`,  a binary signed/encrypted message is returned (same as `armor: false`).
  - if `format: 'object'`, a Message or Signature object is returned (this was not supported before).

  This change is to uniform the output format selection across all top-level functions (following up to #1345).
- All top-level functions now throw if unrecognised options are passed, to make library users aware that those options are not being applied.

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -177,6 +177,9 @@ export function encryptSessionKey(options: SessionKey & {
 export function encryptSessionKey(options: SessionKey & { 
   encryptionKeys?: MaybeArray<PublicKey>, passwords?: MaybeArray<string>, format: 'binary', wildcard?: boolean, encryptionKeyIDs?: MaybeArray<KeyID>, date?: Date, encryptionUserIDs?: MaybeArray<UserID>, config?: PartialConfig
 }) : Promise<Uint8Array>;
+export function encryptSessionKey(options: SessionKey & { 
+  encryptionKeys?: MaybeArray<PublicKey>, passwords?: MaybeArray<string>, format: 'object', wildcard?: boolean, encryptionKeyIDs?: MaybeArray<KeyID>, date?: Date, encryptionUserIDs?: MaybeArray<UserID>, config?: PartialConfig
+}) : Promise<Message<Data>>;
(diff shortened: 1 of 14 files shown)


src/openpgp.js
```diff
@@ -32,6 +32,7 @@ import util from './util';
 
 /**
  * Generates a new OpenPGP key pair. Supports RSA and ECC keys. By default, primary and subkeys will be of same type.
+ * The primary key will not have encryption capabilities.
```
twiss, line 35: If we're gonna write this, I think it would make sense to expand a bit and say what capabilities it does have, e.g.
```suggestion
 * By default, a primary key with signing capabilities and one subkey with encryption capabilities are generated.
```

openpgp.d.ts
```diff
@@ -577,7 +577,7 @@ interface EncryptOptions {
   /** (optional) session key */
   sessionKey?: SessionKey;
   /** if the return values should be ascii armored or the message/signature objects */
-  armor?: boolean;
+  format?: 'armor' | 'binary';
```
twiss, line 580: After staring at this PR for a while, I do think that `format: 'armored'` indeed makes a bit more sense :sweat_smile:
The comment here kind of confirms the point :upside_down_face: 
Since it's the default, the fact that it's a bit longer shouldn't really matter.

src/openpgp.js
larabr: I think this gives the impression that you can change the capabilities of the primary key?
```suggestion
 * The generated primary key will have signing capabilities. By default, one subkey with encryption capabilities is also generated.
```

twiss: Yeah, fair enough :+1: (Maybe we should offer the possibility to generate a primary key with only certification capabilities, and a signing subkey and an encryption subkey, but ok, that's very out of scope here :sweat_smile:)

openpgp.d.ts
larabr: I will open a new PR to rename this also in the openpgp.*Key functions 👍

twiss (approved):

PR #1362, opened 2021-07-03
Update README.md

akii0008:
Changed `openpgp.readKey()` to `openpgp.readPrivateKey()` where applicable. I noticed this documentation error when I was writing my own app in typescript and vscode kept yelling at me that it wanted a PrivateKey object instead of a PublicKey object, and I was confused because I wrote everything correct according to this README. I looked into the type definitions file and found that `readKey()` only returns a PublicKey object, and `readPrivateKey()` only returns a PrivateKey object, which is what was needed for `decryptKey()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ const openpgp = require('openpgp'); // use as CommonJS, AMD, ES6 module or via w
     const publicKey = await openpgp.readKey({ armoredKey: publicKeyArmored });
 
     const privateKey = await openpgp.decryptKey({
-        privateKey: await openpgp.readKey({ armoredKey: privateKeyArmored }),
+        privateKey: await openpgp.readPrivateKey({ armoredKey: privateKeyArmored }),
         passphrase
     });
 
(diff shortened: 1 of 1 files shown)


akii0008:
ah, there's another error while i'm at it (which i'll get to tomorrow morning- i'm writing this from bed lol) `privateKeys` below the lines i edited should be something like `signingKeys` if i recall correctly.

larabr:
The README is not really written with TS in mind at the moment, so we might want to add a separate section just for that.
For people not using TS, I think leaving `readKey` everywhere might be less confusing (`readKey` is not intended as "readPublicKey", since it can be used for both public and private keys, so I wouldn't mix it with `readPrivateKey`). What do you think @twiss?

FYI: you can also downcast the result of `readKey` to a PrivateKey. This will be more obvious with the new release, thanks to #1347 .

twiss:
Tbh I'm not opposed to this patch; this way it works in both TypeScript and JavaScript. While it's true that mixing `readKey` and `readPrivateKey` is a bit strange, that's a quirk of the API atm. If we want, we could add a `readPublicKey`, even though it's a bit useless :shrug:

twiss (approved):

twiss:
(@akii0008 Thanks for the PR!)

PR #1395, opened 2021-08-06
Add `config.rejectCurves` and prevent generating keys using blacklisted algos

larabr:
Breaking changes:
- throw error on key generation if the requested public key algorithm is included in `config.rejectPublicKeyAlgorithms`;
- add `config.rejectCurves` to blacklist a set of ECC curves, to prevent keys using those curves from being generated, or being used to encrypt/decrypt/sign/verify messages.
By default, `config.rejectCurves` includes the brainpool curves (`brainpoolP256r1`, `brainpoolP384r1`, `brainpoolP512r1`) and the Bitcoin curve (`secp256k1`). This is because [it's unclear whether these curves will be standardised](https://gitlab.com/openpgp-wg/rfc4880bis/-/merge_requests/47#note_634199141), and we prefer to blacklist them already, rather than introduce a breaking change after release.

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -179,8 +179,11 @@ export default {
    */
   knownNotations: ['preferred-email-encoding@pgp.com', 'pka-address@gnupg.org'],
   /**
+   * Whether to use the indutny/elliptic library for NIST and Brainpool curves that are not supported by the available native crypto library.
+   * When false, certain standard curves will not be supported (depending on the platform).
+   * Note: the indutny/elliptic curve library is not designed to be constant time.
    * @memberof module:config
-   * @property {Boolean} useIndutnyElliptic Whether to use the indutny/elliptic library. When false, certain curves will not be supported.
+   * @property {Boolean} useIndutnyElliptic
    */
   useIndutnyElliptic: true,
   /**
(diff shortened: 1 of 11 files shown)


src/openpgp.js
```diff
@@ -63,10 +64,12 @@ export async function generateKey({ userIDs = [], passphrase = '', type = 'ecc',
   if (type === 'rsa' && rsaBits < config.minRSABits) {
     throw new Error(`rsaBits should be at least ${config.minRSABits}, got: ${rsaBits}`);
   }
+
   const options = { userIDs, passphrase, type, rsaBits, curve, keyExpirationTime, date, subkeys };
 
   try {
     const { key, revocationCertificate } = await generate(options, config);
+    checkKeyRequirements(key.keyPacket, config);
```
larabr, line 72: This is the more elegant solution, since before generation we don't necessarily have the final algo + curve pairings. The downside of checking the key after generation, is that the thrown error will differ depending on the platform and config options. e.g. if someone disables `config.useIndutnyElliptic`, and leaves the brainpool curves blacklisted, the error will be `This curve is only supported in the full build of OpenPGP.js` (from `generate`), rather than `... support for curve XXX is disabled`.

src/config/config.js
```diff
@@ -175,8 +175,11 @@ export default {
    */
   knownNotations: ['preferred-email-encoding@pgp.com', 'pka-address@gnupg.org'],
   /**
+   * Whether to use the indutny/elliptic library for NIST and Brainpool curves that are not supported by the available native crypto library.
```
twiss, line 178: And secp256k1. Alternatively, maybe we can just say
```suggestion
   * Whether to use the indutny/elliptic library for curves (other than Curve25519) that are not supported by the available native crypto API.
```

src/config/config.js
```diff
@@ -192,9 +195,15 @@ export default {
...
    * @property {Set<Integer>} rejectPublicKeyAlgorithms {@link module:enums.publicKey}
    */
-  rejectPublicKeyAlgorithms: new Set([enums.publicKey.elgamal, enums.publicKey.dsa])
+  rejectPublicKeyAlgorithms: new Set([enums.publicKey.elgamal, enums.publicKey.dsa]),
+  /**
+   * Reject non-standard curves for key generation, message encryption, signing or verification
+   * @memberof module:config
+   * @property {Set<String>} rejectCurves {@link module:enums.curve}
+   */
+  rejectCurves: new Set([enums.curve.brainpoolP256r1, enums.curve.brainpoolP384r1, enums.curve.brainpoolP512r1])
```
twiss, line 208: We should reject secp256k1 as well by default

src/openpgp.js
twiss: I'm not opposed to this, but probably we should check all subkey packets as well here?

twiss (approved):

PR #1401, opened 2021-08-17
Validate email addresses without a subdomain

johnoliverdriscoll:
Trying to generate a key like so:

```javascript
const key = await pgp.generateKey({
  curve: 'secp256k1',
  userIds: [{ name: 'Mocha Tests', email: 'mocha@localhost' }],
})
```

results in an error: `Error: Error generating keypair: Invalid user id format`

While `mocha@localhost` is a valid email on my local network, it does not pass openpgp's validation. However, changing it to `mocha@localhost.zzz` does pass validation, even though it is not a valid email on my network.

This PR changes the regular expression in `isEmailAddress` from
```/^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+([a-zA-Z]{2,}|xn--[a-zA-Z\-0-9]+)))$/```
to
```/^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.?)+([a-zA-Z]{2,}|xn--[a-zA-Z\-0-9]+)))$/```
removing the mandate that the host part of the address contain a subdomain.


twiss:
Hello :wave: While the OpenPGP spec allows this, I'm not a super big fan of changing `isEmailAddress` in this way, especially because the regex now simply checks for a label + TLD concatenated together without a dot, which is a bit of a hack (and `a@b` would still fail, presumably).

Do you have a specific reason to need `mocha@localhost` as the email address, or do you just need a test key? Please note that the email is not required, passing only a name also works. You can even pass the email as the name, if you want.

Issue #1332, opened 2021-06-14
Cannot import library with Vite


langovoi:
I try to use `openpgp@5.0.0-3` with [Vite](https://github.com/vitejs/vite), but got error.

<img width="379" alt="image" src="https://user-images.githubusercontent.com/5055595/121895836-53e80580-cd29-11eb-9616-ea22397cced5.png">
<img width="1089" alt="image" src="https://user-images.githubusercontent.com/5055595/121895871-5f3b3100-cd29-11eb-96aa-9b3aae1b81ad.png">

twiss:
Hello :wave: Probably there's a step in your build process that replaces `process.env.NODE_ENV` with `"development"`, but it doesn't quite work because we use `globalThis.process.env.NODE_ENV` instead: https://github.com/openpgpjs/openpgpjs/blob/0ca83cf1218738f997a5274bd342c58fd01df5f3/src/util.js#L29

I'm not sure what the best fix here is, perhaps this could be tweaked in Vite, or we could modify it slightly, but I believe the code as-is works in Webpack (but perhaps we should double check this).

langovoi:
> Probably there's a step in your build process that replaces `process.env.NODE_ENV` with `"development"`

With is default non-configurable step in Vite. So I also create issue in Vite — vitejs/vite/issues/3801

lee-orr:
@twiss - would you be willing to integrate a PR to make debug mode a manual step for users of the library, rather than building it in like this? It turns out that the Vite "fix" fixes it in a build, but not in dev mode. But beyond that, I can see situations where I might want to manually enable debug mode in a production setting, or disable it in a dev setting. What I was thinking is exporting a single function - `setDebugMode(boolean)` from the openpgp module, and having it set the debugMode variable. I haven't attempted to implement it yet, since its a change in expected behaviour so I wanted to see if it's acceptable.

twiss:
I'd prefer not to, if possible. Would it be possible to figure out on the Vite side why it works in build but not in dev mode? It seems strange to me, and perhaps it could be fixed in the same way?

lee-orr:
[...]

PR #1402, opened 2021-08-17
Enable use with Vite

lee-orr:
Adjust the way openpgp accesses the NODE_ENV, so that it works in both webpack and vite. should close: https://github.com/openpgpjs/openpgpjs/issues/1332

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -26,7 +26,10 @@
 import * as stream from '@openpgp/web-stream-tools';
 import { getBigInteger } from './biginteger';
 
-const debugMode = globalThis.process && globalThis.process.env.NODE_ENV === 'development';
+
+const envVar = 'NODE_ENV';
+
+const debugMode = globalThis.process && globalThis.process.env[envVar] === 'development';
 
 const util = {
   isString: function(data) {


twiss:
If at all possible, I would prefer not to add a workaround for this in OpenPGP.js. It seems like it should already be fixed in Vite? (See https://github.com/vitejs/vite/issues/3801#issuecomment-900545204).

lee-orr:
I'm going to close the PR - it seems like there is another solution. It still doesn't work when calling `import * as openpgp from "openpgp";`, but it works when calling `import * as openpgp from "openpgp/dist/openpgp";`. Not sure why.

lee-orr:
Re-opened this following the discussion here: https://github.com/openpgpjs/openpgpjs/issues/1332

lee-orr:
@twiss Can we please bring this in? It's a minor change that will not affect current behavior with webpack at all, but will enable this to work properly in Vite. I did try to fix the issue on the Vite side, but haven't been able to do so, and don't know if/when they will fix it at all.

twiss:
:+1: Thanks!

PR #1410, opened 2021-08-25
Replace strings with integer algorithm identifiers in packet classes

larabr:
In several packet classes, we used to store string identifiers for public-key, aead, cipher or hash algorithms. To make the code consistent and to avoid having to convert to/from string values, we now always store integer values instead, e.g. `enums.symmetric.aes128` is used instead of `'aes128'`.

This is not expected to be a breaking change for most library users. Note that the type of `Key.getAlgorithmInfo()` and of the session key objects returned and accepted by top-level functions remain unchanged.

Affected classes (type changes for some properties and method's arguments):
- `PublicKeyPacket`, `PublicSubkeyPacket`, `SecretKeyPacket`, `SecretSubkeyPacket`
- `SymEncryptedIntegrityProtectedDataPacket`, `AEADEncryptedDataPacket`, `SymmetricallyEncryptedDataPacket`
- `LiteralDataPacket`, `CompressedDataPacket`
- `PublicKeyEncryptedSessionKey`, `SymEncryptedSessionKeyPacket`
- `SignaturePacket`

Other potentially breaking changes:
- Removed property `AEADEncryptedDataPacket.aeadAlgo`, since it was redudant given `.aeadAlgorithm`.
- Renamed `AEADEncryptedDataPacket.cipherAlgo` -> `.cipherAlgorithm`

TODO:
- [x] TS

### DIFF
--- a/src/crypto/hash/index.js
+++ b/src/crypto/hash/index.js
@@ -16,6 +16,7 @@ import * as stream from '@openpgp/web-stream-tools';
 import md5 from './md5';
 import util from '../../util';
 import defaultConfig from '../../config';
+import enums from '../../enums';
 
 const webCrypto = util.getWebCrypto();
 const nodeCrypto = util.getNodeCrypto();
@@ -143,18 +144,18 @@ export default {
    */
   getHashByteLength: function(algo) {
     switch (algo) {
-      case 1: // - MD5 [HAC]
+      case enums.hash.md5: // - MD5 [HAC]
         return 16;
-      case 2: // - SHA-1 [FIPS180]
-      case 3: // - RIPE-MD/160 [HAC]
+      case enums.hash.sha1: // - SHA-1 [FIPS180]
+      case enums.hash.ripemd: // - RIPE-MD/160 [HAC]
         return 20;
-      case 8: // - SHA256 [FIPS180]
+      case enums.hash.sha256: // - SHA256 [FIPS180]
         return 32;
-      case 9: // - SHA384 [FIPS180]
+      case enums.hash.sha384: // - SHA384 [FIPS180]
         return 48;
-      case 10: // - SHA512 [FIPS180]
+      case enums.hash.sha512: // - SHA512 [FIPS180]
         return 64;
-      case 11: // - SHA224 [FIPS180]
+      case enums.hash.sha224: // - SHA224 [FIPS180]
         return 28;
       default:
         throw new Error('Invalid hash algorithm.');
--- a/src/crypto/hash/index.js
+++ b/src/crypto/hash/index.js
@@ -111,26 +111,19 @@ export default {
    */
   digest: function(algo, data) {
     switch (algo) {
-      case 1:
-        // - MD5 [HAC]
+      case enums.hash.md5:
         return this.md5(data);
-      case 2:
-        // - SHA-1 [FIPS180]
+      case enums.hash.sha1:
         return this.sha1(data);
-      case 3:
-        // - RIPE-MD/160 [HAC]
+      case enums.hash.ripemd:
         return this.ripemd(data);
-      case 8:
-        // - SHA256 [FIPS180]
+      case enums.hash.sha256:
         return this.sha256(data);
-      case 9:
-        // - SHA384 [FIPS180]
+      case enums.hash.sha384:
         return this.sha384(data);
-      case 10:
-        // - SHA512 [FIPS180]
+      case enums.hash.sha512:
         return this.sha512(data);
-      case 11:
-        // - SHA224 [FIPS180]
+      case enums.hash.sha224:
         return this.sha224(data);
       default:
         throw new Error('Invalid hash function.');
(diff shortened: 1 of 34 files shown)


src/crypto/hash/index.js
```diff
@@ -143,18 +144,18 @@ export default {
    */
   getHashByteLength: function(algo) {
     switch (algo) {
-      case 1: // - MD5 [HAC]
+      case enums.hash.md5: // - MD5 [HAC]
```
twiss, line 147: Could you update the above `digest` function in the same way? Also, I think these comments aren't really necessary anymore, now.
```suggestion
      case enums.hash.md5:
```

src/crypto/mode/gcm.js
```diff
@@ -36,12 +37,14 @@ const ALGO = 'AES-GCM';
...
+ * @param {enums.symmetric} cipher - The symmetric cipher algorithm to use
  * @param {Uint8Array} key - The encryption key
  */
 async function GCM(cipher, key) {
-  if (cipher.substr(0, 3) !== 'aes') {
-    throw new Error('GCM mode supports only AES cipher');
+  if (cipher !== enums.symmetric.aes128 &&
+    cipher !== enums.symmetric.aes192 &&
+    cipher !== enums.symmetric.aes256) {
+    throw new Error('EAX mode supports only AES cipher');
```
twiss, line 47:
```suggestion
    throw new Error('GCM mode supports only AES cipher');
```
These error messages are quite broken English btw, but ok, that's a bit unrelated

src/packet/aead_encrypted_data.js
```diff
@@ -64,17 +65,19 @@ class AEADEncryptedDataPacket {
...
   async read(bytes) {
     await stream.parse(bytes, async reader => {
       const version = await reader.readByte();
       if (version !== VERSION) { // The only currently defined value is 1.
         throw new UnsupportedError(`Version ${version} of the AEAD-encrypted data packet is not supported.`);
       }
-      this.cipherAlgo = await reader.readByte();
-      this.aeadAlgo = await reader.readByte();
+      this.cipherAlgorithm = enums.write(enums.symmetric, await reader.readByte());
+      this.aeadAlgorithm = enums.write(enums.aead, await reader.readByte());
```
twiss, line 77: I would prefer not to get into the habit of using `enums.write` for checking an integer value. I know it supports it, but I think we should get a bit more strict about the types, and use `read` only for `int` -> `string`, and `write` only for `string` -> `int`. And then we can do validation in some other way - for example, in this case we will already throw below, in the `read` call in `getAEADMode`, if the AEAD mode is unknown (and similarly for the cipher, but during decryption).

src/packet/compressed_data.js
```diff
@@ -85,7 +85,7 @@ class CompressedDataPacket {
     await stream.parse(bytes, async reader => {
 
       // One octet that gives the algorithm used to compress the packet.
-      this.algorithm = enums.read(enums.compression, await reader.readByte());
+      this.algorithm = enums.write(enums.compression, await reader.readByte());
```
twiss, line 88: Similarly here - I would just read the byte during parsing, and then if we don't support the algorithm we can complain later, during decompression.

src/packet/one_pass_signature.js
```diff
@@ -83,13 +87,13 @@ class OnePassSignaturePacket {
 
     // A one-octet signature type.  Signature types are described in
     //   Section 5.2.1.
-    this.signatureType = bytes[mypos++];
+    this.signatureType = enums.write(enums.signature, bytes[mypos++]);
```
twiss, line 90: The same thing pretty much goes for the rest of the PR

src/packet/public_key.js
```diff
@@ -115,8 +115,11 @@ class PublicKeyPacket {
       pos += 4;
 
       // - A one-octet number denoting the public-key algorithm of this key.
-      this.algorithm = enums.read(enums.publicKey, bytes[pos++]);
-      const algo = enums.write(enums.publicKey, this.algorithm);
+      try {
+        this.algorithm = enums.write(enums.publicKey, bytes[pos++]);
```
twiss, line 119: Here it's a bit of a special case, since knowing the public key algorithm is required for parsing it. But I would let `parsePublicKeyParams` throw in that case, as it already does. (Even though I know the old code also already kind of did this.)

src/packet/aead_encrypted_data.js
twiss: Conceptually speaking, they're not necessarily "wrong enums", it might be that it's an algorithm that we just don't support. In other words, the spec does not forbid any values here, it's just that we might not be able to decrypt all of them. And sometimes it might be useful to parse the packet, e.g. when building a packet listing, and just show it with an "unsupported algorithm", rather than showing the entire packet as "unsupported packet".

> with potentially unexpected side effects

What side effects are you thinking of?

(There also shouldn't be a need for explicit exceptions, the "rule" would just be "throw when you use the enum value if we don't support it.")

twiss (approved):

PR #1411, opened 2021-09-01
CI: Add performance regression monitoring for pull requests

larabr:
Add tests to monitor performance of the main top-level functions, and run them on pull requests:
- if performance is 30% worse than the master branch, an alert comment will be left on the PR commit. A lower threshold risks giving false positive regression alerts: during testing, I've observed a variability of up to 20% when testing the same code.
- if performance is 50% worse, the workflow will fail.

For example workflow behaviour and alert comments, see https://github.com/larabr/openpgpjs/pull/2.


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,45 @@
...
+          name: 'time benchmark'
+          pr-benchmark-file-path: pr/benchmarks.txt
+          base-benchmark-file-path: master/benchmarks.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # trigger alert comment if 1.3 times slower
+          alert-threshold: '130%'
+          comment-on-alert: true
+          # fail workdlow if 1.5 times slower
+          fail-threshold: '150%'
+          fail-on-alert: true
```
twiss, line 45: Newline, pls :)

test/benchmarks/benchmark.js
```diff
@@ -0,0 +1,97 @@
...
+const wrapAsync = func => ({
+  fn: async deferred => {
+    await func().catch(onError);
+    deferred.resolve();
+  },
+  defer: true
+});
+
+const onError = err => {
+  console.error('The time benchmark tests failed by throwing the following error:');
```
twiss, line 13: Please add a `/* eslint-disable no-console */` at the top of the file or so, to stop it from warning about this

Issue #1363, opened 2021-07-04
"export 'stream' (imported as 'openpgp') was not found in 'openpgp'


ehte92:
Getting a warning in the latest version when I am using the openpgp.stream and also getting a type error

`TypeError: Cannot read property 'readToEnd' of undefined`

I am using the following to read the stream to the end and store it in a variable.

`const plainStr = await openpgp.stream.readToEnd(decryptBlob.data);`

Is there any new changes in the new version?

larabr:
Hi @ehte92 ,  `openpgp.stream` was unexported in 5.0.0-2, see https://github.com/openpgpjs/openpgpjs/pull/1291 .

PR #1413, opened 2021-09-04
Remove stream from type definitions

sinderw:
As stated in #1291, `stream` namespace have been unexported from openpgpjs.

#1363 described the problem with typings: although unexported, the type still exists in the type definition file.

Note: for those needing `stream.readToEnd`, a workaround is described in #1365.


twiss:
:+1: Thanks for the PR!

PR #1415, opened 2021-09-09
CI: Add memory usage regression monitoring for pull requests

### LINKED ISSUE #403: Evaluate how to monitor memory usage and implement a unit test for it
While following the development of the typed array branch i'm thinking that an important nice to have in our re-testing setup is a monitor for the memory usage of openpgp.js in order to be sure that:
- no memory leaks exist in the library and eventually spot them early;
- change in implementations like passing from string to typed array and vice versa does not cause important changes in memory allocations.

I've already thought a little about this and i'm imagining that probably in some unit tests we can simply use `process.memoryUsage()` when test are run using Node
and put some assetions in order to see that the memory allocation is the one that we expect to be.

https://nodejs.org/api/process.html#process_process_memoryusage

larabr:
Fix #403 : add tests to monitor memory usage of message encryption and decryption. Each test is run in a separate V8 instance, to make the result as deterministic as possible. These tests are run on pull requests:
- if the peak memory usage (RSS) is 2% worse than the master branch, an alert comment will be left on the PR commit.
- if the peak memory usage (RSS) is 10% worse, the workflow will fail.

TODO:

- [ ] merge after #1411 
- [ ] point to release branch of benchmark action


test/benchmarks/memory_usage.js
```diff
@@ -0,0 +1,219 @@
...
+  constructor() {
+    this.tests = [];
+  }
+
+  add(name, fn) {
+    this.tests.push({ name, fn });
+  }
+
+  async run() {
+    const stats = []; // the size of this data should be negligible compared to the tests
```
larabr, line 83:
```suggestion
    const stats = [];
```

Issue #579, opened 2017-09-12
Explicitly requires string for passphrase in S2K.produce_key


FocuzJS:
https://github.com/openpgpjs/openpgpjs/blob/bc69357ab55d16634dbcbf5e5f7a3643579697ec/src/type/s2k.js#L145

I'm proposing that we should be able pass a `Uint8Array` as the SecretKey passphrase (ex: SecretKey.decrypt) as it will handle it without throwing a String onto the heap.

The library already is ready for this functionality except for this single instance on line 145 that expects string.

One example fix could be:

```js
passphrase = (typeof(passphrase) === 'object') ? passphrase : util.str2Uint8Array(util.encode_utf8(passphrase));
```

Edit: This same line is used during generation of private keys, therefore changing this line allows `Uint8Array` to be passed into .decrypt of a private key **and** as the passphrase when generating key pairs.

twiss:
Closing, see https://github.com/openpgpjs/openpgpjs/pull/1436#issuecomment-965003168.

PR #1436, opened 2021-11-04
Allow Uint8Array passphrases

JDinABox:
Implements proposal from #579


twiss:
Hello :wave: Sorry for the delay, and thanks for the PR!

May I ask what's your use case for this? Is it primarily performance? (If so, I'm somewhat skeptical of the performance gains, as this immediately precedes the "string-to-key" (S2K) password hashing, which is much slower.)

JDinABox:
Hi, thanks for the reply.

This PR is primarily to allow non-string passphrases (For instance, a passphrase already run through a KDF) to be passed as values without increasing complexity with encoding.
I agree that the performance benefits are negligible compared to the S2K's cost.

twiss:
I see, thanks. The issue is that the passphrase is really meant to be a string (it's even called "string-to-key" :sweat_smile:), and I believe for example GnuPG also only supports passing passphrases as a string, and encodes it as UTF-8. So - it may be better to just encode the key as a string (e.g. base64 or hex), so that you don't run into issues with other libraries later.

JDinABox:
Okay, that makes sense.
Thanks for taking a look at this, closing.

PR #1439, opened 2021-11-12
Update web stream tools and improve performance

### LINKED ISSUE #1424: Performance Issue During Signature Verification
```
const readableStream = fs.createReadStream('enc.zip.pgp');
readableStream.setEncoding('UTF8')

const message = await openpgp.readMessage({
      armoredMessage: readableStream 
});

const { data: decrypted, signatures } = await openpgp.decrypt({
        message,
        verificationKeys: [senderPublicKey],
        decryptionKeys: [ privateKey],
        format: 'binary'
});

try {
        await signatures[0].verified;
 } catch (err) {
        console.log(err.message);
}
```

When performing decryption + signature verification for large files (e.g. 1GB in size) using `ReadableStream` , there seems to be a performance overhead where `await signatures[0].verified;` may take up to a couple of minutes to complete. Alternatively, adding `expectSigned: true` also results in UnhandledPromiseRejection thus preventing the error from being handled gracefully - for cases where message is signed with a mismatch verification key.

twiss:
Fix #1424

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -261,9 +261,9 @@
       "dev": true
     },
     "@openpgp/web-stream-tools": {
-      "version": "0.0.6",
-      "resolved": "https://registry.npmjs.org/@openpgp/web-stream-tools/-/web-stream-tools-0.0.6.tgz",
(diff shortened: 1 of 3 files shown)


twiss:
This PR improves performance at the cost of somewhat increased memory usage, so the memory regression test failure is somewhat expected. Decreasing memory usage (without decreasing performance again :sweat_smile:) will requite more work, I will look into that separately.

larabr (approved):

PR #1445, opened 2021-12-01
Add support for constant-time decryption of PKCS#1 v1.5-encoded session keys

larabr:
Implement optional constant-time decryption flow to hinder Bleichenbacher-like attacks against RSA- and ElGamal public-key encrypted session keys.

Changes:
- Add `config.constantTimePKCS1Decryption` to enable the constant-time processing (defaults to `false`). The constant-time option is off by default since it has measurable performance impact on message decryption, and it is only helpful in specific application scenarios (more info below).
- Add `config.constantTimePKCS1DecryptionSupportedSymmetricAlgorithms` (defaults to the AES algorithms). The set of supported ciphers is restricted by default since the number of algorithms negatively affects performance.

Bleichenbacher-like attacks are of concern for applications where both of the following conditions are met:
   1. new/incoming messages are automatically decrypted (without user interaction);
   2. an attacker can determine how long it takes to decrypt each message (e.g. due to decryption errors being logged remotely).

### DIFF
--- a/src/packet/public_key_encrypted_session_key.js
+++ b/src/packet/public_key_encrypted_session_key.js
@@ -138,8 +138,8 @@ class PublicKeyEncryptedSessionKeyPacket {
       // We must not leak info about the validity of the decrypted checksum or cipher algo.
       // The decrypted session key must be of the same algo and size as the random session key, otherwise we discard it and use the random data.
       const isValidPayload = isValidChecksum & symmetricAlgoByte === randomSessionKey.sessionKeyAlgorithm & sessionKey.length === randomSessionKey.sessionKey.length;
-      this.sessionKey = isValidPayload ? sessionKey : randomSessionKey.sessionKey;
-      this.sessionKeyAlgorithm = isValidPayload ? symmetricAlgoByte : randomSessionKey.sessionKeyAlgorithm;
+      this.sessionKeyAlgorithm = util.selectUint8(isValidPayload, symmetricAlgoByte, randomSessionKey.sessionKeyAlgorithm);
+      this.sessionKey = util.selectUint8Array(isValidPayload, sessionKey, randomSessionKey.sessionKey);
 
     } else {
       const isValidPayload = isValidChecksum && enums.read(enums.symmetric, symmetricAlgoByte);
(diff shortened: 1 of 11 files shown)


src/packet/public_key_encrypted_session_key.js
```diff
@@ -110,25 +110,45 @@ class PublicKeyEncryptedSessionKeyPacket {
...
+      randomSessionKey.sessionKey,
+      util.writeChecksum(randomSessionKey.sessionKey)
+    ]) : null;
+    const decoded = await crypto.publicKeyDecrypt(this.publicKeyAlgorithm, key.publicParams, key.privateParams, this.encrypted, key.getFingerprintBytes(), randomPayload);
     const checksum = decoded.subarray(decoded.length - 2);
     const sessionKey = decoded.subarray(1, decoded.length - 2);
-    if (!util.equalsUint8Array(checksum, util.writeChecksum(sessionKey))) {
-      throw new Error('Decryption error');
+    const symmetricAlgoByte = decoded[0];
+    const isValidChecksum = util.equalsUint8Array(checksum, util.writeChecksum(sessionKey));
```
twiss, line 135: `equalsUint8Array` is not constant-time - since both are guaranteed to be two bytes, I would just hardcode an equality check for those two bytes here

src/packet/public_key_encrypted_session_key.js
```diff
@@ -129,10 +129,11 @@ class PublicKeyEncryptedSessionKeyPacket {
       util.writeChecksum(randomSessionKey.sessionKey)
     ]) : null;
     const decoded = await crypto.publicKeyDecrypt(this.publicKeyAlgorithm, key.publicParams, key.privateParams, this.encrypted, key.getFingerprintBytes(), randomPayload);
-    const checksum = decoded.subarray(decoded.length - 2);
-    const sessionKey = decoded.subarray(1, decoded.length - 2);
     const symmetricAlgoByte = decoded[0];
-    const isValidChecksum = util.equalsUint8Array(checksum, util.writeChecksum(sessionKey));
+    const sessionKey = decoded.subarray(1, decoded.length - 2);
+    const expectedChecksum = decoded.subarray(decoded.length - 2);
+    const actualChecksum = util.writeChecksum(sessionKey);
```
twiss, line 135: This is increasingly nitpicky but I think these are named the wrong way around - I would just keep the first one as `checksum` and name the second one `expectedChecksum` or `computedChecksum`, maybe

twiss (approved):

Issue #1449, opened 2021-12-06
Out of memory when decrypting large files


yegorpetrov:
- OpenPGP.js version: 5.0.1
- Affected platform (Browser or Node.js version): both?

OpenPGP.js is crashing with OOM issues when I feed it files larger than about 100 MB. Granted, it's working in a resource-constained environment, but I wouldn't expect memory consumption to be proportional to the input file size in general.

So, I found this line that tries to swallow the entire input stream into memory before passing it to the decryption stage: [`await stream.readToEnd(stream.clone(this.encrypted))`](https://github.com/openpgpjs/openpgpjs/blob/d6139daa79accba980796a78454f5db17b38fdbf/src/packet/symmetrically_encrypted_data.js#L90-L94). Naturally, it fails for larger files. And for smaller ones you get all decrypted data inside one large chunk. It does not care about things like `allowUnauthenticatedStream: true` and `allowUnauthenticatedMessages: true` (not sure which one is applicable).

A limited-use workaround is to encrypt files with MDC enabled, so that they're processed by [a different class](https://github.com/openpgpjs/openpgpjs/blob/d6139daa79accba980796a78454f5db17b38fdbf/src/packet/sym_encrypted_integrity_protected_data.js). In other words, when I encrypt files using `gpg --force-mdc ...` they don't cause any similar issues when decrypted with this package.

I played around with the source code a little and came up with the following to replace that block of code:

    if (stream.isArrayStream(this.encrypted)) {
      // TODO remove this special case when ArrayStream.slice() is implemented in @openpgp/web-stream-tools
      const encrypted = await stream.readToEnd(stream.clone(this.encrypted));
      decrypted = await crypto.mode.cfb.decrypt(sessionKeyAlgorithm, key,
        encrypted.subarray(blockSize + 2),
        encrypted.subarray(2, blockSize + 2)
      );
    } else {
      decrypted = await crypto.mode.cfb.decrypt(sessionKeyAlgorithm, key,
        stream.slice(stream.clone(this.encrypted), blockSize + 2),
        await stream.readToEnd(stream.slice(stream.clone(this.encrypted), 2, blockSize + 2))
      );
    }

This way, it seems to deliver properly-chunked decrypted output without choking on large files. Does this look like a proper solution?

larabr:
I am not following what kind of files you are trying to decrypt.
[...]

Issue #1012, opened 2019-12-03
Streaming decryption fails on files larger than 2GiB


ama4tf:
Running decrypt on large files fails due to array overrun. This was generated by calling openpgp.decrypt using an encrypted file-stream of 2.1gb in length, a private key, and the binary format. Failure message is below:

Error decrypting message: Invalid typed array length: 2322369312

twiss:
Hey :wave: This is a limitation of Chrome / V8; it has a max Uint8Array size of 2GiB. It's possible that you could work around this by using streaming decryption: https://github.com/openpgpjs/openpgpjs/releases/tag/v4.0.0.

ama4tf:
We found a workaround by decrypting the larger files in bash but still not a clean fix. I tried switching to streaming both the input encrypted file and output decrypted stream but still yields the same error

twiss:
@ama4tf For the streaming solution, please see https://github.com/openpgpjs/openpgpjs/releases/tag/v4.0.0, specifically "Streaming CFB". I think it should be possible to stream files larger than 2GB, although I haven't tried it.

ama4tf:
I tried Streaming CFB and that did yield a different kind of error, so still not successful but maybe progress

`RangeError [ERR_INVALID_OPT_VALUE]: The value "3148715314" is invalid for option "size"`

Traceback on the error looks like this

at Object.concat (\node_modules\openpgp\dist\openpgp.js:24175:23)
at Reader.readBytes (\node_modules\openpgp\dist\openpgp.js:24067:51)
at async Reader.peekBytes (\node_modules\openpgp\dist\openpgp.js:24086:17)
at async Object.read (\node_modules\openpgp\dist\openpgp.js:38472:26)
at async \node_modules\openpgp\dist\openpgp.js:38571:22

twiss:
OK yeah, this is a bug in our streaming implementation, that we should fix.

lekhacman:
Hi @twiss, may i fix this?

twiss:
Hey @lekhacman, yes of course, if you want to take a stab at it feel free.

AFAIK, the root cause of this error is this line: https://github.com/openpgpjs/openpgpjs/blob/25118c3f2859c4f5b4c911e7fcd495e20d1cce4f/src/packet/packet.js#L272

[...]

Issue #1442, opened 2021-11-17
Implement specification changes from `draft-ietf-openpgp-crypto-refresh`


larabr:
The crypto-refresh document is still being drafted (https://datatracker.ietf.org/doc/draft-ietf-openpgp-crypto-refresh/), but we should implement/update the new features before its release.

To update/change:
- [ ] AEAD key encryption
...

To add from scratch:
- [ ] Curve448
- [ ] Argon2
...

PR #1462, opened 2022-01-07
Decryption of Large Files/Streams - Update documentation and Guard Test case

justin-lovell:
# Introduction

We are working on larger files which result in the process crashing with the following stack trace, and reproduced with the following commit: https://github.com/justin-lovell/openpgpjs/commit/68c7c29bd140202591550c20d895af426d769e07
```
node:buffer:602
    slice: (buf, start, end) => buf.ucs2Slice(start, end),
                                    ^

Error: Failed to allocate memory
    at Object.slice (node:buffer:602:37)
    at Uint8Array.toString (node:buffer:811:14)
    at TextDecoder.decode (node:internal/encoding:431:18)
    at r (C:\src\card-data-importer-poc\node_modules\openpgp\dist\node\openpgp.min.js:2:14559)
    at C:\src\card-data-importer-poc\node_modules\openpgp\dist\node\openpgp.min.js:2:6762
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  code: 'ERR_MEMORY_ALLOCATION_FAILED'
```

This has been a long standing issue as per the following issues:
* https://github.com/openpgpjs/openpgpjs/issues/1012
* https://github.com/openpgpjs/openpgpjs/issues/1449

# Solution

The reason why memory allocations grow is due to the eager loading of the stream to perform integrity checks on the checksum. The key lines are the following:

https://github.com/openpgpjs/openpgpjs/blob/6da1c53de7852000fd63bacf1809b7e1ea17cf11/src/packet/sym_encrypted_integrity_protected_data.js#L137-L139

There is an `option` which may be set for `allowUnauthenticatedStream` -- this is largely undocumented and not mentioned. This PR would advise on the option and provide an assertion that the client should perform a double stream pass to ensure the integrity check had the opportunity to verify the packet.

# Additional Test Case

An additional test case was added to guard against an accidental regression for processing large files.

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -93,16 +93,10 @@ export default {
   allowUnauthenticatedMessages: false,
   /**
    * Allow streaming unauthenticated data before its integrity has been checked. This would allow the application to
-   * process very large streams by deferring the checksum integrity check to be only verified at the end of the stream.
+   * process large streams while limiting memory usage by releasing the decrypted chunks as soon as possible
+   * and deferring checking their integrity until the decrypted stream has been read in full.
    *
    * This setting is **insecure** if the partially decrypted message is processed further or displayed to the user.
-   *
-   * Therefore, the client application is advised to implement at least one of the following strategies when processing
-   * very large streams.
-   * - Perform a two-phase commit -- process the data in a staged area, and when the stream finalizes, commit if no error
-   *   occurs.
-   * - Read the stream twice -- read the contents to the end of the stream, and if the data is verified to be correct,
-   *   then perform a second pass on the stream to perform the real work.
    * @memberof module:config
    * @property {Boolean} allowUnauthenticatedStream
    */
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -92,8 +92,17 @@ export default {
    */
   allowUnauthenticatedMessages: false,
   /**
-   * Allow streaming unauthenticated data before its integrity has been checked.
+   * Allow streaming unauthenticated data before its integrity has been checked. This would allow the application to
+   * process very large streams by deferring the checksum integrity check to be only verified at the end of the stream.
+   *
    * This setting is **insecure** if the partially decrypted message is processed further or displayed to the user.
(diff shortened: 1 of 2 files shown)


justin-lovell:
Also, the changes are not going to allow options to bypass security features. For large files, having the checksum at the end of the file and NodeJS streams not being seekable is rather annoying.

justin-lovell:
After annotating and fully understanding the library, the key preventing streaming of large payloads comes down to the option of `allowUnauthenticatedStream`.

https://github.com/openpgpjs/openpgpjs/blob/6da1c53de7852000fd63bacf1809b7e1ea17cf11/src/packet/sym_encrypted_integrity_protected_data.js#L137-L139

As such, this PR will be changed to document/prove that large blobs can be handled by `openpgp.js` and make the appropriate annotations on the documentation.

larabr (changes requested):
Thanks for proposing the changes and adding the test. We were looking into the best way to benchmark using large data, and using a generator does seem to work well!
I left some comments, since the code needs some changes to reduce variability of benchmark results across runs.
I would also add a second version of the same test, but setting `allowUnauthenticatedStream = false` (like the version you committed before). For that reason, I left a comment on the "outdated" code too.

About documentation, I'd rather not include "how-to" information in the source code, as it will still miss a lot of context. Instead, we are planning to add a Wiki page about streaming with all the relevant info. This is also because, for instance, `config.allowUnauthenticatedStream` will become less relevant once the new OpenPGP spec is released (#1442), thanks to better support for streaming at the protocol level.

larabr (approved):
Thanks for the contribution @justin-lovell . I've taken over since it's been some time.

In addition to adding tests over large files, I uniformed data generation across all streaming tests to be able to compare the results. I also added output stream consumption, otherwise the decryption operation isn't actually triggered in all cases (this was already an issue with the existing AEAD streaming tests as well).

twiss (approved):

PR #1463, opened 2022-01-24
Moved functionality to check for Edge browser to shimmable function in util object

renatoathaydes:
Moved functionality to check for Edge browser to util object.
Cache value to avoid computing it several times.
This makes the function shimmable as well
(so it can be replaced in non-browser environments).

### DIFF
--- a/src/crypto/hash/index.js
+++ b/src/crypto/hash/index.js
@@ -77,7 +77,7 @@ if (nodeCrypto) { // Use Node native crypto for all hash functions
 } else { // Use JS fallbacks
   hashFunctions = {
     md5: md5,
-    sha1: asmcryptoHash(Sha1, (!navigator.userAgent || navigator.userAgent.indexOf('Edge') === -1) && 'SHA-1'),
+    sha1: asmcryptoHash(Sha1, !util.isEdge() && 'SHA-1'),
     sha224: hashjsHash(sha224),
     sha256: asmcryptoHash(Sha256, 'SHA-256'),
     sha384: hashjsHash(sha384, 'SHA-384'),
--- a/src/crypto/mode/eax.js
+++ b/src/crypto/mode/eax.js
@@ -51,7 +51,7 @@ async function CTR(key) {
   if (
     util.getWebCrypto() &&
     key.length !== 24 && // WebCrypto (no 192 bit support) see: https://www.chromium.org/blink/webcrypto#TOC-AES-support
-    (!navigator.userAgent || navigator.userAgent.indexOf('Edge') === -1)
+    (!util.isEdge())
   ) {
     key = await webCrypto.importKey('raw', key, { name: 'AES-CTR', length: key.length * 8 }, false, ['encrypt']);
     return async function(pt, iv) {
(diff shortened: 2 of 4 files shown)


renatoathaydes:
Will still test  that I can shim this easily with webpack.

renatoathaydes:
This PR helps people work around this bug: https://github.com/openpgpjs/openpgpjs/issues/1464

renatoathaydes:
I was able to remove the navigator using Webpack, I will post what I did here soon... but maybe there's still a better way to handle this. I am not a front-end developer myself so I might be missing something.

twiss:
Thanks for the PR! However, I think this is indeed not the simplest way to fix your issue.

In fact, since Edge Legacy is no longer supported by MS, I think we can drop support for it as well, and remove all uses of `navigator.userAgent` altogether.

However, there's still one other use of `navigator`, here: https://github.com/openpgpjs/openpgpjs/blob/22a68878e499454806df78d63f99ec2d7f177f96/src/util.js#L433

That could be changed to something like
```js
    return typeof navigator !== 'undefined' && navigator.hardwareConcurrency || 1;
```

renatoathaydes:
I noticed that hardwareConcurrency setting as well, but managed to also remove that from the bundled JS code so I didn't worry about that.
I think the important thing is to keep access to these "browser-only" variables in a single place where it's easy to find them and replace them if necessary.

> However, I think this is indeed not the simplest way to fix your issue.

I mean, without changing the functionality, do you have a better way?

If you intend to remove the Edge check, even better, you can decline this PR.

twiss:
> I mean, without changing the functionality, do you have a better way?

Well, even without removing Edge support, we could just check for `typeof navigator !== 'undefined'`, yeah. Then there's no need to shim anything.

> If you intend to remove the Edge check, even better, you can decline this PR.

But yeah, fair enough. I'll close this, then :+1:

renatoathaydes:
> typeof navigator !== 'undefined'

That still won't work on Cloudflare... they limit which APIs are available on their runtime so we really do need to get rid of all references to `navigator`, even if just checking if it's defined.

Here's my webpack config to get rid of it:

```js
  plugins: [
    new webpack.ProvidePlugin({
      'navigator.userAgent': [path.resolve(path.join(__dirname, 'src/shims.js')), 'userAgent'],
      'navigator.hardwareConcurrency': [path.resolve(path.join(__dirname, 'src/shims.js')), 'hardwareConcurrency'],
    }),
  ],
```

And the `src/shims.js` file:

```js
var userAgent = 'none'
var hardwareConcurrency = 1
```

This may help some other poor dev one day to not waste as much as as I had to :).

renatoathaydes:
Notice that the replacements only work after my changes in this PR... to make it work with the current code is difficult because the `navigator` access is spread all over. I still think you should accept my changes until you're able to completely remove that.

twiss:
If you want a temporary workaround, I think you could just create a global `navigator` variable, or shim the entire `navigator` object to `{}`, perhaps. But I'll try to remove the checks for Edge soon.

PR #1474, opened 2022-02-01
Drop MS Edge Legacy support

twiss:


twiss:
There is, it says

> The `dist/openpgp.min.js` bundle works well with recent versions of Chrome, Firefox, Safari and Edge.

which holds true after this PR :)

PR #1475, opened 2022-02-01
Check existence of `navigator` before using it

### LINKED ISSUE #1464: Cannot deploy to Cloudflare Worker due to dependency on browser-only "navigator"
When I try to use this lib in a Cloudflare Worker, the deployment fails due to the `navigator` variable being used in several places in the code.

[My question on StackOverflow](https://stackoverflow.com/questions/70823034/how-to-use-openpgp-js-on-node-js-cloudflare-worker-error-uncaught-referencee) has more details.

I have imports like this:

```typescript
import { encrypt, readKey, createMessage, decryptKey, readPrivateKey } from 'openpgp/lightweight';
```

And my npm dependency is on this merged `main` branch, so I expected it would work (after fixes related to Deno):

```json
    "node_modules/openpgp": {
      "version": "5.1.0",
      "resolved": "git+ssh://git@github.com/openpgpjs/openpgpjs.git#bd13edfc884cf6b6b7f715099529ac6a656733c6",
      "license": "LGPL-3.0+",
      "dependencies": {
        "asn1.js": "^5.0.0"
      },
      "engines": {
        "node": ">= 8.0.0"
      }
    },
```

The error I get when I run `wrangler dev`:

```
webpack 5.67.0 compiled successfully in 1785 ms
Error: Something went wrong with the request to Cloudflare...
Uncaught ReferenceError: navigator is not defined
  at line 782 in ./node_modules/openpgp/dist/lightweight/openpgp.min.mjs
  at line 808 in __webpack_require__
  at line 528 in ./src/crypto.ts
  at line 808 in __webpack_require__
  at line 652 in ./src/handler.ts
  at line 808 in __webpack_require__
  at line 1047
  at line 1052
  at line 1054
 [API code: 10021]
```

This happens with `openpgp/lightweight` and `openpgp`.

twiss:
Fix #1464. Depends on #1474.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -430,7 +430,8 @@ const util = {
       return os.cpus().length;
     }
 
-    return navigator.hardwareConcurrency || 1;
+    // eslint-disable-next-line no-mixed-operators
+    return typeof navigator !== 'undefined' && navigator.hardwareConcurrency || 1;
   },
 
   isEmailAddress: function(data) {
--- a/src/util.js
+++ b/src/util.js
@@ -430,8 +430,7 @@ const util = {
       return os.cpus().length;
     }
 
-    // eslint-disable-next-line no-mixed-operators
-    return typeof navigator !== 'undefined' && navigator.hardwareConcurrency || 1;
+    return (typeof navigator !== 'undefined' && navigator.hardwareConcurrency) || 1;
   },
 
   isEmailAddress: function(data) {


src/util.js
```diff
@@ -430,7 +430,8 @@ const util = {
       return os.cpus().length;
     }
 
-    return navigator.hardwareConcurrency || 1;
+    // eslint-disable-next-line no-mixed-operators
+    return typeof navigator !== 'undefined' && navigator.hardwareConcurrency || 1;
```
larabr, line 434:
```suggestion
    return (typeof navigator !== 'undefined' && navigator.hardwareConcurrency) || 1;
```

twiss: The latter isn't quite the same in case `navigator` exists but `navigator.hardwareConcurrency` doesn't, but the former works, thanks :+1:

larabr (approved):

src/util.js
```diff
@@ -430,7 +430,7 @@ const util = {
       return os.cpus().length;
     }
 
-    return navigator.hardwareConcurrency || 1;
+    return (typeof navigator !== 'undefined' && navigator.hardwareConcurrency) || 1;
```
renatoathaydes, line 433: I appreciate you trying to fix this problem, but just so you know: it doesn't fix the Cloudfare issue because it's still accessing a prohibited API. The mere attempt to read `navigator` is what's not allowed.
I tried to explain that in my pull request.
I will try to check with cloudflare if they could allow code like this to "pass"... but I thought you would be interested in knowing that.

twiss: I understand - but even after your PR, you still had to shim `util.isEdge`, afaiu - after this PR and #1474, you can shim `util.getHardwareConcurrency` instead :)

Other than that, please do check with Cloudflare, as this is the standard way to check for the existence of global objects, it shouldn't cause issues.

PR #1490, opened 2022-02-09
issue #1489 Use this.verified instead of this[verfied]

### LINKED ISSUE #1489: Property 'verified' is not set after successful signature verification
- OpenPGP.js version: 5.1.0
- Affected platform (Browser or Node.js version): Node 12

Following code in `async verify(key, signatureType, data, ...`
```js
this[verified] = await mod.signature.verify( ...
```

breaks following use case (Typescript):
```typescript
  public static lastSig = async (key: Key): Promise<number> => {
    let i = 0;
    console.log(`Key OUID ${(Object as any).__oid(key)}`);
    await key.getExpirationTime(); // will force all sigs to be verified
    const allSignatures: SignaturePacket[] = [];
    for (const user of key.users) {
      i = 0;
      for (const sig of user.selfCertifications) {
        console.log(`user ${user.userID?.email} sig[${i++}](OUID ${(Object as any).__oid(sig)} => ${sig.verified})\n`);
      }
      allSignatures.push(...user.selfCertifications);
    }
    i = 0;
    for (const subKey of key.subkeys) {
      i = 0;
      for (const sig of subKey.bindingSignatures) {
        console.log(`subkey ${subKey.getKeyID().toHex()} sig[${i++}](OUID ${(Object as any).__oid(sig)} => ${sig.verified})}\n`);
      }
      allSignatures.push(...subKey.bindingSignatures);
    }
    allSignatures.sort((a, b) => (b.created ? b.created.getTime() : 0) - (a.created ? a.created.getTime() : 0));
    console.log(`nsig=${allSignatures.length}`);
    i = 0;
    for (const sig of allSignatures) {
      console.log(`sig[${i++}](OUID ${(Object as any).__oid(sig)} => ${sig.verified})}\n`);
    }
    const newestSig = allSignatures.find(sig => sig.verified === true);
    if (newestSig) {
      return newestSig.created ? newestSig.created.getTime() : 0;
    }
    throw new Error('No valid signature found in key');
  }
```
After calling `await key.getExpirationTime(); // will force all sigs to be verified` we expect signatures to be verified, however when enumerating signatures, `verified` is not set.
I could solve it by changing all `this[verified]` into `this.verified`.

Additional info:
Extra stuff used here:
```javascript
// Based on https://stackoverflow.com/a/1997811/1540501
// inject oids
(function() {
  if ( typeof Object.__oid == "undefined" ) {
      var id = 0;

      Object.__oid = function(o) {
          if ( typeof o.__uniqueid == "undefined" ) {
              Object.defineProperty(o, "__uniqueid", {
                  value: ++id,
                  enumerable: false,
                  // This could go either way, depending on your
                  // interpretation of what an "id" is
                  writable: false
              });
          }

          return o.__uniqueid;
      };
  }
})();

```

IvanPizhenko:
close #1489

### DIFF
--- a/src/packet/signature.js
+++ b/src/packet/signature.js
@@ -98,7 +98,7 @@ class SignaturePacket {
     this.preferredAEADAlgorithms = null;
 
     this.revoked = null;
-    this[verified] = null;
+    this.verified = null;
   }
 
   /**
@@ -206,7 +206,7 @@ class SignaturePacket {
       // getLatestValidSignature(this.revocationSignatures, key, data)` later.
       // Note that this only holds up if the key and data passed to verify are the
       // same as the ones passed to sign.
-      this[verified] = true;
+      this.verified = true;
     }
   }
 
@@ -670,7 +670,7 @@ class SignaturePacket {
     const isMessageSignature = signatureType === enums.signature.binary || signatureType === enums.signature.text;
     // Cryptographic validity is cached after one successful verification.
     // However, for message signatures, we always re-verify, since the passed `data` can change
-    const skipVerify = this[verified] && !isMessageSignature;
+    const skipVerify = this.verified && !isMessageSignature;
     if (!skipVerify) {
       let toHash;
       let hash;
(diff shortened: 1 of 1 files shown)


larabr:
Closing since `verified` is intended to be internal-only in v5. Continuing discussion on rationale and alternative solution(s) in https://github.com/openpgpjs/openpgpjs/discussions/1491 .

PR #1502, opened 2022-03-01
TS: Rely on new `web-stream-tools` types, fix `SignOptions`

larabr:
Also, add `EncryptSessionKeyOptions` to make it easier to declare wrapper functions of `encryptSessionKey`.

TODO:
- [ ] point to released web-stream-tools version

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -7,6 +7,8 @@
  *  - Errietta Kostala <https://github.com/errietta>
  */
 
+import type { WebStream as GenericWebStream, NodeStream as GenericNodeStream } from '@openpgp/web-stream-tools';
+
 /* ############## v5 KEY #################### */
 // The Key and PublicKey types can be used interchangably since TS cannot detect the difference, as they have the same class properties.
 // The declared readKey(s) return type is Key instead of a PublicKey since it seems more obvious that a Key can be cast to a PrivateKey.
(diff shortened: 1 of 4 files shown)


twiss (approved):

edmorley:
@larabr @twiss  Hi! :-)

After this change I now get:

```
node_modules/openpgp/openpgp.d.ts:10:85 - error TS2307: Cannot find module '@openpgp/web-stream-tools' or its corresponding type declarations.

10 import type { WebStream as GenericWebStream, NodeStream as GenericNodeStream } from '@openpgp/web-stream-tools';
                                                                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~

Found 1 error in node_modules/openpgp/openpgp.d.ts:10
```

Should `@openpgp/web-stream-tools` have been added to `dependencies` not `devDependencies` in this repo?

Alternatively, could you document the required optional types in the README?

larabr:
Hey @edmorley , thanks for spotting this. I wouldn't add it to the `dependencies` since it's only required by TS, which we do not officially support .

Adding the info on the readme is a good idea, feel free to open a PR :)

KnisterPeter:
@larabr I find it really strange that you do not support typescript 'officially' but do provide typings with this package. And then refuse to fix issues with that.

larabr:
@KnisterPeter We provide typings because we think it adds value, and it's better than the alternative, namely relying on https://github.com/DefinitelyTyped/DefinitelyTyped, which was done in previous versions, and was problematic since we weren't directly checking them for correctness. This specific issue is also easily fixed on the app's side.
That said, we're planning to add full TS support in future versions of openpgpjs.

KnisterPeter:
@larabr Thanks for your answer.

> We provide typings because we think it adds value
Currently it's broken and it's not easy to see how this could be fixed (besides searching all issues). It breaks the npm dependency chain.
Anyway I will for sure accept it.

> That said, we're planning to add full TS support in future versions of openpgpjs
Good to hear :smiley:

larabr:
I agree searching issues is not ideal, which is why we would welcome a PR that adds info about how to integrate Typescript in the README :)
We just have a lot on our hands at the moment, and contributing to smaller issues like this really makes a difference.

KnisterPeter:
Here we go: https://github.com/openpgpjs/openpgpjs/pull/1586

PR #1518, opened 2022-05-09
Fix loading browser built in JSDom environment

larabr:
JSDom simulates a browser environment in node, so we shouldn't rely on runtime detection of Node .

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -425,12 +425,12 @@ const util = {
   },
 
   getHardwareConcurrency: function() {
-    if (util.detectNode()) {
-      const os = require('os');
-      return os.cpus().length;
+    if (typeof navigator !== 'undefined') {
+      return navigator.hardwareConcurrency || 1;
     }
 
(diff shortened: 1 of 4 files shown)


src/util.js
```diff
@@ -425,12 +417,12 @@ const util = {
...
   getHardwareConcurrency: function() {
-    if (util.detectNode()) {
-      const os = require('os');
-      return os.cpus().length;
+    if (typeof navigator !== 'undefined') {
+      return navigator.hardwareConcurrency || 1;
     }
 
-    return (typeof navigator !== 'undefined' && navigator.hardwareConcurrency) || 1;
+    const os = require('os'); // if node
```
twiss, line 424: If we're gonna write a comment let's expand it a bit, e.g.
```suggestion
    const os = require('os'); // Assume we're on Node.js.
```

twiss (commented):
Looks good to me, one nitpick:

twiss (approved):

PR #1519, opened 2022-05-13
types fix

rrrooommmaaa:
allow constructing keys from packet list of `AllowedKeyPackets` (includes Signature, UserId etc.), the reverse of `toPacketList()`


larabr:
Thanks! These constructors should actually take `AnyPacket`, since those that are not `AllowedKeyPackets` are accepted and simply ignored -- they do not cause runtime errors.
That said, a PrivateKey constructor errors [if it encounters Public(Sub)Key packets](https://github.com/openpgpjs/openpgpjs/blob/main/src/key/private_key.js#L21), and similarly a [PublicKey will reject Private(Sub)Key packets](https://github.com/openpgpjs/openpgpjs/blob/main/src/key/private_key.js#L21).

We could improve the types in that sense, if needed, e.g. using `Exclude`.

larabr:
Thanks for the fix! I've included it in #1522 , since it required a related type update.

PR #1549, opened 2022-07-29
CI: Update browser targets

larabr:
TODOs:
- [x] only run crypto API tests where available
- [ ] ~add test to ensure that the crypto API is available in [certain] browsers (not mobile, since it's device dependant)~
- [x] cache built files to share across jobs
- [x] test latest browser versions directly in Github Actions, avoiding browserstack
- [x] switch from checkout@v2 to v3
- [x] ~is turnstyle still needed? github should have implemented support~


test/crypto/eax.js
```diff
@@ -142,6 +143,11 @@ module.exports = () => describe('Symmetric AES-EAX', function() {
   };
 
   describe('Symmetric AES-EAX (native)', function() {
+    before(function () {
+      const detectNative = () => !!(util.getWebCrypto() || util.getNodeCrypto());
+      detectNative() || this.skip();
```
twiss, line 148: I think this would be simpler to read as just
```suggestion
      if (!util.getWebCrypto() && !util.getNodeCrypto()) {
        this.skip();
      }
```

test/crypto/rsa.js
```diff
@@ -105,6 +107,8 @@ module.exports = () => describe('basic RSA cryptography', function () {
   });
 
   it('compare native crypto and bnSign', async function() {
+    detectNative() || this.skip();
+
```
twiss, line 111: And similarly here, I would follow the pattern from the previous test of
```suggestion
    if (!detectNative()) {
      this.skip();
    }
```
(and in the next one too)

test/karma.conf.js
```diff
@@ -85,26 +95,12 @@ module.exports = function(config) {
...
-        browser_version: '88.0',
-        os: 'OS X',
-        os_version: 'Big Sur'
-      },
-      bs_safari_14: {
+      bs_safari_15: { // Webkit and Safari can differ in behavior
         base: 'BrowserStack',
         browser: 'Safari',
-        browser_version: '14.0',
+        browser_version: 'latest',
```
twiss, line 101: This should say `15.0`, no? Or alternatively the name of the browser should be `bs_safari_latest`

PR #1550, opened 2022-08-03
CI: run worker tests first to have enough time to download the required scripts

larabr:
This should fix issues with Safari 13 not managing to load the worker.


twiss (approved):
Interesting. Alrighty :+1:

larabr:
> Interesting. Alrighty 👍

I am not 100% of what the problem is, but without this change it never works anymore, so it's an improvement 😅

PR #1567, opened 2022-09-29
Plugin interface and Nitrokey Webcrypt support

szszszsz:
Hi!

This is a working draft implementation of the plugin system loosely described at https://github.com/Nitrokey/nitrokey-webcrypt-openpgpjs/issues/1 . Please take a look at the below description of the details, as well as the usage example.
Instead of a global plugin object I've decided to pass it through calls, to minimize potential state-related issues.

Requesting for the review and pointers.

Notes:
- Right now only P256 is supported, but this should be easy to extend to other ECCs, and with a slight modifications to RSA as well.
- Similarly to the OpenPGP cards, only one identity is supported, but this should be easy to extend to multiple identities using fingerprints, and over multiple devices utilizing private key data stubs.
- Implemented Nitrokey WebCrypt tests are passing. All basic operations are working.
- Regular OpenPGPjs tests are passing. [OpenPGPJS Unit Tests.pdf](https://github.com/openpgpjs/openpgpjs/files/9674664/OpenPGPJS.Unit.Tests.pdf)
- Dependency on the Nitrokey Webcrypt is added to keep everything in place, and for easier testing. This can be removed before the final merge, along with the tests, and kept separately. The gist of the proposed change is in the first commit.
- Similarly, the usage example placed in Readme can be moved somewhere else, or removed completely. However it seems to illustrate the idea quite well.

To do:
- [x] This PR is based on OpenPGPjs 5.3.1. It probably needs to be rebased.
- [x] Revert serializing the key generation calls (this was needed to make sure that the main key and subkey will not switch places during import).
- [x] Update Typescript definitions for the main calls, and plugin structure.
- [x] Plugin structure Javascript documentation for the generated docs (to be confirmed).
- [ ] Make callbacks more generic (currently taking parameters of ECC curves).
- [ ] Make Nitrokey Webcrypt tests optional (if decided for them to stay).

Other ideas:
- Merge `plugin` object into the global/local config structure. To discuss, whether this should be mixed, and passed deep down to cryptographic primitives.

--------------

A plugin interface for the private keys' operations is available, which allows to increase the security by storing the
private key on another medium than browser or PC, thus providing much greater, if not complete protection from the 
secret material leaks. Additionally, this can increase usage control by introducing user confirmation 
(e.g. through a touch button) for the private key operations.

The plugin interface is realized by a plugin object containing callbacks to the defined private key operations: sign, decrypt,
agree, generateKeyPair. The detailed description of each is provided in the Readme example . The plugin object is then passed
to all related main OpenPGPjs operations. If that's omitted, the standard OpenPGPjs behavior will be executed.

In the following implementation the host of the private key operations will be a Nitrokey WebCrypt device 
(e.g. Nitrokey 3). Nitrokey WebCrypt uses Webauthn standard to create a communication channel with the security key
handling it, which by design should allow to run it on any modern browser and platform.
At the moment it is tested only in the Chrome-based browsers. Support for the Node.js requires additional implementation.
More details at:
- https://github.com/Nitrokey/nitrokey-webcrypt
- https://github.com/Nitrokey/nitrokey-webcrypt-js
- https://github.com/Nitrokey/nitrokey-webcrypt-js-bundle


src/crypto/public_key/elliptic/curves.js
```diff
@@ -169,8 +179,16 @@ class Curve {
...
+  async genKeyPair(plugin) {
     let keyPair;
+    if (this.name === 'webcrypt_p256'){
+      try {
+        return plugin.generateKeyPair(this);
+      } catch (err) {
+        util.printDebugError('Nitrokey WebCrypt failed generating ec key ' + err.message);
+        throw err;
+      }
+    }
```
twiss, line 191: Here we should just check whether the plugin is there.
In addition, this error message is way too specific, that can go inside the plugin function.
And finally, I would prefer not to pass the entire curve object here, if possible, and maybe only the algorithm name and curve name, if that's sufficient?
You can pass the algorithm name from higher up, and get the curve name from `this.name`.

src/crypto/public_key/elliptic/curves.js
```diff
@@ -205,13 +223,20 @@ class Curve {
...
+  const keyPair = await curve.genKeyPair(plugin);
+  let Q;
+  let secret;
+  if (curve.name === 'webcrypt_p256') { //web
+    Q = keyPair.publicKey;
+    secret = keyPair.privateKey;
+  } else {
+    Q = new BigInteger(keyPair.publicKey).toUint8Array();
+    secret = new BigInteger(keyPair.privateKey).toUint8Array('be', curve.payloadSize);
+  }
```
twiss, line 239: Is this change necessary? To be honest, I'm also not sure why the last two lines are needed, they seem like a no-op (unless the public key has leading zeros, or the private key has missing leading zeros). Maybe we can check where it's necessary and do it only there.

src/key/helper.js
```diff
@@ -4,32 +4,34 @@
...
   await secretSubkeyPacket.computeFingerprintAndKeyID();
   return secretSubkeyPacket;
 }
 
-export async function generateSecretKey(options, config) {
+export async function generateSecretKey(options, config, plugin = null) {
+  if (plugin !== null) {
+    plugin.type = 'main';
+    plugin.dataMain = { options, config };
+  }
```
twiss, line 30: I guess this is what you meant by

> Revert serializing the key generation calls (this was needed to make sure that the main key and subkey will not switch places during import).

? Indeed this is quite hacky and should be changed. If you pass the algorithm name, I think that should also be sufficient (instead of passing whether it's the primary key or a subkey), right?

docs/CleartextMessage.html
```diff
@@ -346,7 +346,7 @@ <h5>Parameters:</h5>
     
     <dt class="tag-source">Source:</dt>
     <dd class="tag-source"><ul class="dummy"><li>
-        <a href="https://github.com/openpgpjs/openpgpjs/blob/v5.3.1/src/cleartext.js">cleartext.js</a>, <a href="https://github.com/openpgpjs/openpgpjs/blob/v5.3.1/src/cleartext.js#L112">line 112</a>
+        <a href="https://github.com/openpgpjs/openpgpjs/blob/v5.3.1/src/cleartext.js">cleartext.js</a>, <a href="https://github.com/openpgpjs/openpgpjs/blob/v5.3.1/src/cleartext.js#L114">line 114</a>
```
twiss, line 349: Could you un-commit the changes to the `docs` folder? That will make the diff easier to read, and they will be updated automatically on release anyway 😊

twiss (commented):
Thanks for working on this!

Some high-level comments:

> Other ideas:
>
> - Merge `plugin` object into the global/local config structure. To discuss, whether this should be mixed, and passed deep down to cryptographic primitives.

Indeed I would move `plugin` to `config.plugin`, or maybe `config.hooks`. That way, we don't have to pass two objects everywhere.

Further, I would propose that we make the arguments of each plugin function / hook into a single object, with descriptive property names. That way, it can be extended in the future more easily, if necessary. For example, we could have:

```js
const hooks = {
  generateKey({ algorithmName, curveName, rsaBits }) {
    // ...
  }
};
```

Further, the current sketch of that function returns a private key as well (of all zeros). But, in this case we shouldn't store a private key at all, instead it should probably be a GNU-dummy key. We have some support for that, but not for generating it, I think that will need to be added.

And similarly, the other functions currently take a private key, I assume that should be removed and replaced with a fingerprint, to reference the key on the device?

And finally, some more specific comments below 😊

docs/CleartextMessage.html
szszszsz: Done!

twiss:
Thinking about this a bit more, I think what I wrote before is not quite sufficient.

The difficult part with this use case is that it's not just a hook that replaces existing functions, but the behavior for these is a bit different, since it's referring to keys on a device, instead of keys handled by OpenPGP.js. That means not just that we have to store the generated key as a dummy key, but also that when *selecting* keys to use, we should also select dummy keys, instead of ignoring them.

So, I think this plugin should also be named a bit more specifically, for example `config.hardwareKeys`, or so. Then, we can check if that's set, and if so, generate/use dummy keys.

twiss:
Though actually - it seems like hardware keys are not stored exactly the same as dummy keys. https://lists.gnupg.org/pipermail/gnupg-users/2015-February/052747.html has some documentation. Currently, we only implement `1001` - "Do not store the secret part at all". We probably also need to implement `1002` - "A stub to access smartcards", for this. We could store `gnu-divert-to-card` in the `s2k.type` (see [src/type/s2k.js](https://github.com/openpgpjs/openpgpjs/blob/main/src/type/s2k.js)), and then create an `secretKeyPacket.isStoredInHardware()` function or some such, analogous to `isDummy` for `gnu-dummy`. Then, we could generate and use such keys if `config.hardwareKeys` is defined (or add a special parameter for key generation).

src/crypto/public_key/elliptic/curves.js
larabr: It is needed for p521 (maybe also some brainpool curve, not sure)

src/crypto/public_key/elliptic/curves.js
```diff
@@ -62,6 +62,16 @@ const curves = {
...
+  webcrypt_p256: {
+    oid: [0x06, 0x08, 0x2A, 0x86, 0x48, 0xCE, 0x3D, 0x03, 0x01, 0x07],
+    keyType: enums.publicKey.ecdsa,
+    hash: enums.hash.sha256,
+    cipher: enums.symmetric.aes128,
+    node: nodeCurves.p256,
+    web: webCurves.p256,
+    payloadSize: 32,
+    sharedSize: 256
+  },
```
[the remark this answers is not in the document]
szszszsz: Done as suggested

src/crypto/public_key/elliptic/curves.js
szszszsz: Done as suggested

src/crypto/public_key/elliptic/curves.js
szszszsz: It was not working for me before, but now does. I can't pinpoint the cause right now - perhaps the plugin.generate implementation was missing proper data in its result? Anyway, reverted that change.

src/key/helper.js
szszszsz: That's correct. Instead of this temporary solution I am now passing a new object, which holds the algorithm type as suggested, and restored the concurrent key generation.

szszszsz:
Hey @twiss @larabr  !

Thank you for the review! I got myself busy with something else, but now I would like to finish this PR asap. I have addressed some of the concerns, and minimize the patch.

Further work:
- [ ] move plugin to config.hardwareKeys
- [ ] set S2K identifier to mark the dummy key on generation
- [ ] implement secretKeyPacket.isStoredInHardware() 
- [ ] finish JSdoc

Questions:
1. Regarding the config object - should I pass it as a whole to functions, that did not accept it before, or unpack it? I think the former sounds better for the future extensibility.
2. As for the `secretKeyPacket.isStoredInHardware() `, can you elaborate where do you plan to use it? I presume you would like to check that value instead of whether the plugin field in config is set.

The general idea as far as I see is to generate the keys with the plugin field set once, and then allow the future operations on them without constantly setting it on use (as done right now for decryption and signing), but instead having it all defined in the global config object and use `isStoredInHardware()` to determine that. Please correct me if I see that wrong.

I am open for a video call if that would be faster for you - please contact me over `szczepan@nitrokey.com` in that case.

szszszsz:
Please review the latest changes. Would that s2k behavior work for you?

larabr:
Hey @szszszsz , thanks for the changes, just an heads up that @twiss is on vacation until next week, so the review will have to wait a bit

szszszsz:
@larabr OK! Good to know, thank you!

src/packet/public_key_encrypted_session_key.js
```diff
@@ -125,6 +125,9 @@ class PublicKeyEncryptedSessionKeyPacket {
     if (this.publicKeyAlgorithm !== key.algorithm) {
       throw new Error('Decryption error');
     }
+    if (key.isStoredInHardware() && !plugin){
```
twiss, line 128: Can we pass `config`, rather than `plugin`, all the way here as well?

src/packet/secret_key.js
```diff
@@ -203,7 +203,7 @@ class SecretKeyPacket extends PublicKeyPacket {
     arr.push(new Uint8Array(optionalFieldsArr));
 
     if (!this.isDummy()) {
-      if (!this.s2kUsage) {
+      if (!this.s2kUsage || this.isStoredInHardware()) {
         this.keyMaterial = crypto.serializeParams(this.algorithm, this.privateParams);
       }
```
twiss, line 208: I don't understand why this is needed - if the private key is stored in hardware, the private key material is not here, right? We should simply serialize the S2K specifier (which is already done earlier in the function), and nothing else, I think?

src/packet/secret_key.js
```diff
@@ -212,7 +212,7 @@ class SecretKeyPacket extends PublicKeyPacket {
...
       arr.push(this.keyMaterial);
 
-      if (!this.s2kUsage) {
+      if (!this.s2kUsage || this.isStoredInHardware()) {
         arr.push(util.writeChecksum(this.keyMaterial));
       }
```
twiss, line 217: (same here)

twiss (commented):
Hey :wave: Sorry for the delay. This looks like a large step in the right direction, yeah :+1: Some minor comments below.

szszszsz:
Hey! Thank you looking into it! I will get back to this project next week.
Briefly regarding the comments, I think the deserialization was not working properly yet for me during the tests, hence the `isStoredInHardware` but I wanted to confirm that the current development is going into the expected direction.
I will look how to make it more generic in the next step.

szszszsz:
Hey @twiss ! 
Finally I got back to this PR. Here comes a new set of changes:
- Refactored plugin object to be passed through the config object, instead of separately
- Working stub key serialization (tested by importing it to the GnuPG), with serial number storing in the IV field (done as in GnuPG) - see a6241471f35229c56912e5c91f5fc905c28fbbd9 message for details
- Extended test to forbid some private key material operations, like encrypting it
- Plugin object has now an abstract class it can be extended from, for easier development and extension
- Removed obsolete code
- Updated JSDocs

Kindly asking for a review. Do you see any other work needed before merging?
All tests are passing, both in browser and Node (run from CLI).
I've left changes divided to multiple commits for easier reading.

Left tasks:
- [ ] extract / cleanup Nitrokey WebCrypt tests 
     - please let me know if you want to keep them here, or shall I move it to another repo
- [ ] regenerate docs

src/packet/public_key_encrypted_session_key.js
szszszsz: Done! Config should be now passed where possible. In a couple of places I've left `plugin_with_data` object to keep state for the concurrent keys generation.

src/packet/secret_key.js
szszszsz: Indeed serializing parameters was not needed here. I've corrected the condition here.

src/packet/secret_key.js
szszszsz: Confirmed. Corrected.

szszszsz:
@twiss Hey hey! Friendly ping.

twiss (commented):
Hey :wave: Sorry for the delay! I started writing a response but didn't post it and forgot about it :upside_down_face:

Overall, I think the structure is going in the right direction. However, I think the API still needs a bit more thought and care, as this will essentially become part of our public API, and can't actually be changed that easily later (as we'd also have to change the WebCrypt OpenPGP.js plugin which will live in a separate repo).

I've left some more concrete comments below. But, please also spend some time yourself thinking about whether this API will keep working for you in the future (other algorithms, other devices, etc).

Also, on a much more nitpicky note, please use camel case everywhere, and generally please try to match the style of the surrounding code :blush:

twiss:
Hello :wave: Just a heads up, #1597 will cause some merge conflicts with this PR, since it refactors the S2K code. So we'll need to rebase this PR, if it's merged after that one. I don't necessarily want to burden you with that, we can rebase this PR; but if you did some work on it in the past three weeks, it might be good to push it here first, so that we can (either merge it if it's finished - but no rush, ofc - and then rebase #1597, or) rebase the code in this PR after merging that one :blush:

szszszsz:
Hey, thank you for the tip! About the ready code, I've followed some style code advises, and will update this PR's branch in a moment.
From the brief look it should not be that much of conflicting code - I will check later if I could rebase on that PR.
More updates are incoming in the following days, but I need to clear out some doubts first.

src/key/helper.js
```diff
@@ -4,32 +4,38 @@
...
  */
 
-import {
-  PublicKeyPacket,
-  PublicSubkeyPacket,
-  SecretKeyPacket,
-  SecretSubkeyPacket,
-  SignaturePacket
-} from '../packet';
+import { PublicKeyPacket, PublicSubkeyPacket, SecretKeyPacket, SecretSubkeyPacket, SignaturePacket } from '../packet';
```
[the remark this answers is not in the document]
szszszsz: Yes, definitely this should be provided separately. Reverted this change and other similar.

src/hardwareKeys.js
```diff
@@ -0,0 +1,103 @@
...
+    throw new Error('Method must be implemented.');
+  }
+
+  /**
+   * Implement here all the required steps to initialize the plugin. This method should be called first
+   * before the actual use.
+   * @returns {Promise<void>}
+   * @async
+   */
+  async init() {
```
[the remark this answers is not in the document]
szszszsz: Indeed, this is rather an internal implementation detail, which could be lazily evaluated. Removed.

src/hardwareKeys.js
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @param {Object} obj - An object argument for destructuring
+   * @param {Curve} obj.curve - Elliptic curve object
+   * @param {Uint8Array} obj.V - Public part of ephemeral key
+   * @param {Uint8Array} obj.Q - Recipient public key
+   * @param {Uint8Array} obj.d - Recipient private key
+   * @returns {Promise<{secretKey: Uint8Array, sharedKey: Uint8Array}>}
+   * @async
+   */
+  async agree({ curve, V, Q, d }) {
```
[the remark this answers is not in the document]
szszszsz: 1. Makes sense. I did not want to introduce too much revolutionary changes, hence kept the same names and API, but I like the direction for the better naming. Will keep that in mind.
2. Noted about the the `curveName` preference. In the first iteration I was focused on the curves, hence this object leak. Certainly this would require something else for the RSA.

Renamed as requested, and removed the obsolete secret key field from the return type.

src/packet/secret_key.js
```diff
@@ -393,8 +467,15 @@ class SecretKeyPacket extends PublicKeyPacket {
...
-    const { privateParams, publicParams } = await crypto.generateParams(this.algorithm, bits, curve);
+  /**
+   * @param {{hardwareKeys: HardwareKeys, algo: number}} [hardwareKeys_with_data]
+   */
+  async generate(bits, curve, hardwareKeys_with_data) {
+    const { privateParams, publicParams } = await crypto.generateParams(this.algorithm, bits, curve, hardwareKeys_with_data);
+    if (hardwareKeys_with_data) {
+      const serialNumber = await hardwareKeys_with_data.hardwareKeys.serial_number();
+      this.makeStub(serialNumber);
+    }
```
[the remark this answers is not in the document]
szszszsz: Left to do:
- [ ]  `hardwareKeys_with_data` -> `config`

src/hardwareKeys.js
```diff
@@ -0,0 +1,103 @@
...
+   * @param {Uint8Array} obj.Q - Recipient public key
+   * @param {Uint8Array} obj.d - Recipient private key
+   * @param {Uint8Array} obj.hashed - The hashed message
+   * @returns {Promise<{
+   *   r: Uint8Array,
+   *   s: Uint8Array
+   * }>} Generated signature, 32 bytes in each field
+   * @async
+   */
+  async sign({ oid, hashAlgo, data, Q, d, hashed }) {
```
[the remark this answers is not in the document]
szszszsz: Good points - will do. I am not testing this with the RSA yet, but planning so in the future.

src/hardwareKeys.js
```diff
@@ -0,0 +1,103 @@
...
+   *   privateKey: Uint8Array
+   * }>} Generated key material
+   * @async
+   */
+  async generate({ algorithmName, curveName, rsaBits }) {
+    throw new Error('Method must be implemented.');
+  }
+}
+
+export default { HardwareKeys };
```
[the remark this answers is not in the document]
szszszsz: Good questions. The intention indeed was to share to non-Typescript users an abstract class, which implementation would be then overloaded. Do we consider such use case? I am not sure how much backwards compatibility is needed here.
Or does Typescript generate the abstract class from the provided interface to use in JS? If so, this stub is indeed not needed.

src/key/helper.js
```diff
@@ -4,32 +4,38 @@
...
 import util from '../util';
 import defaultConfig from '../config';
 
 export async function generateSecretSubkey(options, config) {
+  let hardwareKeys_with_data = null;
+  if (config && config.hardwareKeys) {
+    hardwareKeys_with_data = {
+      hardwareKeys: config.hardwareKeys
+    };
+  }
```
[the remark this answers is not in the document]
szszszsz: Hmm, I will check how can I refactor that in the source code. 
As you've noticed, this is indeed a simple object wrap-up, so additional arguments could be attached later for invoking in the async key generation.
I remember I wanted to minimize API changes in the initial implementation, specifically arguments lists, and perhaps that is one of the effects.

src/packet/secret_key.js
```diff
@@ -73,6 +73,13 @@ class SecretKeyPacket extends PublicKeyPacket {
      * @type {Object}
      */
     this.privateParams = null;
+
+    /**
+     * The IV used for S2K operations.
+     * Dubbed as a 16B serial number container for the stub keys for external access (as in GnuPG C implementation).
+     * @type {Uint8Array}
+     */
+    this.iv = null;
```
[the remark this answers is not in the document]
szszszsz: I have followed here the GnuPG implementation, where the IV was handled like that to make it familiar for people working with both codebases. I think it's rather a not pretty C hack in general, and would rather change it to something more readable as well. 
For the S2K extension I will follow the new PR lead here.

openpgp.d.ts
```diff
@@ -337,13 +338,20 @@ interface Config {
   rejectMessageHashAlgorithms: Set<enums.hash>;
   rejectPublicKeyAlgorithms: Set<enums.publicKey>;
   rejectCurves: Set<enums.curve>;
+  hardwareKeys?: HardwareKeys;
```
larabr, line 341:
```suggestion
  hardwareKeys: HardwareKeys | null;
```
All fields are set in `Config` -- `PartialConfig` (just below) declares everything optional

src/crypto/crypto.js
```diff
@@ -276,17 +281,17 @@ export function generateParams(algo, bits, oid) {
       }));
     }
     case enums.publicKey.ecdsa:
-      return publicKey.elliptic.generate(oid).then(({ oid, Q, secret }) => ({
+      return publicKey.elliptic.generate(oid, hardwareKeys_with_data).then(({ oid, Q, secret }) => ({
```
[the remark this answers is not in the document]
larabr: I'd prefer passing around `enums` whenever possible (following up to #1410)

src/hardwareKeys.js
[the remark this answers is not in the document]
larabr: I think this should be in a separate repo, provided as an helper if it's something the people need. TS is the standard way to do type-checking, and we're moving towards full support for that, so I don't think this repo is the right place.

openpgp.d.ts
szszszsz: Hey @larabr!
That makes sense indeed - applied as proposed.

openpgp.d.ts
```diff
@@ -337,13 +338,20 @@ interface Config {
...
 }
 export var config: Config;
 
 // PartialConfig has the same properties as Config, but declared as optional.
 // This interface is relevant for top-level functions, which accept a subset of configuration options
 interface PartialConfig extends Partial<Config> {}
 
+declare abstract class HardwareKeys {
```
larabr, line 349: I think this should be an interface, as it's more generic. We are using abstract classes if we are extending them as part of the definitions or if we need other OOP features (e.g. `instanceof HardwareKeys`), which doesn't seem to be the case?

openpgp.d.ts
```diff
@@ -337,13 +338,20 @@ interface Config {
...
   rejectCurves: Set<enums.curve>;
+  hardwareKeys: HardwareKeys | null;
 }
 export var config: Config;
 
 // PartialConfig has the same properties as Config, but declared as optional.
 // This interface is relevant for top-level functions, which accept a subset of configuration options
 interface PartialConfig extends Partial<Config> {}
 
+declare abstract class HardwareKeys {
```
larabr, line 349: As Daniel [mentioned](https://github.com/openpgpjs/openpgpjs/pull/1567#discussion_r1084389921), I think this should be an interface, as it's more generic. We are using abstract classes if we are extending them as part of the definitions or if we need other OOP features (e.g. `instanceof HardwareKeys`), which doesn't seem to be the case.

szszszsz: I agree. There is no blanket implementation either. I remember I wanted to model it, so that the JavaScript's abstract class would have the same interface to keep it somewhat compatible between languages. I see the point in having this Typescript only, as a proper type checker.
I will change that abstract class to interface as proposed.

src/crypto/signature.js
```diff
@@ -117,13 +117,21 @@ export async function verify(algo, hashAlgo, signature, publicParams, data, hash
  * @param {Object} privateKeyParams - Algorithm-specific public and private key parameters
  * @param {Uint8Array} data - Data to be signed
  * @param {Uint8Array} hashed - The hashed data
+ * @param {Object} [config] - Custom configuration settings to overwrite those in [config]{@link module:config}
  * @returns {Promise<Object>} Signature                      Object containing named signature parameters.
  * @async
  */
-export async function sign(algo, hashAlgo, publicKeyParams, privateKeyParams, data, hashed) {
+export async function sign(algo, hashAlgo, publicKeyParams, privateKeyParams, data, hashed, config = null) {
```
larabr, line 124:
```suggestion
export async function sign(algo, hashAlgo, publicKeyParams, privateKeyParams, data, hashed, config) {
```
Is there a reason to have it `null` ? I saw the same in other places. 
The config should never be undefined, if needed, it's assumed to be defined, possibly set to `defaultConfig` by an higher level function.
This also means you can then remove the `config &&`  checks everywhere 🙂

szszszsz: The idea was to keep the API backwards compatibility, and I initially missed the `defaultConfig` way of specifying the value. That clears implementation a bit, good spot!

szszszsz:
Regarding the CI error:
> npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

Would you like me to merge in the current `package.lock` from the main branch?

src/crypto/signature.js
szszszsz: This should be corrected now - resolving.

openpgp.d.ts
szszszsz: Changed as suggested to interface. Tests are passing:
- browser
- node (without Webcrypt)
- `npm run test-type-definitions`

twiss:
I think the issue might actually be that since you added `nitrokey_webcrypt` to the `dependencies` in package.json, that you have to update package-lock.json as well.

(Although, as you already said, it should probably be removed before merging, and then the package-lock.json should be reverted as well, but you can update it for now for testing indeed.)

szszszsz:
I've just run `npm install` to update, and `npm ci` for the local tests. This should go in the CI now.

larabr:
Thanks for working on this @szszszsz, it'll open up interesting possibilities! Sorry to chime in only now, I would like to have your opinion on a couple of points.

First off, with the current implementation, if `config.hardwareKeys` is set, but I am generating e.g. a RSA key (or EdDSA), the hardware won't be used, and I won't get any indication of that. I would expect the operation to throw telling me that the hardware key does not support RSA keys, "forcing" me to then pass `config: { hardwareKeys: null }` (assuming `openpgp.config.hardwareKeys` is set globally). Otherwise, I might get a false sense of security, and/or think the key was stored on hardware when it wasn't, and lose the key data as a result.

Secondly, I was wondering if a higher-level interface would work for you? I am thinking of having the lib call out to `config.hardwareKeys` already in e.g. [`publicKeyDecrypt`](https://github.com/openpgpjs/openpgpjs/blob/5beff0c38e3a59a5461c14d6152883e2ae7a3b79/src/crypto/crypto.js#L87-L112), for each algorithm (to address the point above). Similarly to what you're doing [here](https://github.com/openpgpjs/openpgpjs/blob/5beff0c38e3a59a5461c14d6152883e2ae7a3b79/src/crypto/signature.js#L132), but without assuming it's an ECDH key. 
The advantage I see is that (1) we'd have a centralised spot where the requests are "proxied", making it easy to follow the flow of execution in all cases,  and (2) the `hardwareKey` integration would be decoupled from our low-level crypto modules. In fact, I am a bit uncomfortable with [this type of change](https://github.com/openpgpjs/openpgpjs/blob/5beff0c38e3a59a5461c14d6152883e2ae7a3b79/src/crypto/public_key/elliptic/curves.js#L224), where parameters are "nullified". Instead, I think that if [`generateParams`](https://github.com/openpgpjs/openpgpjs/blob/5beff0c38e3a59a5461c14d6152883e2ae7a3b79/src/crypto/crypto.js#L270-L308) were to redirect directly to `config.hardwareKeys`, then the implemented interface could simply avoid returning `privateKeys`, and it wouldn't risk disrupting the existing logic (where the single params are always assumed to be defined/non-null).
The downside is that the HardwareKey interface will need to take care of e.g. the [ECDH KDF derivation](https://github.com/openpgpjs/openpgpjs/blob/5beff0c38e3a59a5461c14d6152883e2ae7a3b79/src/crypto/public_key/elliptic/ecdh.js#L199-L211), but that relies on standard constructs, and we can help you with that part of the integration if needed. I also have the impression -- but this is totally speculative -- that some HardwareKeys might already carry out the KDF/padding steps if they were developed with built-in OpenPGP support.

PR #1571, opened 2022-10-20
Add RawNotations Type to d.ts File (openpgpjs#1569)

LeoKotschenreuther:
This commits adds the `rawNotations` property and the appropriate interface definition to the SignaturePacket class in the d.ts file.


larabr (approved):

larabr:
Thanks for the fix!

PR #1576, opened 2022-11-03
TypeScript: fix signature of armor function

LeoKotschenreuther:
In the [JSDoc](https://github.com/openpgpjs/openpgpjs/blob/main/src/encoding/armor.js#L357-L358) the partIndex and partTotal params of the `armor` function are marked as optional. Also when reading the function it appears that it will still work if they are not passed. This commit updates the openpgp.d.ts file to match what I observed in the actual implementaion.


twiss (approved):
:+1: Thanks!

PR #1584, opened 2022-12-29
Add revoke method to User

toberndo:
This allows to create revocation certificates for a user, following the same method signature as `Subkey.revoke()`.


twiss (approved):
:+1: Thanks, and apologies for the long delay! I was on vacation and missed this PR.

PR #1588, opened 2023-01-30
Adding missing functions in SubKey class type definition

FabienForestier:
Typings of the SubKey class is still incomplete but I added a few of the missing function type definitions.

### DIFF
--- a/openpgp.d.ts
+++ b/openpgp.d.ts
@@ -98,6 +98,10 @@ export class Subkey {
   public getCreationTime(): Date;
   public getAlgorithmInfo(): AlgorithmInfo;
   public getKeyID(): KeyID;
+  public getExpirationTime(date?: Date, config?: Config): Promise<Date | typeof Infinity | null>
+  public clone(): Subkey;
+  public isRevoked(signature: SignaturePacket, key: PublicSubkeyPacket | SecretSubkeyPacket | PublicKeyPacket | SecretKeyPacket, date?: Date, config?: Config): Promise<boolean>;
+  public update(subKey: Subkey, date?: Date, config?: Config): Promise<void>
 }
 
 export interface User {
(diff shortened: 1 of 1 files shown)


openpgp.d.ts
```diff
@@ -98,6 +98,10 @@ export class Subkey {
   public getCreationTime(): Date;
   public getAlgorithmInfo(): AlgorithmInfo;
   public getKeyID(): KeyID;
+  public getExpirationTime(date?: Date, config?: Config): Promise<Date | typeof Infinity | null>
+  public clone(): Subkey;
```
larabr, line 102:
```suggestion
```
Not sure about exposing this one. Do you have a use case for it?

FabienForestier: Actually I don't, I'm used to work with typescript and assumed that the method would have been marked private if it wasn't used outside the class. But this is only available in typescript file and not in javascript my bad. I added a commit to remove this line.

openpgp.d.ts
```diff
@@ -98,6 +98,9 @@ export class Subkey {
   public getCreationTime(): Date;
   public getAlgorithmInfo(): AlgorithmInfo;
   public getKeyID(): KeyID;
+  public getExpirationTime(date?: Date, config?: Config): Promise<Date | typeof Infinity | null>
+  public isRevoked(signature: SignaturePacket, key: PublicSubkeyPacket | SecretSubkeyPacket | PublicKeyPacket | SecretKeyPacket, date?: Date, config?: Config): Promise<boolean>;
```
larabr, line 102:
```suggestion
  public isRevoked(signature: SignaturePacket, key: AnyKeyPacket, date?: Date, config?: Config): Promise<boolean>;
```

FabienForestier:
I added a commit with your suggested changes ;)

larabr (approved):
Thanks again!

PR #1594, opened 2023-02-09
TypeScript: Added selfCertification property to PrimaryUser interface definition

rrrooommmaaa:
It's useful to be able to extract additional data from the signature, e.g. key usage flags

Also changed type definition for `Key.verifyAllUsers`


twiss:
Hey :wave: I'm not entirely opposed to adding this, but just FWIW, in v6 keys, this information will be moved to a direct-key binding signature, and the User ID self-certification won't necessarily be there. So I would advise against looking up information directly in the signature, if you can avoid it. Out of curiosity, what do you need the information for?

twiss (approved):
I'll merge it anyway since we already have `User.selfCertifications` in the types too. I guess we can make it optional once we merge v6 keys support :)

rrrooommmaaa:
> Out of curiosity, what do you need the information for?

We display primary key usage flags (certifyKeys, signData, encryptCommunication, encryptStorage) on a diagnostics page in our app,
https://github.com/FlowCrypt/flowcrypt-browser/blob/da4363a6eb51121ae585fc36f2273ce209aae04e/extension/js/common/core/crypto/pgp/openpgp-key.ts#L661

It'd be great to be able to extract this info by simpler means (as well as for each of the subkeys).

PR #1596, opened 2023-02-13
Docs: mark global `generateSessionKeys`'s `encryptionKeys` as optional

LucienLeMagicien:
`generateSessionKeys` calls `Message.generateSessionKeys`, which calls `helper.js`'s `getPreferredAlgo`, which returns `config.preferredSymmetricAlgorithm` when the `encryptionKeys` array is empty.
`encryptionKeys` also defaults to an empty array, and `Message.generateSessionkeys`'s `encryptionKeys` is properly marked as optional, so this seems to just be an oversight from the doc.

I'm not familiar with the code or jsdoc or if there's a particular doc style, so feel free to edit!

In any case, it looks like it renders correctly: 
Screenshot of the updated doc for generateSessionkey

### DIFF
--- a/src/openpgp.js
+++ b/src/openpgp.js
@@ -505,9 +505,9 @@ export async function verify({ message, verificationKeys, expectSigned = false, 
 ///////////////////////////////////////////////
 
 /**
- * Generate a new session key object, taking the algorithm preferences of the passed public keys into account.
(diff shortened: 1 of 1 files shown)


twiss (approved):
:+1: Thanks!

PR #1597, opened 2023-02-14
`crypto-refresh`: add support for Argon2 S2K

### LINKED ISSUE #1565: Clarify usage of config.s2kIterationCountByte
- OpenPGP.js version: 5.5.0
- Affected platform (Browser): 
  - Brave 1.43.89 (Chromium 105.0.5195.102)
  - Firefox 104.0.1 (64-bit)

loading the following content saved to a file `test.html` into one of the browsers listed above (others not tested)
```
<script src="openpgp.js"></script>                                                                                    
<h3>test s2kIterationCountByte 65011712</h3>
<script lang="JavaScript">                                                                                            
(async () => {                                                                                                        
  const { privateKey, publicKey, revocationCertificate } = await openpgp.generateKey({                                
    type: 'ecc',
    curve: 'curve25519',
    config: { s2kIterationCountByte: 65011712 },                                                                                                       
    userIDs: [{ name: '', email: 'test@foo.bar' }],                                
    passphrase: 'iloveyou',                                                                 
    format: 'armored'          
  });

    console.log(privateKey);   // '-----BEGIN PGP PRIVATE KEY BLOCK ... '
    console.log(publicKey);    // '-----BEGIN PGP PUBLIC KEY BLOCK ... '
    console.log(revocationCertificate);  // '-----BEGIN PGP PUBLIC KEY BLOCK ... '
})();
</script>
```

and copying the generated privkey from the browsers console and feeding it into either `pgpdump` or `gpg --list-packets` yields a (protect) count of **1024** (coded count 0) instead an expected value of **65011712**.
If the value given is diminished by one to 6501171**1**, the result actually after _pgpduming_ the privkey reads 6501171**2** so i suppose there is a minor flaw in the calculation of the value for the coded count which could have quite some security implications (render attacks more feasible and efficient) if not double checked.

larabr:
Added support for encrypting and decrypting messages or keys that use Argon2 as S2K (#1442).
This feature is backwards compatible, no breaking changes.

Newly introduced config fields (relevant for encryption only):
- `config.s2kType` (defaulting to `enums.s2k.iterated`): s2k to use on password-based encryption as well as private key encryption;
- `config.s2kArgon2Params` (defaulting to "uniformly safe settings" from Argon RFC): parameters to use on encryption when `config.s2kType` is set to `enums.s2k.argon2`;

More technical details:
Argon2 hashes are computed in Wasm, using a module that is only loaded if needed. The Wasm module is automatically reloaded -- to force memory deallocation -- if more than 1GB of memory is used.

Also fixes #1565 .


src/type/s2k/generic.js
```diff
@@ -28,16 +28,16 @@
...
-import defaultConfig from '../config';
-import crypto from '../crypto';
-import enums from '../enums';
-import util from '../util';
+import defaultConfig from '../../config';
+import crypto from '../../crypto';
+import enums from '../../enums';
+import util from '../../util';
 
 class S2K {
```
twiss, line 36: Please rename this to `GenericS2K`; `S2K` should (if anything) be an abstract base class of `GenericS2K` and `Argon2S2K` (though I don't think it's extremely important to have it)

src/type/s2k/index.js
```diff
@@ -0,0 +1,35 @@
+import defaultConfig from '../../config';
+import Argon2S2k, { Argon2OutOfMemoryError } from './argon2';
+import OtherS2K from './generic';
```
twiss, line 3: ... and here as well (renaming a class on import makes it quite confusing)

src/type/s2k/index.js
```diff
@@ -0,0 +1,35 @@
...
+ * @oaram {Object} [config]
+ * @returns {Object} New s2k object
+ * @throws {Error} for unknown or unsupported types
+ */
+export function newS2KFromType(maybeType, config = defaultConfig) {
+  if (!maybeType && !allowedDefaultS2KTypes.has(config.s2kType)) {
+    throw new Error('The provided `config.s2kType` value is not allowed');
+  }
+
+  const type = maybeType || config.s2kType;
```
twiss, line 20: IMO, this API is a bit strange in the `newS2KFromType(undefined, config)` case. I would prefer to split those cases up, i.e. you could have `newS2KFromConfig(config)`, which then calls `newS2KFromType(type)`.

src/type/s2k/index.js
```diff
@@ -0,0 +1,35 @@
+import defaultConfig from '../../config';
+import Argon2S2k, { Argon2OutOfMemoryError } from './argon2';
+import OtherS2K from './generic';
+import enums from '../../enums';
+
+const allowedDefaultS2KTypes = new Set([enums.s2k.argon2, enums.s2k.iterated]);
```
twiss, line 6: It's a bit nitpicky but this name is a bit strange to me - it's not about default S2K, but just the S2K types that are allowed to be generated. And, maybe instead of creating a set just for these two values, you could just have `if (config.s2kType !== enums.s2k.argon2 && config.s2kType !== enums.s2k.iterated)` in `newS2KFromConfig`?

src/type/s2k/argon2.js
```diff
@@ -0,0 +1,137 @@
...
+    this.t = passes;
+    /** @type {Integer} degree of parallelism (lanes) */
+    this.p = parallelism;
+    /** @type {Integer} exponent indicating memory size */
+    this.encodedM = memoryExponent;
+  }
+
+  generateSalt() {
+    this.salt = crypto.random.getRandomBytes(ARGON2_SALT_SIZE);
+  }
```
twiss, line 48: I know the existing S2K class did something similar, but rather than setting the `t`, `p` and `encodedM` parameters in the constructor, and then optionally overwriting them in `read()`, I would prefer to make it explicit when we actually use those. You could have a `generateParams(config)` function instead of a `generateSalt()`, and set both the salt and the other parameters there.
(And similarly in the `GenericS2K` class.)

src/type/s2k/argon2.js
```diff
@@ -0,0 +1,137 @@
...
+  * hashAlgorithm
+  * @param {String} passphrase - Passphrase containing user input
+  * @returns {Promise<Uint8Array>} Produced key with a length corresponding to `keySize`
+  * @throws {Argon2OutOfMemoryError|Errors}
+  * @async
+  */
+  async produceKey(passphrase, keySize) {
+    const decodedM = 2 << (this.encodedM - 1);
+    if (decodedM < 8 * this.p || this.encodedM > 31) {
+      throw new Error('Memory size and parallelism settings are incompatible.');
```
twiss, line 93: In the `this.encodedM > 31` case, this error message would be a bit confusing, no? I can see that happening if someone accidentally passes the number of KiB to `memoryExponent`, for example. So I think we should have a separate error message for that.

twiss:
Also, just to keep track, this will cause conflicts with #1567, we'll need to rebase one or the other, whichever is merged last.

src/type/s2k/argon2.js
larabr: I should remove this error, it's a leftover, it's already checked by the argon2 lib

src/type/s2k/argon2.js
larabr: I am not sure about this, with packets we also set default values in constructor, even tho they can be overwritten when reading.
API-wise, since the s2k instance is not really meant to be reused, having a function that re-generates potentially different configuration could be confusing. The salt is fundamentally different from the rest of the params, since it's meant to change for each S2K instance. Also, especially if we add `newS2KFromConfig`, selectively applying the config seems unexpected.

src/type/s2k/index.js
larabr: Yeah, I just didn't settle on the name 😅

src/type/s2k/index.js
larabr: Yeah I will rename, but I'd leave the constant since it's a pattern we have in other places (e.g. `allowedUnhashedSubpackets` and `allowedPackets`) and I think it highlights the constraint without having to look much into the code ("when is `newPacketFromConfig` used" etc).

twiss (approved):

PR #1599, opened 2023-02-16
Add support for creating critical signature subpackets

twiss:
Assign most signature subpacket types a criticality based on whether failing to interpret their meaning would negatively impact security.

For Notation Data subpackets, let the user indicate their criticality using the `signatureNotations[*].critical` property.

### DIFF
--- a/src/packet/signature.js
+++ b/src/packet/signature.js
@@ -226,41 +226,41 @@ class SignaturePacket {
     if (this.created === null) {
       throw new Error('Missing signature creation time');
     }
-    arr.push(writeSubPacket(sub.signatureCreationTime, util.writeDate(this.created)));
+    arr.push(writeSubPacket(sub.signatureCreationTime, true, util.writeDate(this.created)));
     if (this.signatureExpirationTime !== null) {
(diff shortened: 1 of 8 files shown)


src/packet/signature.js
```diff
@@ -753,18 +753,19 @@ class SignaturePacket {
...
  * @see {@link https://tools.ietf.org/html/rfc4880#section-5.2.3.2|RFC4880 5.2.3.2}
+ * @param {Boolean} critical - Whether the subpacket should be critical.
  * @param {Integer} type - Subpacket signature type.
  * @param {String} data - Data to be included
- * @returns {String} A string-representation of a sub signature packet.
+ * @returns {Uint8Array} The signature subpacket.
  * @private
  */
-function writeSubPacket(type, data) {
+function writeSubPacket(critical, type, data) {
```
larabr, line 765: I think that `critical` should be passed after the type, otherwise the function calls are confusing to read

twiss: Fair enough, changed :+1:

larabr (approved):

PR #1603, opened 2023-02-21
Add interoperability test suite to CI

marinthiercelin:


.github/workflows/sop-test-suite.yml
```diff
@@ -0,0 +1,89 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Download test results json artifact
+        id: download-test-results
+        uses: actions/download-artifact@v3
+        with:
+          name: test-suite-results.json
+      - name: Compare with baseline
+        uses: ProtonMail/openpgp-interop-test-analyzer@main
```
larabr, line 84: Idk if it's temporary that it's pointing to the main branch of ProtonMail/openpgp-interop-test-analyzer, but I think it'd be best to create and point to tags (e.g. v1) instead, so that if we need to change something on the main branch, we don't have to worry about it being usable immediately by the workflows

.github/workflows/sop-test-suite.yml
```diff
@@ -0,0 +1,89 @@
+name: SOP interoperability test suite
+
+on:
+  push:
+    branches: [ main ]
```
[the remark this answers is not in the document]
marinthiercelin: Good point, I removed the push trigger

.github/workflows/sop-test-suite.yml
marinthiercelin: Right, I added a v1 tag

larabr (approved):

marinthiercelin:
I don't have write access to the repo, so I can't merge it.

PR #1606, opened 2023-02-25
add config setting to allow future keys to be verified due to clock d…

Logicwax:
…rift issues

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -142,7 +142,14 @@ export default {
    * @property {Boolean} allowInsecureDecryptionWithSigningKeys
    */
   allowInsecureVerificationWithReformattedKeys: false,
-
+  /**
+   * Allow verification of keys whose self-signature creation time is invalid such as in the future.
+   * This setting is potentially insecure, but it is needed to deal with keys that were created in the future due to clock drift
+   * and have self-signature's creation date that exists in the future compared to the current time.
+   * @memberof module:config
+   * @property {Boolean} allowInsecureVerificationWithFutureSignatures
+   */
+  allowInsecureVerificationWithFutureSignatures: false,
   /**
    * Enable constant-time decryption of RSA- and ElGamal-encrypted session keys, to hinder Bleichenbacher-like attacks (https://link.springer.com/chapter/10.1007/BFb0055716).
    * This setting has measurable performance impact and it is only helpful in application scenarios where both of the following conditions apply:
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -68,9 +68,9 @@ export async function getLatestValidSignature(signatures, publicKey, signatureTy
   return latestValid;
 }
 
-export function isDataExpired(keyPacket, signature, date = new Date()) {
+export function isDataExpired(keyPacket, signature, date = new Date(), config) {
   const normDate = util.normalizeDate(date);
-  if (normDate !== null) {
+  if (normDate !== null && !config.allowInsecureVerificationWithFutureSignatures) {
     const expirationTime = getKeyExpirationTime(keyPacket, signature);
     return !(keyPacket.created <= normDate && normDate < expirationTime);
   }
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -382,7 +382,7 @@ class Key {
     // check for valid, unrevoked, unexpired self signature
     const { selfCertification } = await this.getPrimaryUser(date, userID, config);
     // check for expiration time in binding signatures
-    if (helper.isDataExpired(primaryKey, selfCertification, date)) {
+    if (helper.isDataExpired(primaryKey, selfCertification, date, config)) {
       throw new Error('Primary key is expired');
     }
     // check for expiration time in direct signatures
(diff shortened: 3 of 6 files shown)


larabr:
Hi @Logicwax , thanks for the contribution, but I don't think we should add a config option for this. The way to go around time skew is to pass a `date` param that is already in the future and accounts for the skew, or to pass `null` as date to skip date checks altogether. Could you check whether either option already covers your use case?

twiss:
I think there's some value in having an asymmetric option, i.e. that allows future keys but not expired ones (unlike `date: null`).
And, if you have multiple short-lived subkeys, one that is valid now, and one that is valid in 24h, you would still want to select the one that's valid now (unlike `date: someFutureDate`).
That being said, maybe what we really want is more like some grace period / skew window, e.g. a config setting to allow keys generated up to 24h in the future (but not more).
Then, you can just add that into the calculation, i.e. check something like `this.created <= normDate - config.gracePeriodForFutureDates` (which would default to 0). @Logicwax what do you think?

Logicwax:
> Hi @Logicwax , thanks for the contribution, but I don't think we need a config option for this. The way to go around time skew is to pass a `date` param that is already in the future and accounts for the skew, or to pass `null` as date to skip date checks altogether. Could you check whether either option already covers your use case?

OMG....i totally didn't realize that the normalize date function just passthroughs `null`.     I just tried passing `null` and now it works great for my use case.   Can't believe I missed that, thanks a lot!

It still would be nice to have a mode to ignore these things with encrypting messages, but for now I just do something like:
```
pgp.encrypt({
    message: await pgp.createMessage({text: message}),
    encryptionKeys: publicKey,
    date: (await publicKey.getEncryptionKey(undefined, null)).bindingSignatures[0].created
  })
  ```

**Comment by Logicwax** [comment]
> I think there's some value in having an asymmetric option, i.e. that allows future keys but not expired ones (unlike `date: null`). 

How would a key be a "future key" (creation time in the future), but be expired?

I think that's a great idea.    The reasons this has came up for me:
1) Code on a client that creates pgp keypairs and submits public key to the server, so that the server can respond back with some information encrypted to that public key.    Problem is, they might have clock drift slightly into the future.
2) PGP keypairs are generated on an airgap laptop, used to provision yubikeys and the public key moved over via SD card to hot connected machines.    Problem is, the operator forgot to set the clock properly so keys are generated in the future.   (of course keys generated WAY into the future would just be unacceptable, but I'm referring to cases of a few hours, or even just bad daylight savings time offsets).

currently, the `.encrypt()`, `.getEncryptionKey()`, and `.getPrimaryUser()` functions have all given me grief over this issue, which is why I was motivated to just make a global config flag to be set for this.

**Comment by twiss** [comment]
> How would a key be a "future key" (creation time in the future), but be expired?

I just meant that `date: null` allows not just future keys but also expired keys. Obviously a key can't be both from the future and already expired :) But you might want to allow only future keys, not expired ones.

> I think that's a great idea.

Would you be interested in modifying this PR, or creating a new PR with that config option?

**Comment by larabr** [comment]
> I don't like that I have to fake the time of encryption

Note that based on the snippet you gave, you are not faking/changing the encryption time, because the message date is set by `createMessage`, whose date defaults to the current time. This means that the encrypted Literal Data packet will contain the current date, not the future one. What you're doing by passing `date` to encrypt, is just provide the time relevant for encryption key selection.

If you were passing also `signingKeys` to encrypt, then the `date` would _also_ be used as creation time of the signatures (and in this sense, yes, you'd be "faking" the signing time). About that, how would a grace period config interact with the signature time to be set, @twiss ? Wouldn't the user still have to pass a `date` that is >= the [signing] key creation time?

> if a message is encrypted at a date BEFORE the key is creation time, isn't that something that client software should just warn about?

The message date is not necessarily relevant for encrypted-only messages (the date information is not authenticated). However, if the message is signed, and the signature date is in the future, then indeed the client should throw a verification error, if it expects the signature to be valid at a time prior to the signature time.

> date: (await publicKey.getEncryptionKey(undefined, null)).bindingSignatures[0].created

FYI as long as the key and subkeys are generated together, it should suffice to just pass `date: publicKey.getCreationTime()`.
If that's not the case and subkey creation date may be different, you can do `publicKey.getEncryptionKey(...).then(encryptionKey => encryptionKey.getCreationTime())`. But as @twiss said, in this case passing `null` as date to `getEncryptionKey` might not be what you want, because it will potentially pick the creation date of expired keys; instead, I would pass the latest time in the future that you deem acceptable, accounting for the time skew (e.g. current time + X hours).

Overall, with the current lib, to use a future date only if needed, while still detecting expired keys, you could do:
```js
const now = new Date();
const GRACE_PERIOD_MS = 3600 * 2 * 1000; // two hours, as example
const keyCreationTime = publicKey.getCreationTime();

if (keyCreationTime > (now + GRACE_PERIOD_MS)) {
  throw new Error('key is too far in the future')
} else if (keyCreationTime > now) {
  // key is in the future but within grace period
 const date = keyCreationTime;
 return pgp.encrypt({
    message: await pgp.createMessage({ text: message, date }), // pass `date` here as well
    encryptionKeys: publicKey,
    date
  });
} else {
 // key creation is not in the future, no need to overwrite date
 return pgp.encrypt({
   message: await pgp.createMessage({ text: message }),
   encryptionKeys: publicKey
  });
}

```

twiss:
> On that note, @twiss: how would a grace period config interact with the signature time to be set ? Wouldn't the user still have to pass a `date` that is >= the [signing] key creation time?

Yes, you're right, that's a bit annoying indeed.

> Overall, with the current lib, to use a future date only if needed, while still detecting expired keys, you could do:

Maybe that's better indeed. @Logicwax does this solution work for you?

PR #1618, opened 2023-03-23
Add additionalAllowedPackets config option

toberndo:
Following the discussion https://github.com/openpgpjs/openpgpjs/discussions/1617 this pull requests adds a new config option `additionalAllowedPackets`.

When parsing for example a SEIP packet, currently only the following nested packets are allowed: `LiteralDataPacket, CompressedDataPacket, OnePassSignaturePacket, SignaturePacket`.

With this new config option we can add additional packet classes to the list of allowed packets.

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -178,6 +178,14 @@ export default {
    */
   ignoreMalformedPackets: false,
   /**
+   * Parsing of packets is normally restricted to a predefined set of packets. For example a Sym. Encrypted Integrity Protected Data Packet can only
+   * contain a certain set of packets including LiteralDataPacket. With this setting we can allow additional packets, which is probably not advisable
+   * as a global config setting, but can be used for specific function calls (e.g. decrypt method of Message).
+   * @memberof module:config
+   * @property {Array} additionalAllowedPackets Allow additional packets on parsing. Defined as array of packet classes, e.g. [PublicKeyPacket]
+   */
+  additionalAllowedPackets: [],
+  /**
    * @memberof module:config
    * @property {Boolean} showVersion Whether to include {@link module:config/config.versionString} in armored messages
    */
--- a/src/packet/packetlist.js
+++ b/src/packet/packetlist.js
@@ -64,6 +64,9 @@ class PacketList extends Array {
    * @async
    */
   async read(bytes, allowedPackets, config = defaultConfig) {
+    if (config.additionalAllowedPackets.length) {
+      allowedPackets = { ...allowedPackets, ...util.constructAllowedPackets(config.additionalAllowedPackets) };
+    }
     this.stream = stream.transformPair(bytes, async (readable, writable) => {
       const writer = stream.getWriter(writable);
       try {
--- a/test/general/packet.js
+++ b/test/general/packet.js
@@ -1052,5 +1052,20 @@ kePFjAnu9cpynKXu3usf8+FuBw2zLsg1Id1n7ttxoAte416KjBN9lFBt8mcu
       expect(parsed.length).to.equal(1);
       expect(parsed[0].tag).to.equal(openpgp.enums.packet.userID);
     });
+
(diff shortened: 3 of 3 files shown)


src/packet/packetlist.js
```diff
@@ -18,7 +18,10 @@ import defaultConfig from '../config';
  * @returns {Object} New packet object with type based on tag
  * @throws {Error|UnsupportedError} for disallowed or unknown packets
  */
-export function newPacketFromTag(tag, allowedPackets) {
+export function newPacketFromTag(tag, allowedPackets, config = defaultConfig) {
+  if (config.additionalAllowedPackets && config.additionalAllowedPackets.length) {
+    allowedPackets = { ...allowedPackets, ...util.constructAllowedPackets(config.additionalAllowedPackets) };
```
twiss, line 23: This function is now called for every packet (if the config option is set); since `util.constructAllowedPackets` is already exported, I would just make the application call that, and pass the returned object in the config. That way, the object can also be shared across calls to OpenPGP.js.

src/config/config.js
```diff
@@ -177,6 +177,11 @@ export default {
    * @property {Boolean} ignoreMalformedPackets Ignore malformed packets on parsing instead of throwing an error
    */
   ignoreMalformedPackets: false,
+  /**
+   * @memberof module:config
+   * @property {Array} additionalAllowedPackets Allow additional packets on parsing. Defined as array of packet classes, e.g. [PublicKeyPacket]
```
larabr, line 182: I'd add a comment that this config should be set locally, for specific function calls (e.g. `readMessage`). Otherwise it's quite unclear how a global setting is useful IMO.

src/packet/packetlist.js
toberndo: I was also wondering where to put the logic. We could also move it up the call stack into `PacketList.read` or `PacketList.fromBinary` to reduce the number of calls.

I find additionalAllowedPackets as an Array easier to document, but if you prefer the object map here then I can change that as well. But I don't see that `util.constructAllowedPackets` is exported and util seems to be a private module?

twiss: Oh, you're right, I forgot we unexported `util`. Yeah, then let's put it in `PacketList.read`?

src/config/config.js
toberndo: I added a comment to describe the option in more detail. Should we change the initial value in the config from `[]` to `undefined`? If we don't want to have it in the global config anyway and this would simplify the check on initial in the coding.

src/packet/packetlist.js
```diff
@@ -64,6 +64,9 @@ class PacketList extends Array {
    * @async
    */
   async read(bytes, allowedPackets, config = defaultConfig) {
+    if (config.additionalAllowedPackets && config.additionalAllowedPackets.length) {
```
larabr, line 67: I don't think the check is needed? `constructAllowedPackets` will simply return an empty map on `[]` input.

src/config/config.js
larabr: I would leave `[]` so that it's clearer what kind of input is expected. Regarding the check, I commented [below](https://github.com/openpgpjs/openpgpjs/pull/1618/files#r1149186658) but I think it's actually simpler as-is

twiss: Either way is fine by me, but even now already you can simplify the check I think, to `if (config.additionalAllowedPackets.length)`, since it should always be an array. If you make it undefined by default, then indeed you can do `if (config.additionalAllowedPackets)`, I guess that's even simpler indeed ^.^

toberndo: I changed the check to `if (config.additionalAllowedPackets.length)`. Looks good to me now overall.

twiss (approved):
:+1: Thanks!

toberndo:
Are there any plans for a new v5.x release before going to v6?

twiss:
Yeah, we can make a release, I'll do so :+1:

toberndo:
@twiss Thanks so much for this!

PR #1620, opened 2023-03-24
`crypto-refresh`: add support for new Ed25519/X25519 keys, signatures and messages

larabr:
This addition is backwards compatible. We offer no way to generate v4 keys in the new format, since existing implementations might not support them.

TODO
- [x] update kdf to include public key
- [x] test with messages (to decrypt and verify) from gopenpgp
- [x] ~merge after introducing support for intended recipient (no longer needed after spec's kdf change)~
- [x] enforce using AES (https://gitlab.com/openpgp-wg/rfc4880bis/-/merge_requests/276) (done in x448 PR, so it's only needed if we don't merge that on in v5)
- [x] merge lukas changes
- [x] ~change base branch (to be merged into v6)~ we merge to v5 as there are no breaking changes

### DIFF
--- a/src/crypto/hkdf.js
+++ b/src/crypto/hkdf.js
@@ -9,13 +9,53 @@ import util from '../util';
 
 const webCrypto = util.getWebCrypto();
 const nodeCrypto = util.getNodeCrypto();
+const nodeSubtleCrypto = nodeCrypto && nodeCrypto.webcrypto && nodeCrypto.webcrypto.subtle;
 
-export default async function HKDF(hashAlgo, key, salt, info, length) {
+export default async function HKDF(hashAlgo, inputKey, salt, info, outLen) {
   const hash = enums.read(enums.webHash, hashAlgo);
   if (!hash) throw new Error('Hash algo not supported with HKDF');
 
-  const crypto = webCrypto || nodeCrypto.webcrypto.subtle;
-  const importedKey = await crypto.importKey('raw', key, 'HKDF', false, ['deriveBits']);
-  const bits = await crypto.deriveBits({ name: 'HKDF', hash, salt, info }, importedKey, length * 8);
-  return new Uint8Array(bits);
+  if (webCrypto || nodeSubtleCrypto) {
+    const crypto = webCrypto || nodeSubtleCrypto;
+    const importedKey = await crypto.importKey('raw', inputKey, 'HKDF', false, ['deriveBits']);
+    const bits = await crypto.deriveBits({ name: 'HKDF', hash, salt, info }, importedKey, outLen * 8);
+    return new Uint8Array(bits);
+  }
+
+  if (nodeCrypto) {
+    const hashAlgoName = enums.read(enums.hash, hashAlgo);
+    // Node-only HKDF implementation based on https://www.rfc-editor.org/rfc/rfc5869
+
+    const computeHMAC = (hmacKey, hmacMessage) => nodeCrypto.createHmac(hashAlgoName, hmacKey).update(hmacMessage).digest();
+    // Step 1: Extract
+    // PRK = HMAC-Hash(salt, IKM)
+    const pseudoRandomKey = computeHMAC(salt, inputKey);
+
+    const hashLen = pseudoRandomKey.length;
+
+    // Step 2: Expand
+    // HKDF-Expand(PRK, info, L) -> OKM
+    const n = Math.ceil(outLen / hashLen);
+    const outputKeyingMaterial = new Uint8Array(n * hashLen);
+
+    // HMAC input buffer updated at each iteration
+    const roundInput = new Uint8Array(hashLen + info.length + 1);
+    // T_i and last byte are updated at each iteration, but `info` remains constant
+    roundInput.set(info, hashLen);
+
+    for (let i = 0; i < n; i++) {
+      // T(0) = empty string (zero length)
+      // T(i) = HMAC-Hash(PRK, T(i-1) | info | i)
+      roundInput[roundInput.length - 1] = i + 1;
+      // t = T(i+1)
+      const t = computeHMAC(pseudoRandomKey, i > 0 ? roundInput : roundInput.subarray(hashLen));
+      roundInput.set(t, 0);
+
+      outputKeyingMaterial.set(t, i * hashLen);
+    }
+
+    return outputKeyingMaterial.subarray(0, outLen);
+  }
+
+  throw new Error('No HKDF implementation available');
 }
--- a/src/crypto/hkdf.js
+++ b/src/crypto/hkdf.js
@@ -1,0 +1,21 @@
+/**
+ * @fileoverview This module implements HKDF using either the WebCrypto API or Node.js' crypto API.
+ * @module crypto/hkdf
+ * @private
+ */
+
+import enums from '../enums';
+import util from '../util';
+
+const webCrypto = util.getWebCrypto();
+const nodeCrypto = util.getNodeCrypto();
+
+export default async function HKDF(hashAlgo, key, salt, info, length) {
+  const hash = enums.read(enums.webHash, hashAlgo);
+  if (!hash) throw new Error('Hash algo not supported with HKDF');
+
+  const crypto = webCrypto || nodeCrypto.webcrypto.subtle;
+  const importedKey = await crypto.importKey('raw', key, 'HKDF', false, ['deriveBits']);
+  const bits = await crypto.deriveBits({ name: 'HKDF', hash, salt, info }, importedKey, length * 8);
+  return new Uint8Array(bits);
+}
--- a/src/crypto/public_key/elliptic/eddsa_legacy.js
+++ b/src/crypto/public_key/elliptic/eddsa_legacy.js
@@ -16,8 +16,9 @@
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
 /**
- * @fileoverview Implementation of EdDSA following RFC4880bis-03 for OpenPGP
- * @module crypto/public_key/elliptic/eddsa
+ * @fileoverview Implementation of legacy EdDSA following RFC4880bis-03 for OpenPGP.
+ * This key type has been deprecated by the crypto-refresh RFC.
+ * @module crypto/public_key/elliptic/eddsa_legacy
  * @private
  */
 
@@ -26,30 +27,12 @@ import nacl from '@openpgp/tweetnacl/nacl-fast-light';
 import util from '../../../util';
 import enums from '../../../enums';
 import hash from '../../hash';
-import { getRandomBytes } from '../../random';
 
 nacl.hash = bytes => new Uint8Array(sha512().update(bytes).digest());
 
 /**
- * Generate (non-legacy) EdDSA key
- * @param {module:enums.publicKey} algo - Algorithm identifier
- * @returns Promise<{ A, seed }>
- */
-export async function generate(algo) {
-  switch (algo) {
-    case enums.publicKey.ed25519: {
-      const seed = getRandomBytes(32);
-      const { publicKey: A } = nacl.sign.keyPair.fromSeed(seed);
-      return { A, seed };
-    }
-    default:
-      throw new Error('Unsupported EdDSA algorithm');
-  }
-}
-
-/**
- * Sign a message using the provided key
- * @param {module:enums.publicKey} algo - Algorithm identifier
+ * Sign a message using the provided legacy EdDSA key
+ * @param {module:type/oid} oid - Elliptic curve object identifier
  * @param {module:enums.hash} hashAlgo - Hash algorithm used to sign (must be sha256 or stronger)
  * @param {Uint8Array} message - Message to sign
  * @param {Uint8Array} publicKey - Public key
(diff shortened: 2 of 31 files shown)


src/packet/public_key_encrypted_session_key.js
```diff
@@ -123,35 +126,86 @@ class PublicKeyEncryptedSessionKeyPacket {
...
+      const decryptedSessionKey = { sessionKeyAlgorithm: result[0], sessionKey: result.subarray(1) };
+      if (randomSessionKey) {
+        // We must not leak info about the validity of the decrypted checksum or cipher algo.
+        // The decrypted session key must be of the same algo and size as the random session key, otherwise we discard it and use the random data.
+        const isValidPayload = isValidChecksum & decryptedSessionKey.sessionKeyAlgorithm === randomSessionKey.sessionKeyAlgorithm & decryptedSessionKey.sessionKey.length === randomSessionKey.sessionKey.length;
+        return {
+          sessionKey: util.selectUint8Array(isValidPayload, decryptedSessionKey.sessionKey, randomSessionKey.sessionKey),
+          sessionKeyAlgorithm: util.selectUint8(
+            isValidPayload,
+            decryptedSessionKey.sessionKeyAlgorithm || randomSessionKey.sessionKeyAlgorithm,
```
larabr, line 190:
```suggestion
            decryptedSessionKey.sessionKeyAlgorithm,
```
?

src/packet/public_key_encrypted_session_key.js
```diff
@@ -123,35 +126,86 @@ class PublicKeyEncryptedSessionKeyPacket {
...
+      // verify checksum in constant time
+      const result = decryptedData.subarray(0, decryptedData.length - 2);
+      const checksum = decryptedData.subarray(decryptedData.length - 2);
+      const computedChecksum = util.writeChecksum(result.subarray(result.length % 8));
+      const isValidChecksum = computedChecksum[0] === checksum[0] & computedChecksum[1] === checksum[1];
+      const decryptedSessionKey = { sessionKeyAlgorithm: result[0], sessionKey: result.subarray(1) };
+      if (randomSessionKey) {
+        // We must not leak info about the validity of the decrypted checksum or cipher algo.
+        // The decrypted session key must be of the same algo and size as the random session key, otherwise we discard it and use the random data.
+        const isValidPayload = isValidChecksum & decryptedSessionKey.sessionKeyAlgorithm === randomSessionKey.sessionKeyAlgorithm & decryptedSessionKey.sessionKey.length === randomSessionKey.sessionKey.length;
```
larabr, line 185:
```suggestion
        const isMatchingSessionKeyAlgorithm = version === 6 || decryptedSessionKey.sessionKeyAlgorithm === randomSessionKey.sessionKeyAlgorithm;
        const isValidPayload = isValidChecksum & isMatchingSessionKeyAlgorithm & decryptedSessionKey.sessionKey.length === randomSessionKey.sessionKey.length;
```

src/packet/public_key_encrypted_session_key.js
```diff
@@ -123,35 +126,86 @@ class PublicKeyEncryptedSessionKeyPacket {
...
+        return {
+          sessionKey: util.selectUint8Array(isValidPayload, decryptedSessionKey.sessionKey, randomSessionKey.sessionKey),
+          sessionKeyAlgorithm: util.selectUint8(
+            isValidPayload,
+            decryptedSessionKey.sessionKeyAlgorithm || randomSessionKey.sessionKeyAlgorithm,
+            randomSessionKey.sessionKeyAlgorithm
+          )
+        };
+      } else {
+        const isValidPayload = isValidChecksum && (decryptedSessionKey.sessionKeyAlgorithm === null || enums.read(enums.symmetric, decryptedSessionKey.sessionKeyAlgorithm));
```
larabr, line 195:
```suggestion
        const isValidPayload = isValidChecksum && (version === 6 || enums.read(enums.symmetric, decryptedSessionKey.sessionKeyAlgorithm));
```

src/type/mont_symkey.js
```diff
@@ -0,0 +1,63 @@
...
+  read(bytes) {
+    let read = 0;
+    let followLength = bytes[read++];
+    this.algo = followLength % 2 === 1 ? bytes[read++] : null; // session key size is always even
+    followLength -= this.algo !== null;
+    this.data = bytes.subarray(read, read + followLength); read += followLength;
+  }
+
+  /**
+   * Write an ECDHSymmetricKey as an Uint8Array
```
larabr, line 50:
```suggestion
   * Write an MontgomerySymmetricKey as an Uint8Array
```

src/type/ecdh_symkey.js
```diff
@@ -16,7 +16,7 @@
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
 /**
- * Encoded symmetric key for ECDH
+ * Encoded symmetric key for ECDH (incl. x25519)
```
larabr, line 19:
```suggestion
 * Encoded symmetric key for ECDH (incl. legacy x25519)
```

src/message.js
```diff
@@ -345,6 +345,15 @@ export class Message {
...
       undefined;
 
+    await Promise.all(encryptionKeys.map(key => key.getEncryptionKey()
+      .catch(() => null) // ignore key strength requirements
+      .then(maybeKey => {
+        if (maybeKey && (maybeKey.keyPacket.algorithm === enums.publicKey.x25519) && !util.isAES(algo)) {
+          throw new Error('Could not generate a session key compatible with the given `encryptionKeys`: X22519 keys can only be used to encrypt AES session keys; change `config.preferredSymmetricAlgorithm` accordingly.');
+        }
+      })
+    ));
```
twiss, line 355: I think it would be safer to just use AES128 in this scenario. After all, we're always allowed to use that, and the config is called `preferredSymmetricAlgorithm`, it's not a guarantee it will be used (and in this case we can just say that the key doesn't support it - although tbh it would be a bit strange if the key's algorithm preferences claim otherwise).

src/enums.js
```diff
@@ -109,13 +109,22 @@ export default {
...
-    aedsa: 24
+    aedsa: 24,
+    /** ECDH 25519 (encrypt only) */
+    x25519: 25,
+    /** ECDH 448 (encrypt only) */
+    x448: 26,
+    /** EdDSA 25519 (sign only) */
+    ed25519: 27,
+    /** EdDSA 448 (sign only) */
+    eddsa448: 28
```
twiss, line 127:
```suggestion
    ed448: 28
```

src/enums.js
```diff
@@ -109,13 +109,22 @@ export default {
...
     /** Reserved for AEDSA */
-    aedsa: 24
+    aedsa: 24,
+    /** ECDH 25519 (encrypt only) */
+    x25519: 25,
+    /** ECDH 448 (encrypt only) */
+    x448: 26,
+    /** EdDSA 25519 (sign only) */
+    ed25519: 27,
+    /** EdDSA 448 (sign only) */
```
twiss, line 126: I think it would be clearer to just list the names the spec uses
```suggestion
    /** X25519 (Encrypt only) */
    x25519: 25,
    /** X448 (Encrypt only) */
    x448: 26,
    /** Ed25519 (Sign only) */
    ed25519: 27,
    /** Ed448 (Sign only) */
```

src/packet/public_key_encrypted_session_key.js
```diff
@@ -123,35 +123,88 @@ class PublicKeyEncryptedSessionKeyPacket {
...
+    case enums.publicKey.rsaEncryptSign:
+    case enums.publicKey.elgamal:
+    case enums.publicKey.ecdh: {
+      // add checksum
+      return util.concatUint8Array([
+        new Uint8Array([cipherAlgo]),
+        sessionKeyData,
+        util.writeChecksum(sessionKeyData.subarray(sessionKeyData.length % 8))
+      ]);
+
```
twiss, line 156:
```suggestion
```

src/packet/public_key_encrypted_session_key.js
```diff
@@ -123,35 +123,88 @@ class PublicKeyEncryptedSessionKeyPacket {
...
+    }
+    case enums.publicKey.x25519: {
+      return sessionKeyData;
+    }
```
twiss, line 160:
```suggestion
    case enums.publicKey.x25519:
      return sessionKeyData;
```

src/packet/public_key_encrypted_session_key.js
```diff
@@ -123,35 +123,88 @@ class PublicKeyEncryptedSessionKeyPacket {
...
+        } else {
+          throw new Error('Decryption error');
+        }
+      }
+    }
+    case enums.publicKey.x25519: {
+      return {
+        sessionKey: decryptedData
+      };
+    }
```
twiss, line 206:
```suggestion
    case enums.publicKey.x25519:
      return {
        sessionKey: decryptedData
      };
```

src/type/mont_symkey.js
```diff
@@ -0,0 +1,64 @@
...
+   * - 1 octect for the length `l`
+   * - `l` octects of encoded session key data (with optional leading algorithm byte)
+   * @param {Uint8Array} bytes
+   * @returns {Number} Number of read bytes.
+   */
+  read(bytes) {
+    let read = 0;
+    let followLength = bytes[read++];
+    this.algorithm = followLength % 2 === 1 ? bytes[read++] : null; // session key size is always even
+    followLength -= this.algorithm !== null;
```
twiss, line 46:
```suggestion
    this.algorithm = followLength % 2 ? bytes[read++] : null; // session key size is always even
    followLength -= followLength % 2;
```

src/type/mont_symkey.js
```diff
@@ -0,0 +1,64 @@
...
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
```
twiss, line 17: I guess this was copied from somewhere but the copyright notice is not accurate, and I don't think we need one in every file anyway. We could remove all of them in a separate PR imo
```suggestion
```
(and the same thing in the other new files)

src/type/mont_symkey.js
```diff
@@ -0,0 +1,64 @@
...
+ * Encoded symmetric key for x25519
+ * The payload format varies for v3 and v6 PKESK:
+ * the former includes an algorithm byte preceeding the encrypted session key.
+ *
+ * @module type/mont_symkey
+ */
+
+import util from '../util';
+
+class MontgomerySymmetricKey {
```
twiss, line 28: I think this class name is a bit confusing if you don't already know what is meant; how about
```suggestion
class AlgoAndSessionKey {
```
(even though the algorithm isn't always there of course, but that still seems clearer to me)

test/karma.conf.js
```diff
@@ -111,10 +111,10 @@ module.exports = function(config) {
       },
       bs_ios_15: {
```
twiss, line 112:
```suggestion
      bs_ios_14: {
```
(why did this need to be changed btw?)

src/type/mont_symkey.js
larabr: This is named based on the ECDHSymKey class; since it applies to Montgomery curves specifically, I'd leave the explicit reference to the publicKeyAlgo, because the two classes are equivalent if no algorithm is present. Maybe `ECDHMontgomery` is clearer?

src/message.js
larabr: This is an edge case which requires passing `config.preferredSymmetricAlgorithm` with a non-AES algo as well as (all) encryptionKeys with that algo appearing in the prefs.
So I think a specific error makes sense in this case, also considering that AES-128 is not the best fallback for x488, so users might want to go with AES-256 at that point 🤞

test/karma.conf.js
larabr: Safari 15 does not expose WebCrypto on localhost 😢 #1624

src/message.js
twiss: Yeah, fair enough :+1:

src/type/mont_symkey.js
twiss: But the difference has nothing to do with the fact that they're Montgomery curves, and furthermore they're never referred to as such in the OpenPGP standard, so someone familiar only with that would be confused I think.
You could call it `X25519X448SymKey` or something like that? Bit ugly but more explicit.

src/crypto/hkdf.js
```diff
@@ -0,0 +1,67 @@
...
+ * @module crypto/hkdf
+ * @private
+ */
+
+import enums from '../enums';
+import util from '../util';
+
+const webCrypto = util.getWebCrypto();
+const nodeCrypto = util.getNodeCrypto();
+const nodeSubtleCrypto = nodeCrypto && nodeCrypto.webcrypto && nodeCrypto.webcrypto.subtle;
```
twiss, line 12: (For now this is fine but maybe long-term we can just make `getWebCrypto` return Node's Web Crypto too? And then the code can decide by the ordering which they prefer between Node Crypto and Node Web Crypto.)

src/crypto/public_key/elliptic/eddsa_legacy.js
```diff
@@ -0,0 +1,99 @@
...
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
```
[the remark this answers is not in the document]
larabr: This was already there, I just renamed the file. Maybe we should review these copyright notes in a separate PR?

twiss (approved):

PR #1623, opened 2023-03-31
Fix shorthand check on user revoked status in getPrimaryUser method

toberndo:
`User` does not have a `revoked` property, instead we need to check the `selfCertification`.


twiss (approved):
:+1: Thanks!

PR #1625, opened 2023-03-31
`crypto-refresh`: add support for Ed488/X448

larabr:
This addition is backwards compatible. We offer no way to generate v4 keys in the new format, since existing implementations might not support them.

TODO:
- [ ]  merge after #1620

Issue #1608, opened 2023-03-06
Reject insecure (RSA1024) primary key when verifying subkeys


rrrooommmaaa:
- OpenPGP.js version: 5.6.0
- Both Browser or Node.js

In this setup PK RSA1024 [certifyKeys, signData], SK RSA4096 [encryptCommunication], we would expect the key to be unusable for both signing and encryption because the encryption subkey can't be securely verified with the weak primary key.
However, the library allows to encrypt with this key regardless of `minRSABits` setting value.
https://github.com/FlowCrypt/flowcrypt-browser/issues/4905#issuecomment-1445841564

twiss:
Hello :wave: To add some background, the reason it works this way is because we still wanted to allow decryption using 1024-bit keys by default (but not encryption), so we allow verifying the binding signatures - but probably we should reject them when using the key for encryption (or signing/verifying), indeed. Long-term, perhaps we could let `minRSABits` apply to all operations (including decryption), and just let applications pass a lower value for decryption if needed - but the risk is that some users of OpenPGP.js will just set a lower `minRSABits` globally.

tomholub:
> we still wanted to allow decryption using 1024-bit keys by default (but not encryption), so we allow verifying the binding signatures - but probably we should reject them when using the key for encryption (or signing/verifying), indeed.

This makes sense - still allowing decryption, but forbid for signing/verifying/encryption, which you could do without introducing additional configuration.

From there, you'll see if there is a need for introducing additional configuration, depending on use-cases that come up.

Issue #720, opened 2018-06-15
MV-02-005 OpenPGP.js: Not all elliptic curve primitives are constant time (Low)


cure53:
It was found that the elliptic curve ciphers offered by OpenPGP.js are exposed by different providers depending on the underlying client platform. Some curves (_secp256k1_, all brainpool variants, _Curve25519_ and _ED25519_) are only exposed via the _elliptic.js_ provider. That provider does not appear to implement constant-time logic for elliptic curve operations.

For _Curve25519_ and _ED25519_, this less of a problem since these curves use the Montgomery elliptic curve equations, which are almost always constant time but with some notable exceptions. However, OpenPGP.js will always run scalar multiplication operations on the other curves in non-constant time. This is because Edwards curves and Weierstrass curves require custom logic in order to behave in a constant time fashion.

**Affected Code:**
https://github.com/openpgpjs/elliptic.git

Since JavaScript makes constant-time execution more difficult to reliably achieve, there exists no sure-fire recommendation for achieving constant time operation on these primitives in the current library. Instead, we recommend simply prioritizing the migration in to lower-level implementations. These can be eventually provided by NodeCrypto (which uses OpenSSL) and WebCrypto.

However, recent research interest in producing constant time primitives in WebAssembly may also lead to easily loadable web libraries that perform all ECC operations in constant time in the near future.

PR #1629, opened 2023-04-03
V6

twiss:
Breaking changes at a platform level:
- Drop support for Node 14 (EOL end of April '23), Node 16 and 17.
- Drop support for platforms without native BigInts (e.g. Safari 13 and below, see [full compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#browser_compatibility))
- Require WebCrypto on web (insecure origins no longer supported)
- Web Streaming: remove embedded Web Streams ponyfill, since it's now supported in all browsers (applications can load a polyfill themselves instead, if they need to support older browser versions).
- Node Streaming: drop support for native Node Readable stream: require passing Node Web Streams
- Declare lib as an ESM module (`type: module` in package.json; legacy entrypoints are still available)

(Further changes to be added via PRs to this branch.)

TODO:
- [x] update readme
- [x] Use WebCrypto's AES_CBC for AES-KW when available (replacing asmcrypto's AES_ECB) (relates to cipher imports in #1654)
- [x] use SubtleCrypto in Node when available (? in some cases, NodeCrypto supports more protocols, see RSA encryption and CFB)
- [ ] add native support for EdDSA and Montgomery ECDH (curve 25519) -> change reverted, waiting for Firefox 130 & 131 to go down in [usage](https://gs.statcounter.com/#all-browser_version-ww-monthly-202409-202409-bar) (support fixed in v132)
  - [ ] keep key material stored in WebCrypto (v6.1?) 
- [x] OpenPGP spec changes listed in #1442
- [x] #1663 
- [x] deprecate/drop `enums.publicKey.eddsa` (only keep `enums.publicKey.eddsaLegacy`) 
   - [x] rename NIST curves values in `enums.curve` (`p256` -> `nistP256`)
   - [x] clean up `enums.curve` "aliases" (e.g. OIDs)? are they used?
- [x] CI: test Node v20 release instead of Canary
- [x] throw if webcrypto not available (in getWebCrypto) -> need to handle Node16, where global `crypto` is not defined
- [x] consider missing key flags as "no capability" as opposed to "any" (e.g. [here](https://github.com/openpgpjs/openpgpjs/blob/main/src/key/helper.js#L391) ) #1677
- [x] drop support for NodeStreams (currently internally converted; require WebStreams directly) #1716 
- [x] replace Elliptic with noble-curves (to address #720 )
- [x] https://github.com/openpgpjs/openpgpjs/issues/1608
- [x] ~add eddsa signature validation after signing (except for v6 sigs, which are non-deterministic thanks to the salt) -> waiting for stats about how frequent random bitflips may be~ randomise v4 signatures by adding notation subpacket (optional, based on new `config` option)
- [x] skip [redundant] key validation for AEAD-encrypted keys (crypto-refresh format)
- [x] gpg parsing failure for armored messages without leading `=` (this is the actual cause of them failing to parse detached signatures without checksum, it's not due to the latter, just a coincidence)
  - [x] the armor checksum is currently added to all detached signatures, but it's prohibited in v6

Includes changes from sub-PRs:
- #1645
- #1630 
- #1625 
- #1676
- #1694 
- #1677
- #1724 
- #1754 
- #1755
- #1756


test/crypto/elliptic.js
```diff
@@ -161,99 +177,81 @@ module.exports = () => describe('Elliptic Curve Cryptography @lightweight', func
...
-        const oid = curve.oid;
-        const message = p384_message;
-        return elliptic_curves.ecdsa.sign(oid, 10, message, keyPublic, keyPrivate, await hashMod.digest(10, message)).then(async signature => {
-          await expect(elliptic_curves.ecdsa.verify(oid, 10, signature, message, keyPublic, await hashMod.digest(10, message)))
-            .to.eventually.be.true;
-        });
+    const curves = ['secp256k1' , 'p256', 'p384', 'p521', 'brainpoolP256r1', 'brainpoolP384r1', 'brainpoolP512r1'];
+    curves.forEach(curveName => it(`${curveName} - Sign and verify message`, async function () {
+      const curve = new elliptic_curves.CurveWithOID(curveName);
+      const { publicKey: keyPublic, privateKey: keyPrivate } = await curve.genKeyPair();
```
larabr, line 245:
```suggestion
      const { publicKey: keyPublic, privateKey: keyPrivate } = await curve.genKeyPair();
```
`genKeyPair()` does not pad the returned values. This causes random test failures in Node as some secret keys are 1 byte short. We should use `elliptic.generate` instead.

larabr (approved):

PR #1505, opened 2022-03-03
Draft: update V5 keys, signatures, SKESK and SEIPD to the latest draft of the specification

twiss:
This fixes part of #1442.

PR #1630, opened 2023-04-03
[v6] Update packet versions to the latest crypto refresh

twiss:
Implement the new packet versions from [draft-ietf-openpgp-crypto-refresh-08](https://datatracker.ietf.org/doc/html/draft-ietf-openpgp-crypto-refresh-08).

Fixes part of #1442; replaces #1505.

TODO:

- [x] update AEAD-encryption mechanism for private keys
- [x] cross-check with go-crypto
- [x] rebase on top of Ed25519 update to then add test vectors from RFC
- [x] add changes from https://github.com/twiss/openpgpjs/pull/3
- [x] add config to support v4 keys with legacy AEAD mode (from werner's draft) -- TODO in separate PR
- [x] properly squash commits

### DIFF
--- a/src/encoding/armor.js
+++ b/src/encoding/armor.js
@@ -369,21 +369,18 @@ export function armor(messageType, body, partIndex, partTotal, customComment, co
     hash = body.hash;
     body = body.data;
   }
-  const bodyClone = stream.passiveClone(body);
   const result = [];
   switch (messageType) {
     case enums.armor.multipartSection:
       result.push('-----BEGIN PGP MESSAGE, PART ' + partIndex + '/' + partTotal + '-----\n');
       result.push(addheader(customComment, config));
       result.push(base64.encode(body));
-      result.push('=', getCheckSum(bodyClone));
       result.push('-----END PGP MESSAGE, PART ' + partIndex + '/' + partTotal + '-----\n');
       break;
     case enums.armor.multipartLast:
       result.push('-----BEGIN PGP MESSAGE, PART ' + partIndex + '-----\n');
       result.push(addheader(customComment, config));
       result.push(base64.encode(body));
-      result.push('=', getCheckSum(bodyClone));
       result.push('-----END PGP MESSAGE, PART ' + partIndex + '-----\n');
       break;
     case enums.armor.signed:
@@ -393,35 +390,30 @@ export function armor(messageType, body, partIndex, partTotal, customComment, co
       result.push('\n-----BEGIN PGP SIGNATURE-----\n');
       result.push(addheader(customComment, config));
       result.push(base64.encode(body));
-      result.push('=', getCheckSum(bodyClone));
       result.push('-----END PGP SIGNATURE-----\n');
       break;
     case enums.armor.message:
       result.push('-----BEGIN PGP MESSAGE-----\n');
       result.push(addheader(customComment, config));
       result.push(base64.encode(body));
-      result.push('=', getCheckSum(bodyClone));
       result.push('-----END PGP MESSAGE-----\n');
       break;
     case enums.armor.publicKey:
       result.push('-----BEGIN PGP PUBLIC KEY BLOCK-----\n');
       result.push(addheader(customComment, config));
       result.push(base64.encode(body));
-      result.push('=', getCheckSum(bodyClone));
       result.push('-----END PGP PUBLIC KEY BLOCK-----\n');
       break;
     case enums.armor.privateKey:
       result.push('-----BEGIN PGP PRIVATE KEY BLOCK-----\n');
       result.push(addheader(customComment, config));
       result.push(base64.encode(body));
-      result.push('=', getCheckSum(bodyClone));
       result.push('-----END PGP PRIVATE KEY BLOCK-----\n');
       break;
     case enums.armor.signature:
       result.push('-----BEGIN PGP SIGNATURE-----\n');
       result.push(addheader(customComment, config));
       result.push(base64.encode(body));
-      result.push('=', getCheckSum(bodyClone));
       result.push('-----END PGP SIGNATURE-----\n');
       break;
   }
--- a/src/encoding/armor.js
+++ b/src/encoding/armor.js
@@ -128,24 +128,21 @@ function verifyHeaders(headers) {
 }
 
 /**
- * Splits a message into two parts, the body and the checksum. This is an internal function
- * @param {String} text - OpenPGP armored message part
- * @returns {Object} An object with attribute "body" containing the body.
- * and an attribute "checksum" containing the checksum.
+ * Remove the (optional) checksum from an armored message.
+ * @param {String} text - OpenPGP armored message
+ * @returns {String} The body of the armored message.
  * @private
  */
-function splitChecksum(text) {
+function removeChecksum(text) {
   let body = text;
-  let checksum = '';
 
   const lastEquals = text.lastIndexOf('=');
 
   if (lastEquals >= 0 && lastEquals !== text.length - 1) { // '=' as the last char means no checksum
     body = text.slice(0, lastEquals);
-    checksum = text.slice(lastEquals + 1).substr(0, 4);
   }
 
-  return { body: body, checksum: checksum };
+  return body;
 }
 
 /**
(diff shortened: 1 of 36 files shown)


src/packet/sym_encrypted_integrity_protected_data.js
```diff
@@ -113,33 +150,148 @@ class SymEncryptedIntegrityProtectedDataPacket {
...
+  /**
+   * En/decrypt the payload.
+   * NOTE: this function is also used by the AEADEncryptedDataPacket.
+   * @param {encrypt|decrypt} fn - Whether to encrypt or decrypt
+   * @param {Uint8Array} key - The session key used to en/decrypt the payload
+   * @param {Uint8Array | ReadableStream<Uint8Array>} data - The data to en/decrypt
+   * @returns {Promise<Uint8Array | ReadableStream<Uint8Array>>}
+   * @async
+   */
+  async aead(fn, key, data) {
```
larabr, line 197: Can this be a non-class function that simply takes a `packet: AEADEncryptedDataPacket | SymEncryptedIntegrityProtectedDataPacket` argument, and gets called by both classes? So we don't need to use `.apply` to reuse it. Also, doesn't `instanceof` work?

src/key/helper.js
```diff
@@ -146,43 +146,58 @@ export async function getPreferredHashAlgo(key, keyPacket, date = new Date(), us
...
+      desiredCipherSuite :
+      defaultCipherSuite;
+  }
+  const defaultSymAlgo = enums.symmetric.aes128;
+  const desiredSymAlgo = config.preferredSymmetricAlgorithm;
+  return [
+    selfSigs.every(selfSig => selfSig.preferredSymmetricAlgorithms && selfSig.preferredSymmetricAlgorithms.includes(desiredSymAlgo)) ?
+      desiredSymAlgo :
+      defaultSymAlgo,
+    0
```
larabr, line 197: Not a fan of this, I had to go and check whether `enums.aead.cfb === 0` since it got me confused; can we set `null` instead? Or just return `[<symAlgo>]`.
Based on the couple of calls I've seen, it's not clear to me whether we actually need an array -- can we can get away with `{ cipherAlgo, aeadAlgo }` ?

twiss: :+1: I've changed it to `{ symmetricAlgo, aeadAlgo }`, with `undefined` meaning CFB.

src/packet/sym_encrypted_integrity_protected_data.js
twiss: It's possible, but I don't want to change the code to accommodate the `AEADEncryptedDataPacket` use case too much; I would rather that the `SymEncryptedIntegrityProtectedDataPacket` is most easy to read, and `AEADEncryptedDataPacket` piggy-backs off of that. That way, if we can ever drop support for `AEADEncryptedDataPacket`, the code can stay essentially the same.
And, `instanceof` also works but it walks the entire prototype chain, which in this case I think isn't necessary since we don't have subclasses of `SymEncryptedIntegrityProtectedDataPacket` and `AEADEncryptedDataPacket`, so we can just check the constructor directly; that might also be slightly faster.

.eslintrc.js
```diff
@@ -109,7 +109,7 @@ module.exports = {
     // Custom errors:
     'no-use-before-define': [2, { 'functions': false, 'classes': true, 'variables': false }],
     'no-constant-condition': [2, { 'checkLoops': false }],
-    'new-cap': [2, { 'properties': false, 'capIsNewExceptionPattern': 'EAX|OCB|GCM|CMAC|CBC|OMAC|CTR', 'newIsCapExceptionPattern': 'type|hash*' }],
+    'new-cap': [2, { 'properties': false, 'capIsNewExceptionPattern': 'EAX|OCB|GCM|CMAC|CBC|OMAC|CTR|HKDF', 'newIsCapExceptionPattern': 'type|hash*' }],
```
larabr, line 112: Very minor but HKDF is really just a function that computes the hash, whereas all the others are somewhat class-like in that they return functions, so I wouldn't appease the linter and just rename the function to `computeHKDF` or the likes.

src/packet/public_key_encrypted_session_key.js
```diff
@@ -67,13 +72,29 @@ class PublicKeyEncryptedSessionKeyPacket {
...
-    this.encrypted = crypto.parseEncSessionKeyParams(this.publicKeyAlgorithm, bytes.subarray(10));
+    if (this.version === 6) {
+      this.publicKeyVersion = bytes[offset++];
+      const fingerprintLength = this.publicKeyVersion ? (this.publicKeyVersion >= 5 ? 32 : 20) : 0;
+      this.publicKeyFingerprint = bytes.subarray(offset, offset + fingerprintLength);
+      offset += fingerprintLength;
+      if (this.publicKeyVersion === 0) {
+        this.publicKeyID = KeyID.wildcard();
+      } else if (this.publicKeyVersion >= 5) {
+        this.publicKeyID.read(this.publicKeyFingerprint);
```
lubux, line 88: Here it should be `this.publicKeyID.read(this.publicKeyFingerprint.subarray(-8));`
For v6 keys the Key ID is the high-order 64 bits of the fingerprint.

lubux: NVM is correct.

src/encoding/armor.js
```diff
@@ -369,59 +273,51 @@ export function armor(messageType, body, partIndex, partTotal, customComment, co
...
       result.push('-----BEGIN PGP MESSAGE, PART ' + partIndex + '-----\n');
       result.push(addheader(customComment, config));
       result.push(base64.encode(body));
-      result.push('=', getCheckSum(bodyClone));
       result.push('-----END PGP MESSAGE, PART ' + partIndex + '-----\n');
       break;
     case enums.armor.signed:
       result.push('-----BEGIN PGP SIGNED MESSAGE-----\n');
-      result.push('Hash: ' + hash + '\n\n');
+      result.push(hash ? `Hash: ${hash} \n\n` : '\n');
```
twiss, line 292:
```suggestion
      result.push(hash ? `Hash: ${hash}\n\n` : '\n');
```

src/message.js
```diff
@@ -543,20 +549,28 @@ export class Message {
...
-      onePassSignatureList.forEach(onePassSig => packetlist.push(onePassSig));
+      onePassSignatureList.forEach(onePassSig => {
+        packetlist.push(onePassSig);
+        signatureSalts.push(onePassSig.salt);
+      });
     });
 
     packetlist.push(literalDataPacket);
-    packetlist.push(...(await createSignaturePackets(literalDataPacket, signingKeys, signature, signingKeyIDs, date, userIDs, notations, false, config)));
+    packetlist.push(...(await createSignaturePackets(literalDataPacket, signingKeys, signature, signingKeyIDs, date, userIDs, notations, signatureSalts, false, config)));
```
twiss, line 569: We could generate the salt when generating the signature packet instead, and then create the one-pass signature packet from the signature packet (though we could also do that as a separate change)

src/packet/sym_encrypted_integrity_protected_data.js
```diff
@@ -54,28 +53,55 @@ class SymEncryptedIntegrityProtectedDataPacket {
   }
 
   constructor() {
-    this.version = VERSION;
+    this.version = 1;
```
larabr, line 56:
```suggestion
    this.version = null;
```
might have to be done in different PR

src/packet/signature.js
```diff
@@ -509,7 +538,7 @@ class SignaturePacket {
         // Issuer Fingerprint
         this.issuerKeyVersion = bytes[mypos++];
         this.issuerFingerprint = bytes.subarray(mypos, bytes.length);
-        if (this.issuerKeyVersion === 5) {
+        if (this.issuerKeyVersion >= 5) {
```
larabr, line 541: Shouldn't we also throw if we read a issuerKeyID packet from a non-v4 key? There is such a key in the sop tests (unhashed packets). We risk that the keyID set here is overwritten, otherwise.

larabr (approved):

PR #1637, opened 2023-04-27
issue #1636 add userID checking on verifyAllUsers()

### LINKED ISSUE #1636: verifyAllUsers() returns userID error for keys that contains image data
- OpenPGP.js version: `"openpgp": "^5.8.0"`
- Affected platform (Browser or Node.js version): Node.js (I haven't tried for the Browser version/dist)

Description:

During my testing, OpenPGP.js `verifyAllUsers()` fails to work and returns `TypeError: Cannot read properties of null (reading 'userID')` when parsing a public key/key pair that contains image data. eg image added with gpg's `addphoto` command.

image

image

File to use: 
[sample.pub.txt](https://github.com/openpgpjs/openpgpjs/files/11319793/sample.pub.txt)

Steps to reproduce:
```javascript
const key = `$armored_key`;
const k = openpgp.readKey({armoredKey: key});
const vk = (await k).verifyAllUsers(k);
console.log(vk);
```

martgil:
This PR adds userID checking on `verifyAllUsers()`.

close #1636

### DIFF
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -690,13 +690,15 @@ class Key {
         await user.verifyAllCertifications(verificationKeys, date, config) :
         [{ keyID: primaryKey.getKeyID(), valid: await user.verify(date, config).catch(() => false) }];
 
-      results.push(...signatures.map(
-        signature => ({
-          userID: user.userID.userID,
-          keyID: signature.keyID,
-          valid: signature.valid
-        }))
-      );
+      if (user.userID !== null && user.userID.userID !== null) {
+        results.push(
+          ...signatures.map((signature) => ({
+            userID: user.userID.userID,
+            keyID: signature.keyID,
+            valid: signature.valid,
+          }))
+        );
+      }
     }));
     return results;
   }


src/key/key.js
```diff
@@ -690,13 +690,15 @@ class Key {
...
         [{ keyID: primaryKey.getKeyID(), valid: await user.verify(date, config).catch(() => false) }];
 
-      results.push(...signatures.map(
-        signature => ({
-          userID: user.userID.userID,
-          keyID: signature.keyID,
-          valid: signature.valid
-        }))
-      );
+      if (user.userID !== null && user.userID.userID !== null) {
```
twiss, line 693: Rather than throwing away the result entirely for user attributes, it might be better to adapt the object below to take them into account.

src/key/key.js
```diff
@@ -690,13 +690,15 @@ class Key {
...
+        results.push(
+          ...signatures.map((signature) => ({
+            userID: user.userID.userID,
```
twiss, line 696: For example, you could do something like:
```suggestion
            userID: user.userID ? user.userID.userID : null,
            userAttribute: user.userAttribute,
```
(untested, lmk if it doesn't work).

twiss (commented):
Thanks for making a PR! See comments below.

src/key/key.js
martgil: Right - thank you very much!

src/key/key.js
martgil: Thank you very much for the help and for allowing me to contribute to the project.

I think it worked!

Code: 
```javascript
const pub = `$armored_key`;
const k = await openpgp.readKey({ armoredKey: pub });
const vk = await k.verifyAllUsers(k);
console.log(vk);
```
Output:
image

twiss: Thanks, and no worries!

Looks good, except it says `valid: false` for both, but I assume that's a separate issue (maybe you're using the wrong key to verify or so)?

If you get it to work, could you add a test case, perhaps? Something like [this](https://github.com/openpgpjs/openpgpjs/blob/726ee55de41a2ca305fce4c12fcd24696869bbc4/test/general/key.js#L2491-L2521) but with a key that has a User Attribute?

martgil: Thanks, I did something wrong with the verification. I have also added a test - thanks for sharing a reference test case.

test/general/key.js
```diff
@@ -3934,4 +4028,48 @@ XvmoLueOOShu01X/kaylMqaT8w==
     expect(signingKeySignature instanceof openpgp.SignaturePacket).to.be.true;
     expect(signingKeySignature.keyFlags[0] & openpgp.enums.keyFlags.signData).to.be.equals(openpgp.enums.keyFlags.signData);
   });
+
+  it("Sign and verify key - key with user attribute", async function () {
```
twiss, line 4032: Could you put the test case next to the other related ones, e.g. after the "Sign key and verify with wrong key - all users" one?

test/general/key.js
```diff
@@ -3934,4 +4028,48 @@ XvmoLueOOShu01X/kaylMqaT8w==
...
+    let publicKey = await openpgp.readKey({ armoredKey: keyWithImageData });
+    const privateKey = await openpgp.decryptKey({
+      privateKey: await openpgp.readKey({ armoredKey: priv_key_rsa }),
+      passphrase: "hello world",
```
twiss, line 4036: The linter complains that this should be
```suggestion
      passphrase: 'hello world'
```
(and also the name of the test should use single quotes :blush:)

test/general/key.js
```diff
@@ -3934,4 +4028,48 @@ XvmoLueOOShu01X/kaylMqaT8w==
...
+      publicKey = await publicKey.signAllUsers([privateKey]);
+      const signatures = await publicKey.verifyAllUsers([privateKey]);
+      const publicSigningKey = await publicKey.getSigningKey();
+      const privateSigningKey = await privateKey.getSigningKey();
+      expect(signatures.length).to.equal(4);
+      expect(signatures[0].userID).to.equal(publicKey.users[0].userID.userID);
+      expect(signatures[0].userAttribute).to.be.null;
+      expect(signatures[0].keyID.toHex()).to.equal(
+        publicSigningKey.getKeyID().toHex()
+      );
```
twiss, line 4051: This is very nitpicky but just for consistency with those ones, could you format these as one line?
(This formatting arguably does look nicer but it visually separates the `.valid` checks from the other ones, which is slightly confusing.)

test/general/key.js
```diff
@@ -3934,4 +4028,48 @@ XvmoLueOOShu01X/kaylMqaT8w==
...
+      expect(signatures[3].userAttribute.attributes[0]).to.be.not.empty;
+      expect(signatures[3].keyID.toHex()).to.equal(
+        privateSigningKey.getKeyID().toHex()
+      );
+      expect(signatures[3].valid).to.be.true;
+    } finally {
+      openpgp.config.minRSABits = minRSABits;
+    }
+  });
 });
```
twiss, line 4075: It's not required but if you still have time & energy, do you want to add a test for the case of verifying with a wrong key (like [this one](https://github.com/openpgpjs/openpgpjs/blob/726ee55de41a2ca305fce4c12fcd24696869bbc4/test/general/key.js#L2523-L2554)) too?

twiss (commented):
Thanks for adding a test case! Some nit picks and further suggestions below.

test/general/key.js
martgil: ok, thank you.

test/general/key.js
martgil: updated, thank you.

test/general/key.js
martgil: Indeed, it's way better in a single line. thank you 😄

test/general/key.js
martgil: yes, no problem. thank you!

test/general/key.js
```diff
@@ -2553,6 +2647,79 @@ function versionSpecificTests() {
     }
   });
 
+  it('Sign and verify a key with user attribute - all users', async function () {
+    let publicKey = await openpgp.readKey({ armoredKey: keyWithImageData });
+    const privateKey = await openpgp.decryptKey({
+      privateKey: await openpgp.readKey({ armoredKey: priv_key_rsa }),
+      passphrase: 'hello world',
```
twiss, line 2654: Thanks! The linter still complains about the trailing comma here (and in the next test). You can also run it locally with `npm run lint` :blush:

martgil: thanks, and sorry for the back and forth 😅

twiss: No worries, and thank for following up on everything! :D

src/key/key.js
```diff
@@ -692,7 +692,8 @@ class Key {
 
       results.push(...signatures.map(
         signature => ({
-          userID: user.userID.userID,
+          userID: user.userID ? user.userID.userID : null,
```
ioanmo226, line 695: I think better to like below?
```ts
userID: user.userID?.userID,
```

ioanmo226:
@twiss, do you have an ETA for when this pull request will be available on npm?

larabr:
Hi @ioanmo226 , we'll make a release this week :)

src/key/key.js
twiss: Sure :+1: I'll convert it into a suggestion and apply it here :upside_down_face:
```suggestion
          userID: user.userID?.userID,
```

twiss: Actually, I'll revert this as the Node.js tests complain

ioanmo226: Do you mean you want/have to support Node.js versions that are below v14?

twiss: The CI tests on Node.js 14+ also failed, perhaps because of esm.js. Not sure, we can look into that separately later if we want to use this syntax

twiss (approved):

src/key/key.js
ioanmo226: Ok, Got it
Optional chaining is supported from Node.js v14 though.

twiss:
We've released [v5.9.0](https://www.npmjs.com/package/openpgp/v/5.9.0) with this change :slightly_smiling_face:

ioanmo226:
Thank you @twiss

PR #1639, opened 2023-05-02
Add `revoke` to `Subkey` in type definition

haryu703:
This PR adds `Subkey.revoke` implemented in following to the type definition.
https://github.com/openpgpjs/openpgpjs/blob/v5.8.0/src/key/subkey.js#L86


twiss (approved):
:+1: Thanks!

PR #1641, opened 2023-05-15
[v6] Relax constraints for UserID email address validity

larabr:
New checks align with the HTML5 W3C spec and should be more lax than the existing ones (meaning, addresses which passed validation before should continue to be valid). Addresses with domain such as `@localhost` are now allowed too, since presence of "." is no longer enforced.

These checks should not be considered exhaustive: library users are encouraged to implement separate checks for email validity if needed.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -457,11 +457,16 @@ const util = {
     return os.cpus().length;
   },
 
+  /**
+   * Test email format based on W3C HTML5 specification.
+   * This check is not exaustive, and includes a willful violation of RFC5322
+   * (see https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email))
+   */
   isEmailAddress: function(data) {
     if (!util.isString(data)) {
       return false;
     }
-    const re = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+([a-zA-Z]{2,}[0-9]*|xn--[a-zA-Z\-0-9]+)))$/;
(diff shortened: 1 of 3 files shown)


test/general/util.js
```diff
@@ -108,19 +108,23 @@ export default () => describe('Util unit tests', function() {
       const data = 'test@example.com';
       expect(util.isEmailAddress(data)).to.be.true;
     });
-    it('should return true for valid email address', function() {
+    it('should return true for valid email address (-- in domain part)', function() {
```
twiss, line 111:
```suggestion
    it('should return true for valid email address (internationalized domain name)', function() {
```

test/general/util.js
```diff
@@ -108,19 +108,23 @@ export default () => describe('Util unit tests', function() {
...
     });
-    it('should return false for invalid email address', function() {
-      const data = 'Test User <test@example.com>';
-      expect(util.isEmailAddress(data)).to.be.false;
+    it('should return true for valid email address (trailing numbers in domain)', function() {
+      const data = 'test1@com.com09';
+      expect(util.isEmailAddress(data)).to.be.true;
     });
-    it('should return false for invalid email address', function() {
+    it('should return false for invalid email address (no . in domain part)', function() {
```
twiss, line 119:
```suggestion
    it('should return true for valid email address (no . in domain part)', function() {
```
(I'm not sure we really need to test this, but if we do, it should match the result :)

test/general/util.js
```diff
@@ -108,19 +108,23 @@ export default () => describe('Util unit tests', function() {
...
       const data = 'test@examplecom';
```
twiss, line 120:
```suggestion
      const data = 'test@localhost';
```
And, might make sense to give a more plausible example at that point

src/util.js
```diff
@@ -457,11 +457,16 @@ const util = {
     return os.cpus().length;
   },
 
+  /**
+   * Test email format based on W3C HTML5 specification.
+   * This check is not exaustive, and includes a willful violation of RFC5322
+   * (see https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email))
```
twiss, line 463: Might make sense to give some justification for using it here, e.g.
```suggestion
   * This check is not exaustive, and does not match RFC 5322 exactly
   * (see https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email)),
   * but is commonly used for email address validation.
```

PR #1643, opened 2023-05-15
CI: set apt mirror to mitigate instability of Github Actions proxy

larabr:


larabr:
Issue was fixed quickly enough, will reopen if the disruption lasts longer in the future.

PR #1644, opened 2023-05-17
TS: Allow nullable date in VerifyOptions

oofdog:
It seems like the library allows passing `null` when calling `verify`, as it checks for null here
- https://github.com/openpgpjs/openpgpjs/blob/400b163f8411204f5de830b70e65fd86246cf933/src/key/helper.js#L73
- https://github.com/openpgpjs/openpgpjs/blob/400b163f8411204f5de830b70e65fd86246cf933/src/util.js#L78

but the TS declaration doesn't reflect that.


twiss (approved):
:+1: Thanks for the PR, and apologies for the delayed review!

PR #1645, opened 2023-05-31
[v6] Switch to pure ESM

### LINKED ISSUE #1607: CI: update Mocha and Sinon to newer versions (using Node's native esm support)
Screenshot (16)

Hi
Some tests are failing in your package. Can you kindly fix them so we can integrate the changes in the Debian package.
Also, I will be very grateful if you can upgrade sinon dependency to at least version 14.0.2

larabr:
This change is primarily internal, as we are keeping support for commonjs targets.

Breaking changes:
downstream bundlers might be affected by the package.json changes depending on how they load the library.
Note that legacy package.json entrypoints are still available.

TODOs:

- [x] flaky stream.cancel test
- [x] ~use WebCrypto over for AES-KW (or at least AES-ECB)~ -> leave to separate PR
- [x] fix JSDoc
- [x] keep using asmcrypto over noble-hash (sha1, sha256), to avoid bundling bn.js? (compare bundle sizes)
    - [x] if we keep noble-hash: import `BigInteger` code from there, drop internal code
- [x] upgrade sinon
- [x] confirm bundle size remains optimised
- [x] release & point to web-stream-tools
- [x] release & point to asmcrypto.js module
- [x] release & point to noble-hash

Fix #1607


src/crypto/mode/cfb.js
```diff
@@ -130,18 +129,120 @@ export async function decrypt(algo, key, ciphertext, iv) {
...
-    pt.length >= 3000 * config.minBytesForWebCrypto // Default to a 3MB minimum. Chrome is pretty slow for small messages, see: https://bugs.chromium.org/p/chromium/issues/detail?id=701188#c2
-  ) { // Web Crypto
-    return webEncrypt(algo, key, pt, iv);
+class WebCryptoEncryptor {
+  constructor(algo, key, iv) {
+    const { blockSize } = getCipher(algo);
+    this.key = key;
+    this.prevBlock = iv;
+    this.nextBlock = new Uint8Array(blockSize);
+    this.i = 0;// pointer inside next block
```
twiss, line 138:
```suggestion
    this.i = 0; // pointer inside next block
```

src/crypto/mode/cfb.js
```diff
@@ -130,18 +129,120 @@ export async function decrypt(algo, key, ciphertext, iv) {
...
+    this.blockSize = blockSize;
+    this.zeroBlock = new Uint8Array(this.blockSize);
+  }
+
+  static async isSupported(algo) {
+    const { keySize } = getCipher(algo);
+    return webCrypto.importKey('raw', new Uint8Array(keySize), 'aes-cbc', false, ['encrypt'])
+      .then(() => true)
+      .catch(() => false);
```
twiss, line 147:
```suggestion
      .then(() => true, () => false);
```

src/crypto/mode/cfb.js
```diff
@@ -130,18 +129,120 @@ export async function decrypt(algo, key, ciphertext, iv) {
...
+    this.prevBlock.fill(0);
+    this.keyRef = null;
+    this.key = null;
+  }
+
+  async encrypt(plaintext) {
+    const ct = await this._runCBC(
+      util.concatUint8Array([new Uint8Array(this.blockSize), plaintext]),
+      this.iv
+    ).then(result => result.subarray(0, plaintext.length));
```
twiss, line 231:
```suggestion
    const ct = (await this._runCBC(
      util.concatUint8Array([new Uint8Array(this.blockSize), plaintext]),
      this.iv
    )).subarray(0, plaintext.length);
```

src/crypto/mode/cfb.js
```diff
@@ -153,21 +254,11 @@ function aesDecrypt(algo, key, ct, iv) {
 }
 
 function xorMut(a, b) {
-  for (let i = 0; i < a.length; i++) {
+  for (let i = 0; i < Math.min(a.length, b.length); i++) {
```
twiss, line 257: This runs `Math.min(a.length, b.length)` on every iteration of the loop. The browser might optimize it, but it's better to be safe and hoist it out into a `const`

src/packet/compressed_data.js
```diff
@@ -168,18 +166,45 @@ function node_zlib(func, create, options = {}) {
...
+    return new ReadableStream({
+      async start(controller) {
+        zlibStream.ondata = async (value, isLast) => {
+          controller.enqueue(value);
+          if (isLast) {
+            controller.close();
+          }
+        };
+
+        while (true) {
```
twiss, line 199: Since it seems that fflate calls `ondata` synchronously in all these cases (right?), you can actually handle backpressure without any extra effort by doing something like:
```suggestion
      }

      async pull() {
```
(and then the `pull` function will get called repeatedly until the internal queue is full).

twiss (commented):
Bunch of nitpicks below, but overall LGTM :+1:

src/crypto/cmac.js
```diff
@@ -73,20 +72,25 @@ export default async function CMAC(key) {
...
+    if (util.getWebCrypto()) {
+      key = await webCrypto.importKey('raw', key, { name: 'AES-CBC', length: key.length * 8 }, false, ['encrypt']);
+      return async function(pt) {
+        const ct = await webCrypto.encrypt({ name: 'AES-CBC', iv: zeroBlock, length: blockLength * 8 }, key, pt);
+        return new Uint8Array(ct).subarray(0, ct.byteLength - blockLength);
+      };
+    }
+  } catch (err) {
+    // no 192 bit support in Chromium, see: https://www.chromium.org/blink/webcrypto#TOC-AES-support
+    if (err.name !== 'OperationError') throw err;
```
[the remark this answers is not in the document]
larabr: The problem I have with `NotSupportedError` is that we'd have to potentially add it everywhere, and I'd do that only if there is a relevant use-case. AES-GCM and CBC have widespread support.
It seems that Chrome not implementing one AES size is not spec-respecting to begin with: https://github.com/w3c/webcrypto/issues/62 . 
> As the spec is currently written, it is not an option to support AES-GCM but to not support one of the key sizes. 

I'd like to be as strict as possible with these fallbacks as far as decryption goes, since it's sensitive. In this sense, ;ooking at the other mode's code, I think we should probably add a browser + algo check , when it comes to Chrome's OperationError.

With encryption, it's not as important, but since the fallback is unexpected at this point, I would still write specific code, so that we can keep track of other special cases too (and decide how important they are if we want to drop support).

src/crypto/mode/cfb.js
larabr: Is this faster?
I don't find awaits inside `(...)` very readable

twiss: It shouldn't matter much, but I think mixing `async/await` and `then` unnecessarily is a bit messy. If you prefer, you can split it up into a separate variable instead

src/crypto/cmac.js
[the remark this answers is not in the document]
twiss: Yeah, it's not compliant, but at the same time the existence of a spec can't force implementations to implement stuff they don't want to. The spec should accommodate that to some extent, and if we need to add a provision for it, allowing to throw `NotSupportedError` would make much more sense than allowing to throw `OperationError` - so we shouldn't hardcode the latter.

But yes, we can add a browser + key length check instead :+1:

src/packet/compressed_data.js
larabr: After discussing offline, we decided to leave this as future improvement, as the spec says (differing from MDN):

> Additionally, it will only be called repeatedly if it enqueues at least one chunk or fulfills a BYOB request; a no-op pull() implementation will not be continually called.

src/crypto/cmac.js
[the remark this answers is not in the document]
larabr: We do catch [specific browser errors in other parts](https://github.com/openpgpjs/openpgpjs/blob/main/src/crypto/public_key/elliptic/ecdsa.js#L62). I can add `NotSupportedError` as an additional condition, but I think it's important to keep checking for exceptional browser-specific errors too, rather than fallback on all cases (which has given issues in the past with e.g. RSA signature verification in the context of key validation, for instance).
Not all browsers versions give the same error, nor do they fail to begin with, that's why we originally introduced error checking rather than platform checking.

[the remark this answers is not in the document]
twiss: Alrighty, yeah, seems good to me :+1:
(Old versions of Chrome did throw `NotSupportedError` instead of `OperationError`, though to me the more important part is keeping open the possibility of someone throwing that in the future)

twiss (approved):

PR #1676, opened 2023-09-07
[v6] `crypto-refresh`: support generating Curve448 and Curve25519 keys (new format)

larabr:
`generateKey` now accepts `type: 'curve25519'` and  `'curve448'` to generate keys as per crypto-refresh (see #1620 and #1625).

Neither type is set as default for now, since they are not widely supported.
Note that the generate key options `{ type: 'curve25519' }` are not equivalent to `{ type: 'ecc', curve: 'curve25519' }`: the latter is still the default (same as OpenPGP.js v5), and generates keys in legacy - but widely adopted - format.

Breaking change:
- the value `enums.publicKey.eddsa` has been dropped, in favour of `enums.publicKey.eddsaLegacy`. This is to highlight the difference with the new format algo identifier (`enums.publicKey.ed25519`).
- similarly, the string values of `enums.curve.ed25519Legacy` and `.curve25519Legacy` have been renamed from `'ed25519'`/`'curve25519'` to `'ed25519Legacy'`/`'curve25519Legacy'`.

TODO
- [x] merge after #1625


test/general/x25519.js
```diff
@@ -396,7 +396,7 @@ function omnibus() {
       const primaryKey = hi.keyPacket;
       const subkey = hi.subkeys[0];
       expect(hi.getAlgorithmInfo().curve).to.equal('ed25519');
-      expect(hi.getAlgorithmInfo().algorithm).to.equal('eddsa');
+      expect(hi.getAlgorithmInfo().algorithm).to.equal('ed25519Legacy');
```
larabr, line 399: @twiss: I know we already introduced for `ed25519Legacy` on v5, but I was thinking that `eddsaLegacy` would be a better name than `ed25519Legacy`, because
1) the reference with the old name is clear even if someone is not familiar with what ed25519 is, and
2) `getAlgorithmInfo()` (and some errors) will now refer to `algo = ed25519Legacy` and `curve = ed25519`, which I think it's ugly, confusing and ultimately redundant.
What do you think?

src/crypto/public_key/elliptic/ecdh_x.js
```diff
@@ -28,7 +30,12 @@ export async function generate(algo) {
       const k = getRandomBytes(32);
       k[0] &= 248;
       k[31] = (k[31] & 127) | 64;
-      const { publicKey: A } = nacl.box.keyPair.fromSecretKey(k);
+      const { publicKey: A } = X25519.box.keyPair.fromSecretKey(k);
+      return { A, k };
+    }
+    case enums.publicKey.x448: {
+      const k = X448.utils.randomPrivateKey();
```
larabr, line 37:
```suggestion
      const k = X448.utils.randomPrivateKey();
      k[0] &= 252
      k[55] |= 128
```

src/crypto/public_key/elliptic/ecdh_x.js
```diff
@@ -28,7 +30,12 @@ export async function generate(algo) {
       const k = getRandomBytes(32);
       k[0] &= 248;
```
larabr, line 31: No need for this anymore, it only applied to LegacyCurve25519, see [spec](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-10.html#name-curve25519legacy-ecdh-secre).

src/crypto/public_key/elliptic/ecdh_x.js
```diff
@@ -28,7 +30,12 @@ export async function generate(algo) {
...
       k[31] = (k[31] & 127) | 64;
```
larabr, line 32:
```suggestion

```
No need for this anymore, it only applied to LegacyCurve25519, see [spec](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-10.html#name-curve25519legacy-ecdh-secre).

test/general/x25519.js
twiss: Yeah, I agree; actually in the crypto refresh the name is also [EdDSALegacy](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-10.html#name-public-key-algorithms), and then the curve name is [Ed25519Legacy](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-10.html#name-ecc-curves-for-openpgp), we should do the same (uncapitalized) here.

larabr: Should i change v5 too? It might be breaking if people started using those names

twiss: Probably best to change it in v6 only

larabr: Should still add an additional alias in v5, to ease the transition? 😅

src/key/helper.js
```diff
@@ -331,7 +326,7 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
   options.sign = options.sign || false;
 
   switch (options.type) {
-    case 'ecc':
+    case 'ecc': // NB: this case also handles legacy eddsa (ed25519) and x25519 keys, based on `options.curves`
```
twiss, line 329:
```suggestion
    case 'ecc': // NB: this case also handles legacy eddsa (ed25519) and x25519 keys, based on `options.curve`
```

src/key/private_key.js
```diff
@@ -238,4 +244,24 @@ class PrivateKey extends PublicKey {
...
+    case enums.publicKey.elgamal: // impossible since it's from the primary key
+      return 'rsa';
+    case enums.publicKey.ed448:
+    case enums.publicKey.x448:
+      return 'curve448';
+    case enums.publicKey.ed25519:
+    case enums.publicKey.x25519:
+      return 'curve25519';
+    default:
+      return 'ecc';
```
twiss, line 263: Rather than defaulting to ECC, I would prefer to spell out the options here and throw in the default case, such that we'll catch it if we forget to update it for new algorithms later

twiss, line 263: Also, this is very nitpicky but can we try to have a somewhat consistent ordering in these kinds of switches, so that it's easier to find the option you're looking for? I.e. it would be best to match the order of the enums, i.e. rsa, ecc, curve25519, curve448

src/openpgp.js
```diff
@@ -32,11 +32,13 @@ import { checkKeyRequirements } from './key/helper';
...
 /**
- * Generates a new OpenPGP key pair. Supports RSA and ECC keys. By default, primary and subkeys will be of same type.
+ * Generates a new OpenPGP key pair. Supports RSA and ECC keys, as well as the newer Curve448 and Curve25519 keys.
+ * By default, primary and subkeys will be of same type.
  * The generated primary key will have signing capabilities. By default, one subkey with encryption capabilities is also generated.
  * @param {Object} options
  * @param {Object|Array<Object>} options.userIDs - User IDs as objects: `{ name: 'Jo Doe', email: 'info@jo.com' }`
- * @param {'ecc'|'rsa'} [options.type='ecc'] - The primary key algorithm type: ECC (default) or RSA
+ * @param {'ecc'|'rsa'|'curve448'|'curve25519'} [options.type='ecc'] - The primary key algorithm type: ECC (default for v4 keys), RSA, Curve448 or Curve25519 (new format, default for v6 keys).
+ *                                                                     Note: Curve448 or Curve25519 (new format) are not widely supported yet.
```
twiss, line 41:
```suggestion
 *                                                                     Note: Curve448 and Curve25519 (new format) are not widely supported yet.
```

src/packet/public_key.js
```diff
@@ -124,6 +124,17 @@ class PublicKeyPacket {
 
       // - A series of values comprising the key material.
       const { read, publicParams } = crypto.parsePublicKeyParams(this.algorithm, bytes.subarray(pos));
+      // The deprecated OIDs for Ed25519Legacy and Curve25519Legacy are used in legacy version 4 keys and signatures.
+      // Implementations MUST NOT accept or generate v6 key material using the deprecated OIDs.
+      if (
+        this.version === 6 &&
+        publicParams.oid && (
+          publicParams.oid.getName() === enums.curve.X25519 ||
+          publicParams.oid.getName() === enums.curve.Ed25519
```
twiss, line 133:
```suggestion
          publicParams.oid.getName() === enums.curve.ed25519Legacy ||
          publicParams.oid.getName() === enums.curve.x25519Legacy
```

src/packet/secret_key.js
```diff
@@ -484,6 +487,14 @@ class SecretKeyPacket extends PublicKeyPacket {
   }
 
   async generate(bits, curve) {
+    // The deprecated OIDs for Ed25519Legacy and Curve25519Legacy are used in legacy version 4 keys and signatures.
+    // Implementations MUST NOT accept or generate v6 key material using the deprecated OIDs.
+    if (this.version === 6 && (
+      (this.algorithm === enums.publicKey.ecdh && curve === enums.curve.X25519) ||
```
twiss, line 493:
```suggestion
      (this.algorithm === enums.publicKey.ecdh && curve === enums.curve.x25519Legacy) ||
```

larabr: Pending rebase 👍

test/general/x25519.js
larabr: I did at it to the last v5 release, so I'll have to also add `curve.curve25519Legacy` over `curve.x25519Legacy`, but it can be done after merging the outstanding v6 sub-PRs.

twiss (approved):

PR #1677, opened 2023-09-11
[v6] Refuse to use keys without key flags, add `config.allowMissingKeyFlags`

larabr:
Key flags are needed to restrict key usage to specific purposes: https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-10.html#section-5.2.3.29 . Some older keys (e.g. from OpenPGP.js v0) do not declare any key flags. In previous OpenPGP.js versions, we've allowed such keys to be used for any operation for which they were compatible. This behaviour has now changed, and these keys are not allowed to be used for any operation.

The setting  `config.allowMissingKeyFlags` has been added to selectively revert to the past behaviour.

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -174,7 +174,14 @@ export default {
    * @property {Boolean} allowInsecureDecryptionWithSigningKeys
    */
   allowInsecureVerificationWithReformattedKeys: false,
-
+  /**
+   * Allow using keys that do not have any key flags set.
+   * Key flags are needed to restrict key usage to specific purposes: for instance, a signing key could only be allowed to certify other keys, and not sign messages
+   * (see https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-10.html#section-5.2.3.29).
+   * Some older keys do not declare any key flags, which means they are not allowed to be used for any operation.
+   * This setting allows using such keys for any operation for which they are compatible, based on their public key algorithm.
+   */
+  allowMissingKeyFlags: false,
   /**
    * Enable constant-time decryption of RSA- and ElGamal-encrypted session keys, to hinder Bleichenbacher-like attacks (https://link.springer.com/chapter/10.1007/BFb0055716).
    * This setting has measurable performance impact and it is only helpful in application scenarios where both of the following conditions apply:
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -357,7 +357,7 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
   return options;
 }
 
-export function isValidSigningKeyPacket(keyPacket, signature, config) {
+export function validateSigningKeyPacket(keyPacket, signature, config) {
   switch (keyPacket.algorithm) {
     case enums.publicKey.rsaEncryptSign:
     case enums.publicKey.rsaSign:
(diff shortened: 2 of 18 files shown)


src/config/config.js
```diff
@@ -166,7 +166,14 @@ export default {
    * @property {Boolean} allowInsecureDecryptionWithSigningKeys
    */
   allowInsecureVerificationWithReformattedKeys: false,
-
+  /**
+   * Allow using keys that do not have any key flags set.
+   * Key flags are needed to restrict key usage to specific purposes: for instance, a signing key could only be allowed to certify other keys, and not sign messages
+   * (see https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-10.html#section-5.2.3.29).
+   * Some older keys (e.g. from OpenPGP.js v1) do not declare any key flags, which means they are not allowed to be used for any operation.
```
larabr, line 173:
```suggestion
   * Some older keys do not declare any key flags, which means they are not allowed to be used for any operation.
```

src/key/helper.js
```diff
@@ -355,28 +355,49 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
...
+    case enums.publicKey.ecdsa:
+    case enums.publicKey.ed25519Legacy:
+    case enums.publicKey.ed25519: {
+      if (!signature.keyFlags && !config.allowMissingKeyFlags) {
+        throw new Error('None of the key flags is set: consider passing `config.allowMissingKeyFlags`');
+      }
+
+      return !signature.keyFlags ||
+        (signature.keyFlags[0] & enums.keyFlags.signData) !== 0;
+    }
```
twiss, line 372:
```suggestion
    }
    default:
      return false;
```

src/key/helper.js
```diff
@@ -355,28 +355,49 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
...
+    case enums.publicKey.ecdh:
+    case enums.publicKey.x25519: {
+      if (!signature.keyFlags && !config.allowMissingKeyFlags) {
+        throw new Error('None of the key flags is set: consider passing `config.allowMissingKeyFlags`');
+      }
+
+      return !signature.keyFlags ||
+        (signature.keyFlags[0] & enums.keyFlags.encryptCommunication) !== 0 ||
+        (signature.keyFlags[0] & enums.keyFlags.encryptStorage) !== 0;
+    }
```
twiss, line 390:
```suggestion
    }
    default:
      return false;
```

src/key/helper.js
```diff
@@ -355,35 +355,70 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
...
+    case enums.publicKey.ecdsa:
+    case enums.publicKey.ed25519Legacy:
+    case enums.publicKey.ed25519: {
+      if (!signature.keyFlags && !config.allowMissingKeyFlags) {
+        throw new Error('None of the key flags is set: consider passing `config.allowMissingKeyFlags`');
+      }
+
+      return !signature.keyFlags ||
+        (signature.keyFlags[0] & enums.keyFlags.signData) !== 0;
+    }
```
twiss, line 372: Sorry for missing these last time (and being very nitpicky ^.^), but these brackets are also not necessary. They're only necessary if you need a block scope, e.g. if you're creating block-scoped variables.
```suggestion
    case enums.publicKey.ed25519:
      if (!signature.keyFlags && !config.allowMissingKeyFlags) {
        throw new Error('None of the key flags is set: consider passing `config.allowMissingKeyFlags`');
      }

      return !signature.keyFlags ||
        (signature.keyFlags[0] & enums.keyFlags.signData) !== 0;
```

src/key/helper.js
```diff
@@ -355,35 +355,70 @@ export function sanitizeKeyOptions(options, subkeyDefaults = {}) {
...
+    case enums.publicKey.ecdh:
+    case enums.publicKey.x25519: {
+      if (!signature.keyFlags && !config.allowMissingKeyFlags) {
+        throw new Error('None of the key flags is set: consider passing `config.allowMissingKeyFlags`');
+      }
+
+      return !signature.keyFlags ||
+        (signature.keyFlags[0] & enums.keyFlags.encryptCommunication) !== 0 ||
+        (signature.keyFlags[0] & enums.keyFlags.encryptStorage) !== 0;
+    }
```
twiss, line 392:
```suggestion
    case enums.publicKey.x25519:
      if (!signature.keyFlags && !config.allowMissingKeyFlags) {
        throw new Error('None of the key flags is set: consider passing `config.allowMissingKeyFlags`');
      }

      return !signature.keyFlags ||
        (signature.keyFlags[0] & enums.keyFlags.encryptCommunication) !== 0 ||
        (signature.keyFlags[0] & enums.keyFlags.encryptStorage) !== 0;
```

twiss (approved):

src/key/helper.js
larabr: I know they are not needed, but it's not a big deal to add them...? I just find it confusing to have a lot of code without brackets.

twiss: Er, OK, but I think it's more important to have a consistent code style. JS cases normally don't have brackets, and most other code in the library doesn't have them, so we should only add them when needed.

larabr: Ok, but then I'll add a linting rule for this, so that it's automatically enforced

PR #1678, opened 2023-09-11
[v6] Disregard `config.aeadProtect` when encrypting to public keys

larabr:
Determine whether AEAD should be used for encryption solely based the encryption key preferences.
Previously, the config flag was also used to control the behaviour, since AEAD messages were not standardised nor widely supported.

To generate keys that declare AEAD in their preferences, use `generateKey` with `config.aeadProtect = true`.

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -42,12 +42,14 @@ export default {
    * @property {Integer} deflateLevel Default zip/zlib compression level, between 1 and 9
    */
   deflateLevel: 6,
-
   /**
    * Use Authenticated Encryption with Additional Data (AEAD) protection for symmetric encryption.
+   * This setting is relevant for:
+   * - key generation (encryption key preferences)
+   * - password-based message encryption
+   * - private key encryption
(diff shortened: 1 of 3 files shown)


src/config/config.js
```diff
@@ -42,12 +42,14 @@ export default {
    * @property {Integer} deflateLevel Default zip/zlib compression level, between 1 and 9
    */
   deflateLevel: 6,
-
   /**
    * Use Authenticated Encryption with Additional Data (AEAD) protection for symmetric encryption.
+   * This setting is relevant for:
+   * - key generation (encryption key preferences)
+   * - password-based message encryption
+   * - private key encryption
```
twiss, line 50:
```suggestion
   * This option is applicable to:
   * - key generation (encryption key preferences),
   * - password-based message encryption, and
   * - private key encryption.
   * In the case of message encryption using public keys, the encryption key preferences are respected instead.
```

twiss (approved):

Issue #1654, opened 2023-06-20
Exclude CAST5, Blowfish, Twofish and DES from lightweight bundle 


larabr:
Legacy ciphers take up 50kb of space in the lightweight (main) bundle. Since they are not commonly used, we should import them dynamically.

PR #1679, opened 2023-09-12
Fix CFB decryption performance in JS fallback for ciphers other than AES

### LINKED ISSUE #1668: CAST5 file decryption hangs indefinitely
- OpenPGP.js version: 5.9.0
- Affected platform (Browser or Node.js version): 18.17.0

When attempting to decrypt a 512kb csv file, the decryption process hangs indefinitely and the CPU spins up to 100%.

```
const fixturePath = `util/fixtures`;
const private_key = fs.readFileSync('util/fixtures/test_priv_key.txt', 'utf-8');

const file_stream = fs.createReadStream(`${fixturePath}/test_encrypted.csv`);

const message = await openpgp.readMessage({ binaryMessage: file_stream });

const pk = await openpgp.decryptKey({
  privateKey: await openpgp.readKey({
    armoredKey: private_key.toString(),
  }),
  passphrase: 'password',
});

const { data } = await openpgp.decrypt({
  message, 
  verificationKeys: [], 
  decryptionKeys: pk, 
});
```
A few other details of note:
- Sometimes it will complete when using the debugger, but never when calling the decryption script directly
- It works just fine on Node 16.16.0
- If I decrypt the file using Node 16.16.0 and re-encrypt the file using OpenPGP 5.9.0, I can decrypt the newly encrypted file without issue using node 18.17.0
- There are no issues with OpenPGP V4
- Unfortunately, I cannot share the file in question since it contains sensitive information

This seems like a race condition of some sort, but I could be wrong. Any help or suggestions to debug this would be greatly appreciated - we would love to upgrade our codebase to OpenPGP V5 but without understanding the root cause of this issue and building tests to catch this moving forward, we are unable to do so.

larabr:
By calling the underlying cipher decryption function for one block at a time.
Before, the entire ciphertext was passed for decryption, but only a subset of it (equal to the block size) was actually processed in each iteration.
This resulted in a lot of extra work for e.g. Cast5 decryption, whose decryption function supported decrypting more than a block at once.

This issue affected non-AES ciphers (legacy), such as Cast5, in Node 18+ or in browser.

Fix #1668 .

### DIFF
--- a/src/crypto/mode/cfb.js
+++ b/src/crypto/mode/cfb.js
@@ -119,7 +119,7 @@ export async function decrypt(algo, key, ciphertext, iv) {
     let j = 0;
     while (chunk ? ct.length >= block_size : ct.length) {
       const decblock = cipherfn.encrypt(blockp);
-      blockp = ct;
+      blockp = ct.subarray(0, block_size);
       for (i = 0; i < block_size; i++) {
         plaintext[j++] = blockp[i] ^ decblock[i];
       }


twiss (commented):
I think part of the issue here is that some of the block ciphers encrypt only one block (which this code seems to expect), while some ciphers encrypts everything. It would be good to make that uniform, and if all block ciphers encrypt only one block, then this code doesn't need to be changed.

If we do want to change it here, we should also change the `encrypt` function (even though it's less likely to be used).

larabr:
@twiss even if the block cipher were to decrypt only one block, passing the full ciphertext in input to them seems confusing. I'd leave more extensive changes to the underlying ciphers be done as part as #1654 .

The encryption function is fine as-is since the code successfully deals with multiple blocks, which should be better for performance

twiss (approved):

PR #1680, opened 2023-09-18
Add SHA-3 support

RyanHopkins7:
These changes should add SHA-3 support from noble hashes as I discussed with @twiss on Gitter. This change is in regards to #1663.
I've added tests in [test/crypto/hash/sha.js](https://github.com/openpgpjs/openpgpjs/compare/v6...RyanHopkins7:openpgpjs:v6#diff-3f233d0a793b93f25e1db46f9055772f3cfcd792ed974e753dc609730bc548fb), but I'm not sure if I should add more tests using a PGP key which uses SHA3 internally. What do you think? If you think that I should, where do you think I should add that test? Maybe test/crypto/validate.js?


src/crypto/hash/index.js
```diff
@@ -55,7 +56,9 @@ const hashFunctions = {
   sha256: nodeHash('sha256') || nobleHash(sha256, 'SHA-256'),
   sha384: nodeHash('sha384') || nobleHash(sha384, 'SHA-384'),
   sha512: nodeHash('sha512') || nobleHash(sha512, 'SHA-512'),
-  ripemd: nodeHash('ripemd160') || nobleHash(ripemd160)
+  ripemd: nodeHash('ripemd160') || nobleHash(ripemd160),
+  sha3_256: nobleHash(sha3_256),
+  sha3_512: nobleHash(sha3_512)
```
larabr, line 61: SHA3 should also be natively implemented by NodeCrypto:
```suggestion
  sha3_256: nodeHash('sha3-256') || nobleHash(sha3_256),
  sha3_512: nodeHash('sha3-512') || nobleHash(sha3_512)
```

larabr (changes requested):
Thanks for working on this! 🙂 
I just left a small comment about leveraging NodeCrypto as well.

> I've added tests but I'm not sure if I should add more tests using a PGP key which uses SHA3 internally. What do you think? If you think that I should, where do you think I should add that test?

I would add a test checking that
- a key using SHA3 internal signatures can be read using `openpgp.readKey` (e.g. key from https://github.com/openpgpjs/openpgpjs/issues/1663#issue-1856832910 )
- signing with the key (using `openpgp.sign`) generates a sha3 signature
- the newly generated SHA3 signature is verified (by `openpgp.verify`)

This could go under the [unit tests for signing](https://github.com/openpgpjs/openpgpjs/blob/v6/test/general/openpgp.js#L1806) in test/general/openpgpjs.js.

Since SHA3 is just another hash algo, I don't think more ad-hoc tests are needed: everything is expected to work as long as the underlying hash function is correct.

RyanHopkins7:
Sounds great, thanks for your quick review!! I'll add the test(s) you suggested.

RyanHopkins7:
Hi, I've added a tests for `openpgp.readKey`, `openpgp.sign`, and `openpgp.verify` using the private key from https://github.com/openpgpjs/openpgpjs/issues/1663#issue-1856832910. Please let me know if that's sufficient or if you had any other suggestions!

test/general/openpgp.js
```diff
@@ -1821,6 +1846,24 @@ aOU=
...
+      const config = { minRSABits: 1024 };
+
+      const cleartextMessage = await openpgp.sign({ message, signingKeys: privKey, config, format: 'armored' });
+      const parsedArmored = await openpgp.readCleartextMessage({ cleartextMessage });
+      expect(parsedArmored.text).to.equal(text);
+      expect(parsedArmored.signature.packets.filterByTag(openpgp.enums.packet.signature)).to.have.length(1);
+
+      const objectMessage = await openpgp.sign({ message, signingKeys: privKey, config, format: 'object' });
+      expect(objectMessage.signature.packets.filterByTag(openpgp.enums.packet.signature)).to.have.length(1);
+      const verified = await openpgp.verify({ message: objectMessage, verificationKeys: pubKey, expectSigned: true, config });
```
larabr, line 1863: `parsedArmored` is an `CleartextMessage` object just like the one returned when requesting `format: 'object'` , and you can pass it directly to verification:

```suggestion
      const verified = await openpgp.verify({ message: parsedArmored, verificationKeys: pubKey, expectSigned: true, config });
```

test/general/openpgp.js
```diff
@@ -1821,6 +1846,24 @@ aOU=
       })).to.be.rejectedWith(/No signing keys provided/);
     });
 
+    it('Signing with key which uses sha3 should generate a valid sha3 signature', async function() {
+      const privKey = await openpgp.readKey({ armoredKey: priv_key_sha3 });
+      const pubKey = privKey.toPublic();
+      const text = 'Hello, world.';
+      const message = await openpgp.createCleartextMessage({ text });
+      const config = { minRSABits: 1024 };
```
larabr, line 1854:
```suggestion
```

This is only necessary when using weak RSA keys, which some other tests do, but `priv_key_sha3` is fine 🙂

test/general/openpgp.js
```diff
@@ -1821,6 +1846,24 @@ aOU=
...
+
+      const cleartextMessage = await openpgp.sign({ message, signingKeys: privKey, config, format: 'armored' });
+      const parsedArmored = await openpgp.readCleartextMessage({ cleartextMessage });
+      expect(parsedArmored.text).to.equal(text);
+      expect(parsedArmored.signature.packets.filterByTag(openpgp.enums.packet.signature)).to.have.length(1);
```
larabr, line 1859: Could you add a check that this filtered signature packet has `.hashAlgorithm` prop that equals `openpgp.enums.hash.sha3_x` (not sure what sha3 size it uses)?

test/general/openpgp.js
```diff
@@ -928,6 +947,12 @@ export default () => describe('OpenPGP.js public api tests', function() {
     it('readPrivateKeys should throw on armored public keys', async function() {
       await expect(openpgp.readPrivateKeys({ armoredKeys: twoPublicKeys })).to.be.rejectedWith(/Armored text not of type private key/);
     });
+
+    it('readKey should work for a key that uses SHA3 internal signatures', async function() {
+      const key = await openpgp.readKey({ armoredKey: priv_key_sha3 });
+      expect(key.isPrivate()).to.be.true;
+      expect(key.isDecrypted()).to.be.true;
+    });
```
larabr, line 955:
```suggestion
```

No need to test this in isolation, because thanks to the forward compatibility changes included in v5.10.0, reading the key should work fine even without SHA3 support.
It's only when trying to use the key that errors would be triggered .
(At least, this should be true in principle, but in practice it's not happening due to a small oversight in the v6 signature implementation, which I'll fix shortly 😇 ).

test/general/openpgp.js
```diff
@@ -1821,6 +1846,24 @@ aOU=
...
+    it('Signing with key which uses sha3 should generate a valid sha3 signature', async function() {
+      const privKey = await openpgp.readKey({ armoredKey: priv_key_sha3 });
+      const pubKey = privKey.toPublic();
+      const text = 'Hello, world.';
+      const message = await openpgp.createCleartextMessage({ text });
+      const config = { minRSABits: 1024 };
+
+      const cleartextMessage = await openpgp.sign({ message, signingKeys: privKey, config, format: 'armored' });
+      const parsedArmored = await openpgp.readCleartextMessage({ cleartextMessage });
+      expect(parsedArmored.text).to.equal(text);
```
larabr, line 1858:
```suggestion
```
The check at line 1864 covers this as well 🙂

test/general/openpgp.js
RyanHopkins7: Yes! Thanks so much, I thought about this but I wasn't sure exactly how to check the alg in the signature. :)

RyanHopkins7:
Thanks a bunch for your feedback, explanations, and patience. :)
I believe that the test should be good now-- let me know if there's anything else though!

test/general/openpgp.js
```diff
@@ -14,6 +14,7 @@ import keyIDType from '../../src/type/keyid.js';
 import { getPreferredCipherSuite } from '../../src/key';
 
 import * as input from './testInputs.js';
+import enums from '../../src/enums.js';
```
larabr, line 17:
```suggestion
```
Better to access `openpgp.enums` directly 👍

test/general/openpgp.js
```diff
@@ -1821,6 +1841,23 @@ aOU=
...
+      const pubKey = privKey.toPublic();
+      const text = 'Hello, world.';
+      const message = await openpgp.createCleartextMessage({ text });
+
+      const cleartextMessage = await openpgp.sign({ message, signingKeys: privKey, format: 'armored' });
+      const parsedArmored = await openpgp.readCleartextMessage({ cleartextMessage });
+      expect(parsedArmored.signature.packets.filterByTag(openpgp.enums.packet.signature)).to.have.length(1);
+      expect(
+        parsedArmored.signature.packets.filterByTag(openpgp.enums.packet.signature)[0].hashAlgorithm
+      ).to.equal(enums.hash.sha3_512);
```
larabr, line 1855:
```suggestion
      ).to.equal(openpgp.enums.hash.sha3_512);
```

larabr (commented):
Thanks for the prompt changes! Almost there, just noted a couple of minor things 🙂 (in particular, see https://github.com/openpgpjs/openpgpjs/pull/1680#discussion_r1331772081 ).

src/crypto/hash/index.js
RyanHopkins7: Oops, so sorry about that! I should be more careful with force push. Thanks!!

larabr (approved):
Thanks again! Looks good to me, let's see if @twiss has something to add 🙂

twiss (approved):
Thanks for working on this! It looks good to me. Just as a note, we may want to consider adding these algorithms to the preferred hash algorithms (in key/factory.js) as well, but before we do that we should do some benchmarks on which algorithms are fastest, to decide the order. So IMO we can merge this as-is and do that later :)

PR #1691, opened 2023-10-11
Fix stream closure when using Node's stream.pipeline

### LINKED ISSUE #1512: Encrypted stream stays readable once all data read
- OpenPGP.js version: 5.1.0
- Affected platform (Browser or Node.js version): v14.15.5

I was using the pre-release version of OpenPGP, v5.0.0-1, to encrypt a stream and it worked perfectly. After upgrading to v5.1.0, the promisified pipeline we use never resolved, or threw an error. The original code can be seen here: https://github.com/DEFRA/rod-licensing/blob/5f658b4081207d27ef60017fc76785b6f7fd29b5/packages/fulfilment-job/src/staging/deliver-fulfilment-files.js#L75.

After some investigation, I could see that the encrypted data stream does emit an end event, however if you use stream.finished then it doesn't return. Only if readable is set to false will stream.finished be fired. So, I think this is causing a problem when the encrypted stream is used in conjunction with node's built-in stream.pipeline function.

The following will write out 'encrypted finished writing without failure' and 'encrypted stream end event', but not 'file written':
```
import { pipeline, finished } from 'stream'
import openpgp from 'openpgp'
import fs from 'fs'

const publicKey = await openpgp.readKey({ armoredKey: fs.readFileSync('paragon-public.asc', { encoding: 'utf8' }) });
const now = new Date();
const encryptedFileName = `files_to_encrypt/EAFF${now.getFullYear()}${(now.getMonth() + 1).toString().padStart(2, '0')}${(now.getDate()).toString().padStart(2, '0')}0001.json.enc`;
const fileToEncrypt = 'files_to_encrypt/sample_data.json';
const stream = fs.createReadStream(fileToEncrypt);
const encrypted = await openpgp.encrypt({
  message: await openpgp.createMessage({ text: stream }),
  encryptionKeys: publicKey
});
const ws = fs.createWriteStream(`${fileToEncrypt}.enc`);
pipeline(encrypted, ws, (err, val) => {
  console.log('file written', err, val)
});
finished(encrypted, {readable: false}, (err) => {
  if (!err) {
    console.log('encrypted finished writing without failure')
  } else {
    console.log('uh oh:', err)
  }
})
encrypted.on('end', () => {
  console.log('encrypted steam end event');
});
```

If `finished(encrypted, {readable: false}, (err) => {` is changed to `finished(encrypted, (err) => {`, then the callback provided to finished is never called. 

To me, this seems like a bug, as it makes it impossible to use an encrypted file stream with pipeline, unless some sort of failsafe is put in using either the end event or using the finish function in the way I've described.

larabr:
Fix #1512

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "@openpgp/pako": "^1.0.12",
     "@openpgp/seek-bzip": "^1.0.5-git",
     "@openpgp/tweetnacl": "^1.0.3",
-    "@openpgp/web-stream-tools": "^0.0.11",
+    "@openpgp/web-stream-tools": "^0.0.11-patch-0",
     "@rollup/plugin-commonjs": "^11.1.0",
     "@rollup/plugin-node-resolve": "^7.1.3",
     "@rollup/plugin-replace": "^2.3.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@openpgp/pako": "^1.0.12",
         "@openpgp/seek-bzip": "^1.0.5-git",
         "@openpgp/tweetnacl": "^1.0.3",
-        "@openpgp/web-stream-tools": "^0.0.11",
+        "@openpgp/web-stream-tools": "^0.0.11-patch-0",
         "@rollup/plugin-commonjs": "^11.1.0",
         "@rollup/plugin-node-resolve": "^7.1.3",
         "@rollup/plugin-replace": "^2.3.2",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
     "@openpgp/pako": "^1.0.12",
     "@openpgp/seek-bzip": "^1.0.5-git",
     "@openpgp/tweetnacl": "^1.0.3",
-    "@openpgp/web-stream-tools": "^0.0.11",
+    "@openpgp/web-stream-tools": "^0.0.11-patch-0",
```
twiss, line 63: In this particular case it's probably better to specify the exact version as semver considers 0.0.11 > 0.0.11-patch-0:
```suggestion
    "@openpgp/web-stream-tools": "0.0.11-patch-0",
```

PR #1711, opened 2023-12-13
Add contributing guides and improve documentation

AlbaHerrerias:
This PR includes several contribution guides, implements a code of conduct and adds some improvements in documentation.

Thank you

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,38 +79,14 @@ Additionally, please try to follow the style of the commit messages on our [main
 
 ## Testing
 
-We have multiple types of tests. They are located in the root [`/test`](https://github.com/openpgpjs/openpgpjs/tree/main/test) folder.
-Before you can run the tests you need to build the project using `npm run build` and this will bundle the project and the tests for it be used in our different testing environments.
+Tests. They are located in the root [`/test`](https://github.com/openpgpjs/openpgpjs/tree/main/test) folder.
+Before running the tests, run `npm run build` to build OpenPGP.js if you've made changes since running `npm install`.
 
 **Unit Tests:**
 To run the tests in Node.js, run `npm test`.
 
-**Browser Tests:**
-For debugging browser errors, you can run `npm start` and open [`http://localhost:8080/test/unittests.html`](http://localhost:8080/test/unittests.html) in a browser, or run one the following commands:
+To run the tests in a browser, run `npm run browsertest`.
 
-This will run the tests inside a browser environment:
-
-```sh
-   npm run browsertest
-```
-
-You can run the tests for multiple headless browsers using KarmaJs:
-
-` npm run test-browser`
-
-If you wanna start the tests using KarmaJs and BrowserStack:
-`npm run test-browserstack`
-
-**_NOTE_**: You need to configure you env with `BROWSER_STACK_ACCESS_KEY` and `BROWSERSTACK_USERNAME` to be able to run the BrowserStack tests.
-
-
-**Performance:** 
-To avoid performance regressions, we measure time and memory usage with the following testing suites (in `/benchmarks`):
-
-```sh
-npm run benchmark-time
-npm run benchmark-memory-usage
-```
 
 **Typescript definitions**
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,192 @@
+# Contributing to OpenPGP.js
+
+Please take a moment to review this document in order to make the contribution process easy and effective for everyone involved.
+
+Following these guidelines helps to communicate that you respect the time of the developers managing and developing this open source project. In return, they should reciprocate that respect in addressing your issue, assessing changes, and helping you finalize your pull requests.
+
+As for everything else in the project, the contributions are governed by our [Code of Conduct](https://github.com/openpgpjs/openpgpjs/blob/main/CODE_OF_CONDUCT.md).
+
+- [Contributing to OpenPGP.js](#contributing-to-openpgpjs)
+  - [Getting started](#getting-started)
+    - [Decide what you want to work on](#decide-what-you-want-to-work-on)
+    - [Notify your interest](#notify-your-interest)
+    - [Setting up the project in your local machine](#setting-up-the-project-in-your-local-machine)
+  - [Coding conventions](#coding-conventions)
+  - [Commit conventions](#commit-conventions)
+  - [Testing](#testing)
+  - [Pull requests](#pull-requests)
+    - [I have submitted my PR, what are the next steps?](#i-have-submitted-my-pr-what-are-the-next-steps)
+  - [Bug reports](#bug-reports)
+  - [Communication](#communication)
+  - [Thank you](#thank-you)
+
+## Getting started
+
+### Decide what you want to work on
+
+If you are looking for something to work on, we try to maintain a list of issues that should be suitable for first time contributions, they can be found tagged [`good-first-issue`](https://github.com/openpgpjs/openpgpjs/labels/good-first-issue). You can also look through our issues and pick some you like. If you're still unsure, please [reach out](#communication) and we would help you to the best of our abilities.
+
+### Notify your interest
+
+Please let us know you want to work on it so we can avoid multiple people working on the same issue. Ideally, notify us via our [Gitter](https://gitter.im/openpgpjs/openpgpjs), but expressing your interest in the issue itself is also ok.
+
+**Please ask first** before embarking on any significant pull request (e.g. implementing features, refactoring code), otherwise you risk spending a lot of time working on something that the project's developers might not want to merge into the project.
+
+### Setting up the project in your local machine
+
+1. [Fork](https://docs.github.com/en/get-started/quickstart/fork-a-repo) the project, clone your fork, and configure the remotes:
+
+   ```bash
+   # Clone your fork of the repo into the current directory
+   git clone https://github.com/<your-username>/<repo-name>
+
+   # Navigate to the newly cloned directory
+   cd <repo-name>
+
+   # Assign the original repo to a remote called "upstream"
+   git remote add upstream https://github.com/openpgpjs/<repo-name>
+   ```
+
+2. Install dependencies with your preferred package manager:
+
+   ```bash
+   npm install
+   ```
+
+3. Create a new topic branch (off the main project development branch) to contain your feature, change, or fix:
+
+   ```bash
+   git checkout -b <topic-branch-name>
+   ```
+
+4. Follow our [coding conventions](#coding-conventions) while developing your changes. You can also run `npm run lint` to match our requirements.
+
+5. Write clear and meaningful git commit messages. Please follow our [commit message conventions](#commit-conventions) while committing to your branch.
+
+6. Make sure to update or add to the tests when appropriate. Run the appropriate testing suites to check that all tests pass after you've made changes. You can read about our different types of tests in our [Testing](#testing) section.
+
+7. If you added or changed a feature, make sure to document it accordingly in the [README.md](https://github.com/openpgpjs/openpgpjs/blob/main/README.md) file.
+
+## Coding conventions
+
+We ensure code consistency through our [ESLint config](https://github.com/openpgpjs/openpgpjs/blob/main/.eslintrc.js). You can run `npm run lint` to lint your changes. When in doubt, follow the style of the surrounding code.
+
+## Commit conventions
+
+We follow the seven rules of a great Git commit message, which is extensively described in [this blog post](https://cbea.ms/git-commit/), but in short the principles are:
+
+1. Separate subject from body with a blank line
+2. Limit the subject line to 50 characters
+3. Capitalize the subject line
+4. Do not end the subject line with a period
+5. Use the imperative mood in the subject line
+6. Wrap the body at 72 characters
+7. Use the body to explain what and why vs. how
+
+You can also take a look at our [main branch commits](https://github.com/openpgpjs/openpgpjs/commits/main).
+
+## Testing
+
+We have multiple types of tests. They are located in the root [`/test`](https://github.com/openpgpjs/openpgpjs/tree/main/test) folder.
+Before you can run the tests you need to build the project using `npm run build` and this will bundle the project and the tests for it be used in our different testing environments.
+
+**Unit Tests:**
+Unit tests are handled by mocha and run inside node, you can run them using this command:
+
+```sh
+npm run test
+```
+
+**Browser Tests:**
+For debugging browser errors, you can run `npm start` and open [`http://localhost:8080/test/unittests.html`](http://localhost:8080/test/unittests.html) in a browser, or run one the following commands:
+
+This will run the tests inside a browser environment:
+
+```sh
+   npm run browsertest
+```
+
+You can run the tests for multiple headless browsers using KarmaJs:
+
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,192 @@
...
+```sh
+npm run benchmark-time
+npm run benchmark-memory-usage
+```
+
+**Typescript definitions**
+
+```sh
+npm run test-type-definitions
+```
```
twiss, line 131:
```suggestion
**Typescript definitions**

When changing TypeScript definitions, update [the definitions.ts test file](https://github.com/openpgpjs/openpgpjs/blob/main/test/typescript/definitions.ts) accordingly.

Then, run `npm run test-type-definitions` to verify that it still type-checks. (This is also automatically done for every PR.)
```

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+Support requests, eg. issues that are not bugs and exclusively concern the usage of the code as it is, will be converted in to a discussion and closed.
+
+Here is [an example](https://github.com/openpgpjs/openpgpjs/issues/1661) of this occuring
+
+If you find yourself doing this a lot, your `.github/ISSUE_TEMPLATE` needs improvement. The template should be effectively guiding people in need of support to the discussions and away from the issues.
+
+If you realise that the question is a duplicate of an existing discussion, and you can easily find that discussion, it’s worth linking to the existing one instead of creating a new one.
+
+If questions are frequently asked, they might be a frequently asked question! Consider adding the solution to the documentation, Wiki or FAQ, whichever applies and is easily discoverable.
+
```
twiss, line 67: These should go in Discussions instead. Also, the link 404s, which I guess is good (it might've been moved to Discussions :upside_down_face:)
```suggestion
```

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+#### `Bug` for things that don’t work as intended
+
+1. Assuming this is not or no longer a `Need more info` issue, attempt to reproduce the bug
+   1. Search for duplicates, if one exists, point the current issue there, and close the new one
+   2. If you can reproduce it and there are no duplicates, prioritize it and add the `Bug` label
+      1. If you need more info before this issue becomes actionable despite being able to reproduce it, prod the author and add a `Need more info` label
+   3. If you can’t reproduce it, contact the author with the reason, and add a `Need more info` to the `Bug` label
+      1. If the issue cannot be reproduced by anyone other than the author, agree with them that the issue should be closed
+
+As with regular `Need more info` issues, `Need more info` + `Bug` issues are also subject to the [Abandoned/Stale](#abandonedstale) rule above, and should be closed after a set period of time.
```
twiss, line 93: I would move this one up, as being the more "default" label an issue should have (unless there's something wrong with it).

Also, we might want to list a `Feature` label, but say that feature requests should usually go in https://github.com/openpgpjs/openpgpjs/discussions/categories/ideas; they can then be converted into an issue if the maintainers agree.

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+
+#### Good First Issues and Help Wanted Issues
+
+To identify issues that are specifically groomed for new contributors, we use the `Help wanted` _and_ `Good first issue` labels. As to the difference between the two:
+
+- a `Good first issue` always also implies `Help wanted` from outside the project
+- but `Help wanted` does not necessarily mean it’s a `Good first issue`. Might be really tricky but none of the current maintainers has the resources to tackle it
```
twiss, line 100: The first and last sentences seem contradictory to me... I would go with the latter interpretation, and drop the promise that "Help wanted" issues are suitable for new contributors - they should just mean that the core contributors won't get to it, IMO.

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+
+##### Criteria for a `Help wanted` Issue
+
+- **Low Barrier to Entry:** It should be easy for new contributors. Documentation on how that type of change should be made should already exist.
+
```
twiss, line 105:
```suggestion
```

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+- **Clear Task:** The task is agreed upon and does not require further discussions in the community. Call out if that area of code is untested and requires new fixtures.
+
+- **Goldilocks priority:** The priority should not be so high that a core contributor should do it, but not too low that it isn't useful enough for a core contributor to spend time reviewing it, answering questions, helping get it into a release, etc.
+
+- **Up-To-Date:** Often these issues become obsolete and have already been completed, are no longer desired, no longer make sense, or have changed priority or difficulty.
```
twiss, line 110: This seems like a very pessimistic comment in this context. Obviously it might be true but it's not really an actionable comment; obviously at the time when it's labeled as such it will be up-to-date, so I would just remove this tbh.

And again, IMO the main point of this label should be that the core maintainers won't have time for it and would welcome external contributors, so perhaps I would just only say that?

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+
+##### Criteria for a `Good first issue`
+
+Items marked with the `Good first issue` label are intended for first-time contributors. It indicates that maintainers will _proactively_ keep an eye out for the corresponding pull requests and shepherd them along.
+
+New contributors should not be left to find an approver, ping for reviews, or identify that their build failed due to a flake. It is important to make new contributors feel welcome and valued. We should assure them that they will have an extra level of help with their first contribution.
+
+After a contributor has successfully completed one or two `Good first issue` items, they should be ready to move on to `Help wanted` items.
+
```
twiss, line 119:
```suggestion
```

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+Must be staffed and worked on either currently or very soon—ideally in time for the next release. Important, but wouldn't block a release.
+
+#### `priority/important-longterm`
+
+Important over the long term, but may not be currently staffed and/or may require multiple releases to complete. Wouldn't block a release.
+
+#### `priority/backlog`
+
+Would be nice, but no one has the resources to work on it. Good candidates for the `Help wanted` label.
+
```
twiss, line 179: We don't really prioritize issues in OpenPGP.js; because the priority of issues is specific to each application: what's important to Proton might not be important to someone else, and vice versa. So we don't really want to categorize whether any given issue is "objectively" important or not.

Also, issues are not staffed by OpenPGP.js directly, but rather applications and contributors decide what they work on.

```suggestion
```

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+### Step 3: Assign the Work
+
+It should be clear who is responsible for working on an issue once it has been triaged and deemed actionable. This also includes eventual reviews. Ideally don’t assign things to people who aren’t expecting it.
```
twiss, line 182: The above also means that this isn't always possible. If one of the maintainers will do it, then yes, but otherwise it's up to the applications and contributors to pick up the work, not for us to assign it.

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Step 4: Follow up
+
+- **If no PR is opened on an issue within a month**, a maintainer should contact the assignee and ask them to create a PR or unassign themselves
```
twiss, line 186: This also seems overly rigid. Obviously, if someone else comes along and wants to work on the issue we can re-assign, but otherwise unassigning doesn't seem strictly necessary?

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+Triage can happen asynchronously and continuously, or in regularly scheduled meetings.
+
+**Benefits of triaging are:**
+- Maintaining project momentum
+- Engagement of contributors by increasing responsiveness
+- Prevents piling up of issues and work
+- Prevents issues and the project itself becoming stale
+- Provides structure and an opportunity for collaborative decision-making with regards to the project trajectory and roadmap
+
+It can be beneficial to have triage be part of already existing regular meetings, and if the volume of incoming issues requires it, have a regular, public triage meeting. If there is a regular meeting, make sure it is added here and in the [CONTRIBUTING.md](/CONTRIBUTING.md). Also consider adding it to the [NEW-CONTRIBUTORS.md](/NEW-CONTRIBUTORS.MD) and explicitly invite new and prospective contributors to listen in or participate. This will give interested people a plannable, social opportunity to meet the maintainers and gain insight into the project, with both a low barrier to entry for the new contributors and a low amount of additional effort for the maintainers.
```
twiss, line 50: This would seem like a reasonable suggestion if we had ore maintainers, but at the moment Lara is doing most of the work, so then having a meeting seems overkill (although we do have meetings, but not specifically for this).

Also, even if it is/becomes a good suggestion, it would make more sense to me for this document to describe the current practice? So I would drop this for now.

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+Would be nice, but no one has the resources to work on it. Good candidates for the `Help wanted` label.
+
+### Step 3: Assign the Work
+
+It should be clear who is responsible for working on an issue once it has been triaged and deemed actionable. This also includes eventual reviews. Ideally don’t assign things to people who aren’t expecting it.
+
+### Step 4: Follow up
+
+- **If no PR is opened on an issue within a month**, a maintainer should contact the assignee and ask them to create a PR or unassign themselves
+- **If a PR is ready for review**, use your triage meeting to find someone to review it within a reasonable amount of time. If you cannot manage a review soon, explain that to the contributor so they’re not left hanging and know what to expect.
```
twiss, line 187:
```suggestion
- **If a PR is ready for review**, find someone to review it within a reasonable amount of time. If you cannot manage a review soon, explain that to the contributor so they’re not left hanging and know what to expect.
```

MAINTAINERS.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Release process
+
+We keep it simple. We use npm’s `version` command as following:
+
+```sh
+npm version {major,minor,patch}
+```
+
+Which will trigger the `preversion`, `version` and `postversion` scripts specified in the `package.json`. These will handle the testing, building, publish to npm and pushing to GitHub.
```
twiss, line 197:
```suggestion
(depending on whether a major, minor or patch release is to be created.) This command will run tests, create a build, publish it to npm, and push to GitHub. 
```

twiss (commented):
Hey :wave: Happy new year, and thanks for the added docs! Some comments below :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,192 @@
...
+4. Do not end the subject line with a period
+5. Use the imperative mood in the subject line
+6. Wrap the body at 72 characters
+7. Use the body to explain what and why vs. how
+
+You can also take a look at our [main branch commits](https://github.com/openpgpjs/openpgpjs/commits/main).
+
+## Testing
+
+We have multiple types of tests. They are located in the root [`/test`](https://github.com/openpgpjs/openpgpjs/tree/main/test) folder.
```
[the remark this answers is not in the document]
larabr: I think they are referring to there being benchmark tests (memory & time), the TS ones, and the to-be-merged fuzzing ones. IMO the original version makes sense.

CONTRIBUTING.md
```diff
@@ -0,0 +1,192 @@
...
+Before you can run the tests you need to build the project using `npm run build` and this will bundle the project and the tests for it be used in our different testing environments.
```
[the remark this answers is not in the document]
larabr: To simplify things, IMO it's fine to leave the redundant suggestion to re-run `npm run build`. Then also add `npm run build-test`. That would should all cases?

CONTRIBUTING.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+You can run the tests for multiple headless browsers using KarmaJs:
+
+` npm run test-browser`
+
+If you wanna start the tests using KarmaJs and BrowserStack:
+`npm run test-browserstack`
+
+**_NOTE_**: You need to configure you env with `BROWSER_STACK_ACCESS_KEY` and `BROWSERSTACK_USERNAME` to be able to run the BrowserStack tests.
```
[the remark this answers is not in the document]
larabr: I would spell out the various options tbh, it's good to have a reference.

Running the tests with Karma (`npm run test-browser`) is considerably faster than using `browsertest`, and automatically targets multiple browsers.
Similarly, it can be useful to trigger the BrowserStack tests locally for debugging -- I wouldn't implicitly encourage using the CI for that.

CONTRIBUTING.md
```diff
@@ -0,0 +1,192 @@
...
+
+
+**Performance:** 
+To avoid performance regressions, we measure time and memory usage with the following testing suites (in `/benchmarks`):
+
+```sh
+npm run benchmark-time
+npm run benchmark-memory-usage
+```
```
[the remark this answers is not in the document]
larabr: As mentioned above, I think it makes sense to add documentation about all the types of tests we have. So instead of removing this info, I'd explicitly write that this one is run in the CI, where it also compares the performance to the main branch 👍

MAINTAINERS.md
larabr: IMO this could fit here as part of the "guidelines" idea mentioned in https://github.com/openpgpjs/openpgpjs/pull/1711#discussion_r1444374718 🙂 

In fact, we do have some Wiki entries we'd been wanting to write (e.g. about streaming), and a I think we have a few candidate questions for a FAQ as well (e.g. session key decryption errors).

MAINTAINERS.md
larabr: I think I understand what you mean to say, but since this is a public forum, I'd like to clarify that we do security and priority assessment of issues, and immediately tackle those with high priority.
In this sense, "critical-urgent" and "important-soon" will typically be fixed by us [maintainers] in a matter of days, without the need to add labels (esp. if there is some confidentiality involved in the case of security issues).

I feel your reply is primarily applicable to "feature requests", rather than bugs/issues in the proper sense 🙂

MAINTAINERS.md
twiss: From the current text it's not really clear which option is current practice and which is a suggestion, though. And I think for OpenPGP.js contributors and future maintainers it would be more useful to know the actual process being followed here, rather than an aspirational document - we can have that elsewhere :) Or just refer to https://github.com/kubernetes/community/blob/master/contributors/guide/issue-triage.md (which is already linked) for future ideas in case the project grows significantly.

CONTRIBUTING.md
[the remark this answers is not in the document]
twiss: The following two subheaders are "Unit Tests" and "Browser Tests", so this sentence makes it sound like those are different types of tests, which is not the case.

CONTRIBUTING.md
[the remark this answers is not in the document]
twiss: `npm run build-test` is part of `npm run browsertest`, so shouldn't be necessary. `npm run build` can be useful to run after making changes (and before running the tests), but then I'd rather say that, e.g.:

```suggestion
Before running the tests, run `npm run build` to build OpenPGP.js if you've made changes since running `npm install`.
```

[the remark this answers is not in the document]
larabr: > `npm run build-test` is part of `npm run browsertest`

but it's not part of `npm run test-browser` and `-browserstack`. I just think it's less confusing to present a consistent procedure; it doesn't have side-effects anyway.

PR #1713, opened 2024-01-11
[v6] Skip key validation for keys encrypted with non-legacy AEAD mechanism

larabr:
Follow up to: #1116 .

The public key material integrity is guaranteed by the new encryption mechanism,
hence `.validate()` does not need to run further checks.
This results in a performance speed-up.

TODO:

- [x] merge after #1672

### DIFF
--- a/src/packet/secret_key.js
+++ b/src/packet/secret_key.js
@@ -83,6 +83,13 @@ class SecretKeyPacket extends PublicKeyPacket {
      * @type {Object}
      */
     this.privateParams = null;
+    /**
+     * `true` for keys whose integrity is already confirmed, based on
+     * the AEAD encryption mechanism
+     * @type {Boolean}
+     * @private
+     */
+    this.usedModernAEAD = null;
   }
 
   // 5.5.3.  Secret-Key Packet Formats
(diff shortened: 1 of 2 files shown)


src/packet/secret_key.js
```diff
@@ -83,6 +83,13 @@ class SecretKeyPacket extends PublicKeyPacket {
      * @type {Object}
      */
     this.privateParams = null;
+    /**
+     * `true` for keys whose integrity is already confirmed, based on
+     * the AEAD encryption mechanism
+     * @type {Boolean}
+     * @private
+     */
+    this.skipValidation = null;
```
twiss, line 92: Bit nitpicky but this naming sounds a bit like a config option rather than a property of the key, I would name it something like `integrityChecked` or `usedModernAEAD`; that way below it's more obvious why it's safe to check the validation.

src/packet/secret_key.js
```diff
@@ -493,6 +506,10 @@ class SecretKeyPacket extends PublicKeyPacket {
       throw new Error('Key is not decrypted');
     }
 
+    if (this.skipValidation) {
```
twiss, line 509: Might want a comment here as well about why it's safe to skip, e.g.
```suggestion
    if (this.integrityChecked) {
      // The public key's integrity was already verified, based on the AEAD encryption mechanism.
```

src/packet/secret_key.js
```diff
@@ -493,6 +506,10 @@ class SecretKeyPacket extends PublicKeyPacket {
...
+      return true;
```
twiss, line 510: The function normally doesn't return a bool so we shouldn't here either.
```suggestion
      return;
```

twiss (approved):

PR #1719, opened 2024-02-01
[v6] Ensure primary key meets strength and algo requirements when encrypting/verifying/signing using subkeys

### LINKED ISSUE #1608: Reject insecure (RSA1024) primary key when verifying subkeys

larabr:
Fix #1608: the requirements of `config.minRSABits`, `rejectPublicKeyAlgorithms` and `rejectCurves`
are now applied to the primary key, aside from the selected subkey. This is a breaking change.

The motivation is that the subkeys are certified by the primary key, but if the latter is
weak, arbitrary subkeys could potentially be added.

Note that the change does not affect decryption, to allow decrypting older messages.

### DIFF
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -257,8 +257,8 @@ class Key {
    * @async
    */
   async getSigningKey(keyID = null, date = new Date(), userID = {}, config = defaultConfig) {
-    const primaryKey = this.keyPacket;
     await this.verifyPrimaryKey(date, userID, config);
+    const primaryKey = this.keyPacket;
     try {
       helper.checkKeyRequirements(primaryKey, config);
     } catch (err) {
@@ -316,8 +316,8 @@ class Key {
    * @async
    */
   async getEncryptionKey(keyID, date = new Date(), userID = {}, config = defaultConfig) {
-    const primaryKey = this.keyPacket;
     await this.verifyPrimaryKey(date, userID, config);
+    const primaryKey = this.keyPacket;
     try {
       helper.checkKeyRequirements(primaryKey, config);
     } catch (err) {
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -257,8 +257,13 @@ class Key {
    * @async
    */
   async getSigningKey(keyID = null, date = new Date(), userID = {}, config = defaultConfig) {
-    await this.verifyPrimaryKey(date, userID, config);
     const primaryKey = this.keyPacket;
+    await this.verifyPrimaryKey(date, userID, config);
+    try {
+      helper.checkKeyRequirements(primaryKey, config);
+    } catch (err) {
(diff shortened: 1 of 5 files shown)


src/key/key.js
```diff
@@ -257,8 +257,13 @@ class Key {
    * @async
    */
   async getSigningKey(keyID = null, date = new Date(), userID = {}, config = defaultConfig) {
-    await this.verifyPrimaryKey(date, userID, config);
     const primaryKey = this.keyPacket;
+    try {
+      await this.verifyPrimaryKey(date, userID, config);
```
twiss, line 262: I wouldn't wrap this function's errors, since all of them already mention the primary key, so it seems a bit redundant

src/key/key.js
```diff
@@ -257,8 +257,13 @@ class Key {
...
+      helper.checkKeyRequirements(primaryKey, config);
+    } catch (err) {
+      throw util.wrapError('Error verifying primary key', err);
```
twiss, line 265:
```suggestion
      throw util.wrapError('Could not verify primary key', err);
```
(since this error will most likely be wrapped again, it sounds a bit stilted otherwise)

src/key/key.js
```diff
@@ -311,8 +316,13 @@ class Key {
...
    */
   async getEncryptionKey(keyID, date = new Date(), userID = {}, config = defaultConfig) {
-    await this.verifyPrimaryKey(date, userID, config);
     const primaryKey = this.keyPacket;
+    try {
+      await this.verifyPrimaryKey(date, userID, config);
+      helper.checkKeyRequirements(primaryKey, config);
+    } catch (err) {
+      throw util.wrapError('Error verifying primary key', err);
+    }
```
twiss, line 325: (same here)

src/key/key.js
```diff
@@ -257,8 +257,13 @@ class Key {
    * @async
    */
   async getSigningKey(keyID = null, date = new Date(), userID = {}, config = defaultConfig) {
-    await this.verifyPrimaryKey(date, userID, config);
     const primaryKey = this.keyPacket;
+    await this.verifyPrimaryKey(date, userID, config);
```
twiss, line 261: And please minimize the diff :) i.e. revert the ordering (below as well)

twiss (approved):

PR #1722, opened 2024-02-05
[v6] Node: drop asn1.js dependency

larabr:
This also allows us to drop BN.js as a direct dependency, optimising the BigInteger-related chunking in the lightweight build.

asn1.js is a fairly large lib and was simply needed to handle DER encodings in some NodeCrypto operations.
This change replaces the dependency by moving to:
- JWT encoding for RSA (support added in Node v15)
- a much lighter dependency ([eckey-utils](https://github.com/tibetty/eckey-utils)) for ECDSA, where JWT cannot be used for now, as Node has yet to add decoding support for Brainpool curves.

### DIFF
--- a/src/crypto/public_key/elliptic/ecdsa.js
+++ b/src/crypto/public_key/elliptic/ecdsa.js
@@ -24,7 +24,7 @@ import enums from '../../../enums';
 import util from '../../../util';
 import { getRandomBytes } from '../../random';
 import hash from '../../hash';
-import { CurveWithOID, webCurves, privateToJWK, rawPublicToJWK, validateStandardParams } from './oid_curves';
+import { CurveWithOID, webCurves, privateToJWK, rawPublicToJWK, validateStandardParams, nodeCurves } from './oid_curves';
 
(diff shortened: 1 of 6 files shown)


src/crypto/public_key/elliptic/ecdsa.js
```diff
@@ -247,85 +242,45 @@ async function webVerify(curve, hashAlgo, { r, s }, message, publicKey) {
...
-    label: 'EC PRIVATE KEY'
-  });
 
-  return ECDSASignature.decode(sign.sign(key), 'der');
+  const signature = new Uint8Array(sign.sign({ key: derPrivateKey, format: 'der', type: 'sec1', dsaEncoding: 'ieee-p1363' }));
+  const len = curve.payloadSize;
+
+  return {
+    r: signature.slice(0, len),
+    s: signature.slice(len, len << 1)
```
twiss, line 263:
```suggestion
    r: signature.subarray(0, len),
    s: signature.subarray(len, len << 1)
```

twiss (approved):

PR #1723, opened 2024-02-05
[v6] Import legacy ciphers (CAST5, TwoFish, BlowFish, DES) only on demand

larabr:
This primarily affects the lightweight build, which will not include these (fairly large) modules in the main bundle file.


src/crypto/cipher/index.js
```diff
@@ -1,80 +1,73 @@
...
-import { DES, TripleDES } from './des';
-import CAST5 from './cast5';
-import TF from './twofish';
-import BF from './blowfish';
+export async function getCipher(algo) {
+  switch (algo) {
+    case enums.symmetric.aes128:
+    case enums.symmetric.aes192:
+    case enums.symmetric.aes256:
+      return aes(getCipherKeySize(algo));
```
twiss, line 9: It seems a bit wasteful to create the AES constructor and methods each time AES is requested. So I would prefer to still make constants for `aes{128,192,256}` as before, and then return one of them as needed here.

twiss (approved):

PR #1724, opened 2024-02-08
[v6] Use WebCrypto for AES-KW, drop AES_ECB dependency

larabr:
AES_CBC is now used instead of AES_ECB for single-block encryption. AES_CBC is implemented by WebCrypto, which can be used when encrypting (but not when decrypting, due to mandatory PKCS7 padding).

TODO:
- [x] Benchmarks
- [x] Point to asmcrypto release without AES_ECB
- [x] Merge after #1723


src/crypto/mode/ocb.js
```diff
@@ -242,7 +246,7 @@ async function OCB(cipher, key) {
       const tag = ciphertext.subarray(-tagLength);
       ciphertext = ciphertext.subarray(0, -tagLength);
 
-      const crypted = crypt(decipher, ciphertext, nonce, adata);
+      const crypted = await crypt(decipher, ciphertext, nonce, adata);
```
larabr, line 249:
```suggestion
      const crypted = crypt(decipher, ciphertext, nonce, adata);
```

twiss (approved):

PR #1726, opened 2024-02-16
README: replace "IETF proposal" with "4880bis proposal"

Artoria2e5:
According to LWN reporters, 4880bis is unlikely to become the next IETF RFC, as another proposal is getting more favor. As a result, it's probably a good idea to not vaguely call it "IETF proposal".

https://lwn.net/Articles/953797/

N.B. The new name is draft-koch-librepgp https://datatracker.ietf.org/doc/draft-koch-librepgp, but 4880bis is better known right now. It does have a shiny new website at https://librepgp.org/updates.html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ library to convert back and forth between them.
 
 * If the user's browser supports [native WebCrypto](https://caniuse.com/#feat=cryptography) via the `window.crypto.subtle` API, this will be used. Under Node.js the native [crypto module](https://nodejs.org/api/crypto.html#crypto_crypto) is used.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,7 +75,7 @@ library to convert back and forth between them.
...
+* The library implements the [4880bis proposal](https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-10) for authenticated encryption using native AES-EAX, OCB, or GCM. This makes symmetric encryption up to 30x faster on supported platforms. Since the specification has not been finalized and other OpenPGP implementations haven't adopted it yet, the feature is currently behind a flag. **Note: activating this setting can break compatibility with other OpenPGP implementations, and also with future versions of OpenPGP.js. Don't use it with messages you want to store on disk or in a database.** You can enable it by setting `openpgp.config.aeadProtect = true`.
```
larabr, line 78:
```suggestion
* The library implements the [RFC4880bis proposal](https://tools.ietf.org/html/draft-ietf-openpgp-rfc4880bis-10) for authenticated encryption using native AES-EAX, OCB, or GCM. This makes symmetric encryption up to 30x faster on supported platforms. Since the specification has not been finalized and other OpenPGP implementations haven't adopted it yet, the feature is currently behind a flag. **Note: activating this setting can break compatibility with other OpenPGP implementations, and also with future versions of OpenPGP.js. Don't use it with messages you want to store on disk or in a database.** You can enable it by setting `openpgp.config.aeadProtect = true`.
```

larabr:
Thank you!

Issue #1727, opened 2024-02-16
Node patch release 20.11.1 breaks some usages of decrypt.


adevine:
- OpenPGP.js version: 5.11.0
- Affected platform (Browser or Node.js version): Node.js 20.11.1

We hit this bug after an automatic upgrade in Google App Engine to the latest stable patch release of Node, 20.11.1, broke all of our PGP decryption. Reverting to Node.js 20.9.0 fixed the issue for us (but pretty sure the problem is just in the 20.11.1 patch, as explained below).

I believe this commit in this Node version, https://github.com/nodejs/node/commit/7079c062bb, which disables PKCS # 1, is what is responsible for this issue. We were getting an error thrown when calling `decrypt`, down in the `nodeDecrypt` method here: https://github.com/openpgpjs/openpgpjs/blob/main/src/crypto/public_key/rsa.js#L483 (one note about where that bug is thrown - note it swallows the bug thrown from Node's `privateDecrypt` method and throws a new Error, which made this a little harder to debug). Note I understand this was a security issue from Node's perspective, but I still think doing this in a patch release was a bad decision from Node for a change that is fundamentally not backwards compatible.

I still need to dig into our code to see why we were using PKCS # 1 in the first place, or if it's something that OpenPGP does by default. We were just calling `encrypt` for these files by:

```
encrypt({
    message: ...result of a createMessage call...,
    encryptionKeys: [...result of calling readKeys({ armoredKeys: ...some public keys... })...],
    signingKeys: ...result of calling readPrivateKey and then decryptKey...,
    format: 'binary',
});
```

I'm just need to check how our keys were originally created. But it doesn't look to me like we were explicitly calling out the use of PKCS  # 1 anywhere.

adevine:
Just wanted to add a link to https://github.com/orgs/nodejs/discussions/51784#discussioncomment-8494516 - Node maintainers are great, explained the policy is that critical security fixes can go in minor version updates, which means this change should have gone in a minor version and not a patch, but honestly that still would have likely hit us anyway. Just good to know the defined policy on this.

chazcross:
NodeJS 18.19.1 introduced the same change breaking OpenPGP.js decryption.

adevine:
Thanks @chazcross , I updated the title and also added a minimal test case to the description.

sharmankita:
[...]

PR #1728, opened 2024-02-19
Use JS fallback code for RSA decryption on Node if PKCS#1 is not supported

larabr:
Fix for #1727, necessary as Node v18.19.1, 20.11.1 and 21.6.2 have disabled support for PKCS#1 decryption.

### DIFF
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -140,8 +140,15 @@ export async function encrypt(data, n, e) {
  * @async
  */
 export async function decrypt(data, n, e, d, p, q, u, randomPayload) {
-  if (util.getNodeCrypto()) {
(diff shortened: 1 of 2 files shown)


src/crypto/public_key/rsa.js
```diff
@@ -140,8 +140,12 @@ export async function encrypt(data, n, e) {
  * @async
  */
 export async function decrypt(data, n, e, d, p, q, u, randomPayload) {
-  if (util.getNodeCrypto()) {
-    return nodeDecrypt(data, n, e, d, p, q, u, randomPayload);
+  if (util.getNodeCrypto() && !randomPayload) {
```
twiss, line 143: This change seems redundant with the change in `nodeDecrypt` below.. is this one intentional?

twiss (approved):

PR #1734, opened 2024-03-21
[v6] Fix wrong serialization of PKESK v6 for x25519/x448

larabr:
The cleartext session key symmetric algorithm was accidentally included in the packet.
As a result, the generated messages could fail to parse and/or decrypt in other implementations.
The messages would still decrypt successfully in OpenPGP.js, due to a laax parsing procedure, which simply discarded the unused additional byte.

This only affects v6 PKESK packets, that are only generated starting from OpenPGP.js v6, if the keys were generated enabling `config.aeadProtect` (off by default).

### DIFF
--- a/src/packet/public_key_encrypted_session_key.js
+++ b/src/packet/public_key_encrypted_session_key.js
@@ -128,9 +128,12 @@ class PublicKeyEncryptedSessionKeyPacket {
     }
     this.publicKeyAlgorithm = bytes[offset++];
     this.encrypted = crypto.parseEncSessionKeyParams(this.publicKeyAlgorithm, bytes.subarray(offset));
-    if (this.version === 3 && (
-      this.publicKeyAlgorithm === enums.publicKey.x25519 || this.publicKeyAlgorithm === enums.publicKey.x448)) {
-      this.sessionKeyAlgorithm = enums.write(enums.symmetric, this.encrypted.C.algorithm);
+    if (this.publicKeyAlgorithm === enums.publicKey.x25519 || this.publicKeyAlgorithm === enums.publicKey.x448) {
+      if (this.version === 3) {
+        this.sessionKeyAlgorithm = enums.write(enums.symmetric, this.encrypted.C.algorithm);
+      } else if (this.encrypted.C.algorithm !== null) {
+        throw new Error('Unexpected cleartext symmetric algorithm');
+      }
     }
   }
 
--- a/src/packet/public_key_encrypted_session_key.js
+++ b/src/packet/public_key_encrypted_session_key.js
@@ -174,9 +174,12 @@ class PublicKeyEncryptedSessionKeyPacket {
    */
   async encrypt(key) {
     const algo = enums.write(enums.publicKey, this.publicKeyAlgorithm);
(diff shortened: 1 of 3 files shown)


src/packet/public_key_encrypted_session_key.js
```diff
@@ -174,9 +177,9 @@ class PublicKeyEncryptedSessionKeyPacket {
    */
   async encrypt(key) {
     const algo = enums.write(enums.publicKey, this.publicKeyAlgorithm);
-    const encoded = encodeSessionKey(this.version, algo, this.sessionKeyAlgorithm, this.sessionKey);
+    const { dataToEncrypt, cleartextSessionKeyAlgorithm } = encodeSessionKey(this.version, algo, this.sessionKeyAlgorithm, this.sessionKey);
     this.encrypted = await crypto.publicKeyEncrypt(
-      algo, this.sessionKeyAlgorithm, key.publicParams, encoded, key.getFingerprintBytes());
+      algo, cleartextSessionKeyAlgorithm, key.publicParams, dataToEncrypt, key.getFingerprintBytes());
```
twiss, line 182: Rather than tunneling this value through `encodeSessionKey`, we can do something like:
```js
    const sessionKeyAlgorithm = this.version === 3 ? this.sessionKeyAlgorithm : null;
    // No symmetric encryption algorithm identifier is passed to the public-key algorithm for a
    // v6 PKESK packet, as it is included in the v2 SEIPD packet.
    const encoded = encodeSessionKey(algo, sessionKeyAlgorithm, this.sessionKey);
    this.encrypted = await crypto.publicKeyEncrypt(
      algo, sessionKeyAlgorithm, key.publicParams, dataToEncrypt, key.getFingerprintBytes());
```
And then in `encodeSessionKey` you can just check whether `sessionKeyAlgorithm` was passed (instead of looking at the version), and the same thing in `publicKeyEncrypt` (for X25519).
That is more akin to how the spec writes it, and also seems simpler to me.

larabr: But `decodeSessionKey` also takes the version to decide how to process the result, so I think it's clearer to have the two functions aligned in terms of interface, and in the fact that they take care of the whole logic.

larabr: It's unused in v5 because we only support PKESK v3, but in v6 it's definitely used

twiss: Right, OK. But also there, we could use `decrypted.length % 2`, rather than checking the version. (Obviously we don't have to do that here and it's not urgent, but I think passing the version shouldn't be necessary)

larabr: Checking the length is brittle and it's part of what caused this bug to go undetected on parsing -- it is a necessary evil in the `crypto.publicKeyDecrypt` context as we want the module to be decoupled and unaware of packet versions, but in the PKESK module context we can and should avoid the hack.

I could integrated your suggested change and already pass a `null` cipher to encodeSessionKey on encryption, to avoid returning it from the helper -- but the encoding function is also used on decryption, and having to manually nullify the cipher algo in multiple places is problematic IMO, so I would still keep the version check.

twiss (approved):

PR #1737, opened 2024-03-26
[v6] Randomise v4 and v5 signatures via custom notation, add `config.nonDeterministicSignaturesViaNotation` to disable feature

larabr:
EdDSA is known to be vulnerable to fault attacks which can lead to secret key extraction if two signatures over the same data can be collected. Randomly occurring bitflips in specific parts of the computation might in principle result in vulnerable faulty signatures being generated.
To protect signatures generated using v4 and v5 keys from this possibility, we randomise each signature by adding a custom notation with a random value, functioning as a salt. 
For simplicity, we add the salt to all algos, not just EdDSA, as it may also serve as protection in case of weaknesses in the hash algo, potentially hindering e.g. some chosen-prefix attacks.
v6 signatures do not need to rely on this, as they are non-deterministic by design.

While this notation solution is interoperable, it will reveal that the signature has been generated using OpenPGP.js, which may not be desirable in some cases.
For this reason, the option `config.nonDeterministicSignaturesViaNotation` (defaulting to true) has been added to turn off the feature.


twiss:
Perhaps we can just do this for all algorithms, not just EdDSA? I believe Sequoia does so also as a precaution against vulnerabilities in the hash algorithm, for example. And, while that's unlikely to be an issue for SHA2 anytime soon, it'd simplify the code (and security analyses, in the sense that then all signature hashes are salted, regardless of algorithm and signature version), and doesn't cost most.

PR #1741, opened 2024-04-08
[v6] Add back armor checksum for non-v6 artifacts

larabr:
We need to include the checksum to workaround a GnuPG bug where data fails to be decoded if the base64 ends with no padding chars (=) (see https://dev.gnupg.org/T7071).
Pure v6 artifacts are unaffected and won't include the checksum, as mandated by the spec.

Breaking change:
`openpgp.armor` takes an additional `emitChecksum` argument (defaults to false).
NB: some types of data must not include the checksum, but compliance is left as responsibility of the caller: this function does not carry out any checks. Refer to the [crypto-refresh](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#section-6.1) for more details.

### DIFF
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -111,9 +111,9 @@ export class CleartextMessage {
    * @returns {String | ReadableStream<String>} ASCII armor.
    */
   armor(config = defaultConfig) {
+    const includesNonV6Signatures = this.signature.packets.some(packet => packet.version !== 6);
     // emit header if one of the signatures has a version not 6
-    const emitHeader = this.signature.packets.some(packet => packet.version !== 6);
-    const hash = emitHeader ?
+    const hash = includesNonV6Signatures ?
       Array.from(new Set(this.signature.packets.map(
         packet => enums.read(enums.hash, packet.hashAlgorithm).toUpperCase()
       ))).join() :
@@ -124,7 +124,9 @@ export class CleartextMessage {
       text: this.text,
       data: this.signature.packets.write()
     };
-    return armor(enums.armor.signed, body, undefined, undefined, undefined, config);
+
+    // An ASCII-armored sequence of Signature packets that only includes v6 Signature packets MUST NOT contain a CRC24 footer.
+    return armor(enums.armor.signed, body, undefined, undefined, undefined, includesNonV6Signatures, config);
   }
 }
 
--- a/src/cleartext.js
+++ b/src/cleartext.js
@@ -111,9 +111,9 @@ export class CleartextMessage {
    * @returns {String | ReadableStream<String>} ASCII armor.
    */
   armor(config = defaultConfig) {
-    const includesNonV6Signatures = this.signature.packets.some(packet => packet.version !== 6);
-    // emit header if one of the signatures has a version not 6
-    const hash = includesNonV6Signatures ?
+    // emit header and checksum if one of the signatures has a version not 6
+    const emitHeaderAndChecksum = this.signature.packets.some(packet => packet.version !== 6);
+    const hash = emitHeaderAndChecksum ?
       Array.from(new Set(this.signature.packets.map(
         packet => enums.read(enums.hash, packet.hashAlgorithm).toUpperCase()
       ))).join() :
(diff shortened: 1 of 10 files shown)


src/cleartext.js
```diff
@@ -111,9 +111,9 @@ export class CleartextMessage {
    * @returns {String | ReadableStream<String>} ASCII armor.
    */
   armor(config = defaultConfig) {
+    const includesNonV6Signatures = this.signature.packets.some(packet => packet.version !== 6);
     // emit header if one of the signatures has a version not 6
-    const emitHeader = this.signature.packets.some(packet => packet.version !== 6);
-    const hash = emitHeader ?
+    const hash = includesNonV6Signatures ?
```
twiss, line 116: Bit nitpicky but I think it's easier to understand the code if this variable says what it's used for rather than how it was computed, closer to how it was before:
```js
    // Emit header and checksum if one of the signatures has a version other than 6
    const emitHeaderAndChecksum = this.signature.packets.some(packet => packet.version !== 6);
    const hash = emitHeaderAndChecksum ?
```

src/key/key.js
```diff
@@ -612,7 +612,9 @@ class Key {
     const revocationSignature = await helper.getLatestValidSignature(this.revocationSignatures, this.keyPacket, enums.signature.keyRevocation, dataToVerify, date, config);
     const packetlist = new PacketList();
     packetlist.push(revocationSignature);
-    return armor(enums.armor.publicKey, packetlist.write(), null, null, 'This is a revocation certificate');
+    // An ASCII-armored Transferable Public Key packet sequence of a v6 key MUST NOT contain a CRC24 footer.
+    const v6Key = this.keyPacket.version === 6;
+    return armor(enums.armor.publicKey, packetlist.write(), null, null, 'This is a revocation certificate', !v6Key);
```
twiss, line 617: And similarly here, the code becomes more descriptive if you rename the variable to:
```suggestion
    // An ASCII-armored Transferable Public Key packet sequence of a v6 key MUST NOT contain a CRC24 footer.
    const emitChecksum = this.keyPacket.version !== 6;
    return armor(enums.armor.publicKey, packetlist.write(), null, null, 'This is a revocation certificate', emitChecksum);
```

And below as well.

src/encoding/armor.js
```diff
@@ -334,29 +334,38 @@ export function unarmor(input) {
...
  * @param {Integer} [partTotal]
  * @param {String} [customComment] - Additional comment to add to the armored string
+ * @param {Boolean} [withChecksum] - Whether to compute and include the CRC checksum
+ *  (NB: some types of data must not include it, but compliance is left as responsibility of the caller: this function does not carry out any checks)
+ * @param {Object} [config] - Full configuration, defaults to openpgp.config
  * @returns {String | ReadableStream<String>} Armored text.
  * @static
  */
-export function armor(messageType, body, partIndex, partTotal, customComment, config = defaultConfig) {
+export function armor(messageType, body, partIndex, partTotal, customComment, withChecksum = false, config = defaultConfig) {
```
twiss, line 343: Just to match the variable naming elsewhere:
```suggestion
export function armor(messageType, body, partIndex, partTotal, customComment, emitChecksum = false, config = defaultConfig) {
```

src/encoding/armor.js
```diff
@@ -334,29 +334,38 @@ export function unarmor(input) {
...
   let text;
   let hash;
   if (messageType === enums.armor.signed) {
     text = body.text;
     hash = body.hash;
     body = body.data;
   }
+  // unless explicitly forbidden by the spec, we need to include the checksum to workaround an GnuPG bug
+  // where data fails to be decoded if the base64 ends with no padding chars (=) (see https://dev.gnupg.org/T7071)
```
twiss, line 352:
```suggestion
  // For non-v6 artifacts, we include the checksum to work around a bug in GnuPG
  // where data fails to be decoded if the base64 ends with no padding chars (=) (see https://dev.gnupg.org/T7071)
```

src/key/key.js
```diff
@@ -612,7 +612,9 @@ class Key {
     const revocationSignature = await helper.getLatestValidSignature(this.revocationSignatures, this.keyPacket, enums.signature.keyRevocation, dataToVerify, date, config);
     const packetlist = new PacketList();
     packetlist.push(revocationSignature);
-    return armor(enums.armor.publicKey, packetlist.write(), null, null, 'This is a revocation certificate');
+    // An ASCII-armored Transferable Public Key packet sequence of a v6 key MUST NOT contain a CRC24 footer.
+    const emitChecksum = this.keyPacket.version !== 6;
+    return armor(enums.armor.publicKey, packetlist.write(), null, null, 'This is a revocation certificate', emitChecksum);
```
twiss, line 617: Not really related to this change but maybe we should pass the config also here?

PR #1754, opened 2024-05-06
[v6] Drop support for browsers without native BigInts (e.g. Safari <14)

larabr:
`BigInt` is not polyfillable, and in order to support legacy browsers that do not implement it natively (e.g. Safari <14), a large alternative library (BN.js) needs to be bundled as well, and imported also by modern browsers (in OpenPGP.js v6, which depends the `@noble` libs).

We're finally dropping the fallback support as usage of Safari 11-13 has gone down significantly.
Aside from optimising the bundle size, thanks to the change we reduce the number of forked dependencies we need to maintain.


openpgp.d.ts
```diff
@@ -200,25 +202,25 @@ export function createMessage<T extends MaybeStream<string>>(options: { text: T,
...
-  T extends WebStream<infer X> ? WebStream<string> :
-  T extends NodeWebStream<infer X> ? NodeWebStream<string> :
+  T extends WebStream<Data> ? WebStream<string> :
+  T extends NodeWebStream<Data> ? NodeWebStream<string> :
   string
 >;
-export function encrypt<T extends MaybeStream<Data>>(options: EncryptOptions & { message: Message<T>, format: 'binary' }): Promise<
-  T extends WebStream<infer X> ? WebStream<Uint8Array> :
-  T extends NodeWebStream<infer X> ? NodeWebStream<Uint8Array> :
+export function encrypt<T extends MaybeStream<Data>>(options: EncryptOptions & { message: Message<T>, format: 'binary' }):Promise<
```
twiss, line 209:
```suggestion
export function encrypt<T extends MaybeStream<Data>>(options: EncryptOptions & { message: Message<T>, format: 'binary' }): Promise<
```

openpgp.d.ts
```diff
@@ -745,40 +747,40 @@ export function unarmor(input: string, config?: Config): Promise<{ text: string,
...
     noReason = 0, // No reason specified (key revocations or cert revocations)
     keySuperseded = 1, // Key is superseded (key revocations)
     keyCompromised = 2, // Key material has been compromised (key revocations)
     keyRetired = 3, // Key is retired and no longer used (key revocations)
-    userIDInvalid = 32, // User ID information is no longer valid (cert revocations)
+    userIDInvalid = 32 // User ID information is no longer valid (cert revocations)
   }
 
   export type compressionNames = 'uncompressed' | 'zip' | 'zlib' | 'bzip2';
   enum compression {
```
twiss, line 779: Should(n't) this one (and a few other ones below) be exported too?

src/biginteger.ts
```diff
@@ -0,0 +1,499 @@
+/**
+ * @fileoverview
+ * BigInteger implementation of basic operations
+ * that wraps the native BigInt library.
+ * Operations are not constant time,
+ * but we try and limit timing leakage where we can
+ */
+export default class BigInteger {
```
twiss, line 8: Do we still need this wrapper class? Perhaps we can just replace the uses of this class with direct uses of the operators (and perhaps a few helper functions) everywhere?

src/crypto/public_key/elliptic/brainpool/brainpoolP256r1.ts
```diff
@@ -0,0 +1,24 @@
+import { createCurve } from '@noble/curves/_shortw_utils';
+import { sha256 } from '@noble/hashes/sha256';
+import { Field } from '@noble/curves/abstract/modular';
+
+// brainpoolP256r1: https://datatracker.ietf.org/doc/html/rfc5639#section-3.4
+
+// eslint-disable-next-line new-cap
+const Fp = Field(BigInt('0xa9fb57dba1eea9bc3e660a909d838d726e3bf623d52620282013481d1f6e5377'));
```
twiss, line 8: Can we use BigInt syntax directly, now? E.g.
```suggestion
const Fp = Field(0xa9fb57dba1eea9bc3e660a909d838d726e3bf623d52620282013481d1f6e5377n);
```
and so on (everywhere else too)

src/crypto/public_key/prime.js
```diff
@@ -31,8 +31,6 @@ import { getRandomBigInteger } from '../random';
  * @async
  */
 export async function randomProbablePrime(bits, e, k) {
-  const BigInteger = await util.getBigInteger();
```
twiss, line 34: All the functions in this file don't need to be async anymore, AFAICT. Could you drop the `async`s (and `await`s wherever they're called)?

src/crypto/public_key/rsa.js
```diff
@@ -223,8 +222,6 @@ export async function generate(bits, e) {
  * @async
  */
 export async function validateParams(n, e, d, p, q, u) {
-  const BigInteger = await util.getBigInteger();
```
twiss, line 226: Same for this one and the functions below

src/crypto/random.js
```diff
@@ -51,8 +52,6 @@ export function getRandomBytes(length) {
  * @async
  */
 export async function getRandomBigInteger(min, max) {
-  const BigInteger = await util.getBigInteger();
```
twiss, line 54: Same here

src/openpgp.js
```diff
@@ -301,7 +301,7 @@ export async function encrypt({ message, encryptionKeys, signingKeys, passwords,
     // serialize data
     const armor = format === 'armored';
     const data = armor ? message.armor(config) : message.write();
-    return convertStream(data);
+    return await convertStream(data);
```
twiss, line 304: Nice catch :)

src/crypto/public_key/rsa.js
larabr: These functions better keep the same return signature, so that it's consistent across algorithms

src/crypto/public_key/elliptic/brainpool/brainpoolP256r1.ts
larabr: `BigInt` syntax is more widely supported apparently: https://github.com/paulmillr/noble-curves/issues/22 .

twiss: That issue is talking about their build environment, rather than any browsers. As far as I can tell, all browsers that support the BigInt() constructor also support the syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#browser_compatibility

twiss: OK. I suppose we can always revisit this later and see whether BigInt notation is widely supported enough yet.

twiss (approved):

PR #1756, opened 2024-05-16
[v6] Better error messages for unexpected ECDSA, EdDSALegacy, ECDH param encodings in keys and PKESK

larabr:
We got a report of a message including a PKESK packet where the ECDH x25519Legacy public point was missing the leading byte (0x40).
While decryption would naturally fail afterwards, this change ensures we fail sooner, with a clearer error message, and do not blindly pass down invalid data to the low-level crypto functions.
Note that such errors involving EC point encoding issues are thrown on usage, not on parsing.

Also, for EdDSALegacy, we now throw on key parsing in case of unexpected OID.

All of the above are technically breaking changes, but only for invalid data malformed in specific ways.
They are not to be considered security fixes, and are primarily aimed at helping debug.

### DIFF
--- a/src/crypto/public_key/elliptic/oid_curves.js
+++ b/src/crypto/public_key/elliptic/oid_curves.js
@@ -157,6 +157,7 @@ class CurveWithOID {
     this.web = params.web;
     this.payloadSize = params.payloadSize;
     this.sharedSize = params.sharedSize;
+    this.wireFormatLeadingByte = params.wireFormatLeadingByte;
     if (this.web && util.getWebCrypto()) {
       this.type = 'web';
     } else if (this.node && util.getNodeCrypto()) {
@@ -172,7 +173,7 @@ class CurveWithOID {
     switch (this.type) {
       case 'web':
         try {
-          return await webGenKeyPair(this.name);
+          return await webGenKeyPair(this.name, this.wireFormatLeadingByte);
         } catch (err) {
           util.printDebugError('Browser did not support generating ec key ' + err.message);
           return jsGenKeyPair(this.name);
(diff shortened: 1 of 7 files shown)


src/crypto/public_key/elliptic/oid_curves.js
```diff
@@ -268,8 +277,23 @@ async function validateStandardParams(algo, oid, Q, d) {
   return true;
 }
 
+/**
+ * Check whether the public point has a valid encoding.
+ * NB: this function does not check e.g. whether the point belongs to the curve.
+ */
+function checkPublicPointEnconding(oid, V) {
+  const curveName = oid.getName();
+  const { payloadSize, wireFormatLeadingByte } = curves[curveName];
```
twiss, line 286: This is a bit nitpicky but it would be better to copy these parameters to the curve object in the`CurveWithOID` constructor, like it does for the other parameters, and then pass the `curve` object here instead of the `oid`. That way, you don't have to get the name and look up the params by name again.

src/crypto/public_key/elliptic/oid_curves.js
```diff
@@ -57,7 +57,8 @@ const curves = {
     node: nodeCurves[enums.curve.nistP256],
     web: webCurves[enums.curve.nistP256],
     payloadSize: 32,
-    sharedSize: 256
+    sharedSize: 256,
+    wireFormatLeadingByte: 0x04
```
twiss, line 61: Perhaps we can reuse this param in generation/serialization too? (Not super critical ofc)

twiss (approved):

PR #1766, opened 2024-06-13
[v6] Delay checking unknown critical signature subpackets

twiss:
Throw when verifying signatures with unknown critical subpackets, instead of when parsing them.

Fixes #1765.

PR #1767, opened 2024-06-13
[v6] Store unhashed signature subpackets in structured objects

twiss:
Rather than storing the raw `Uint8Array` of the serialized unhashed subpacket, store the type and body separately, to match the new `unknownSubpackets` property added in #1766.

To be merged after #1766.

### DIFF
--- a/src/packet/signature.js
+++ b/src/packet/signature.js
@@ -396,10 +396,8 @@ class SignaturePacket {
    * @returns {Uint8Array} Subpacket data.
    */
   writeUnhashedSubPackets() {
-    const arr = [];
-    this.unhashedSubpackets.forEach(data => {
-      arr.push(writeSimpleLength(data.length));
-      arr.push(data);
+    const arr = this.unhashedSubpackets.map(({ type, critical, body }) => {
+      return writeSubPacket(type, critical, body);
     });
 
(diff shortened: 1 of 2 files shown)


src/packet/signature.js
```diff
@@ -407,23 +406,27 @@ class SignaturePacket {
...
+
     if (!hashed) {
-      this.unhashedSubpackets.push(bytes.subarray(mypos, bytes.length));
+      this.unhashedSubpackets.push({
+        type,
+        critical,
+        body: bytes.subarray(mypos, bytes.length)
+      });
       if (!allowedUnhashedSubpackets.has(type)) {
         return;
```
larabr, line 426: Not relevant for the PR, but we might want to log this as a debug error...?

twiss: Yeah, we could, though arguably it's not really an error, as storing stuff in the unhashed subpackets is allowed; it just shouldn't be trusted. But there could be semi-legitimate cases for e.g. a keyserver to store some information there if it trusts its own storage, for example (though it could also store it externally to the key, of course).

Very pedantically, it could also be argued that this variable should be called `trustableUnhashedSubpackets` instead of `allowedUnhashedSubpackets` ^.^

PR #1768, opened 2024-06-21
[v6] Use noble-ed25519 instead of tweetnacl for signature verification

larabr:
Faster than tweetnacl, and no constant-timeness required.

We are not using v2 for now, despite it being smaller, because it relies on bigint literals, and it requires polyfilling the WebCrypto lib manually in Node < 19.


larabr:
Applying this to Proton's fork only.

PR #1771, opened 2024-07-02
[v6] Fix ECDH encryption when using v6 keys

larabr:
Messages encrypted using v6 keys of type `'ecc'`, are not compliant with the crypto-refresh and will fail to decrypt in libraries other than OpenPGP.js v6 (up to beta.1) and gopenpgp v3.

This is due to incorrectly truncating the v6 key fingerprint input as part of the ECDH KDF step.
For a message to be affected, it must have been encrypted with keys generated using the `config.v6Keys` option in OpenPGP.js v6, and manually setting `type: 'ecc'` and `curve` to a NIST or Brainpool curve option. x25519/x448 keys are not affected.

### DIFF
--- a/src/packet/public_key_encrypted_session_key.js
+++ b/src/packet/public_key_encrypted_session_key.js
@@ -180,9 +180,10 @@ class PublicKeyEncryptedSessionKeyPacket {
     // No symmetric encryption algorithm identifier is passed to the public-key algorithm for a
     // v6 PKESK packet, as it is included in the v2 SEIPD packet.
     const sessionKeyAlgorithm = this.version === 3 ? this.sessionKeyAlgorithm : null;
+    const fingerprint = key.version === 5 ? key.getFingerprintBytes().subarray(0, 20) : key.getFingerprintBytes();
     const encoded = encodeSessionKey(this.version, algo, sessionKeyAlgorithm, this.sessionKey);
     this.encrypted = await crypto.publicKeyEncrypt(
-      algo, sessionKeyAlgorithm, key.publicParams, encoded, key.getFingerprintBytes());
+      algo, sessionKeyAlgorithm, key.publicParams, encoded, fingerprint);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/packet/public_key_encrypted_session_key.js
```diff
@@ -180,9 +180,10 @@ class PublicKeyEncryptedSessionKeyPacket {
     // No symmetric encryption algorithm identifier is passed to the public-key algorithm for a
     // v6 PKESK packet, as it is included in the v2 SEIPD packet.
     const sessionKeyAlgorithm = this.version === 3 ? this.sessionKeyAlgorithm : null;
+    const fingerprint = this.version === 6 ? key.getFingerprintBytes() : key.getFingerprintBytes().subarray(0, 20);
```
twiss, line 183: I would make v5 the special case, rather than v6; that way, v7 is more likely to work out of the box ^.^
```suggestion
    const fingerprint = this.version !== 5 ? key.getFingerprintBytes() : key.getFingerprintBytes().subarray(0, 20);
```
(v4 keys have 20 byte fingerprints so both branches work there)

src/packet/public_key_encrypted_session_key.js
```diff
@@ -202,7 +203,8 @@ class PublicKeyEncryptedSessionKeyPacket {
     const randomPayload = randomSessionKey ?
       encodeSessionKey(this.version, this.publicKeyAlgorithm, randomSessionKey.sessionKeyAlgorithm, randomSessionKey.sessionKey) :
       null;
-    const decryptedData = await crypto.publicKeyDecrypt(this.publicKeyAlgorithm, key.publicParams, key.privateParams, this.encrypted, key.getFingerprintBytes(), randomPayload);
+    const fingerprint = this.version === 6 ? key.getFingerprintBytes() : key.getFingerprintBytes().subarray(0, 20);
```
twiss, line 206:
```suggestion
    const fingerprint = this.version !== 5 ? key.getFingerprintBytes() : key.getFingerprintBytes().subarray(0, 20);
```

twiss (approved):

PR #1781, opened 2024-07-11
[v6] Try more AEAD ciphersuites for SEIPDv2

twiss:
Stick more closely to the algorithm preferences when creating an SEIPDv2 message, by trying additional combinations of the preferred symmetric algorithm and the preferred AEAD algorithm. If one of them is supported but not the other, we still use it (with the mandatory-to-implement algorithm for the other one).

For example, if the key we're encrypting with supports `[(AES-256, OCB)]`, we'll use that now (with the default algorithm preferences of `(AES-256, GCM)`). Before this change, we would use `(AES-128, OCB)` instead (i.e. both mandatory-to-implement algorithms).

### DIFF
--- a/test/general/key.js
+++ b/test/general/key.js
@@ -4129,6 +4129,34 @@ CNa5yq6lyexhsn2Vs8DsX+SOSUyNJiy5FyIJ
     expect(aeadAlgo).to.equal(openpgp.enums.aead.gcm);
   });
 
+  it('getPreferredCipherSuite with AEAD - one key - AES256-OCB', async function() {
+    const [key1] = await openpgp.readKeys({ armoredKeys: twoKeys });
+    const primaryUser = await key1.getPrimaryUser();
+    primaryUser.selfCertification.features = [9]; // Monkey-patch SEIPDv2 feature flag
+    primaryUser.selfCertification.preferredCipherSuites = [[9, 2]];
+    const { symmetricAlgo, aeadAlgo } = await getPreferredCipherSuite([key1], undefined, undefined, {
+      ...openpgp.config,
+      aeadProtect: true,
+      preferredAEADAlgorithm: openpgp.enums.aead.gcm
+    });
+    expect(symmetricAlgo).to.equal(openpgp.enums.symmetric.aes256);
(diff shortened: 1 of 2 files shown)


test/general/key.js
```diff
@@ -4129,6 +4129,34 @@ CNa5yq6lyexhsn2Vs8DsX+SOSUyNJiy5FyIJ
     expect(aeadAlgo).to.equal(openpgp.enums.aead.gcm);
   });
 
+  it('getPreferredCipherSuite with AEAD - one key - AES256-OCB', async function() {
+    const [key1] = await openpgp.readKeys({ armoredKeys: twoKeys });
+    const primaryUser = await key1.getPrimaryUser();
+    primaryUser.selfCertification.features = [9]; // Monkey-patch SEIPDv2 feature flag
+    primaryUser.selfCertification.preferredCipherSuites = [[9, 2]];
```
larabr, line 4136:
```suggestion
    primaryUser.selfCertification.preferredCipherSuites = [[openpgp.enums.symmetric.aes256, openpgp.enums.aead.ocb]];
```

test/general/key.js
```diff
@@ -4129,6 +4129,34 @@ CNa5yq6lyexhsn2Vs8DsX+SOSUyNJiy5FyIJ
...
+    });
+    expect(symmetricAlgo).to.equal(openpgp.enums.symmetric.aes256);
+    expect(aeadAlgo).to.equal(openpgp.enums.aead.ocb);
+  });
+
+  it('getPreferredCipherSuite with AEAD - one key - AES128-GCM', async function() {
+    const [key1] = await openpgp.readKeys({ armoredKeys: twoKeys });
+    const primaryUser = await key1.getPrimaryUser();
+    primaryUser.selfCertification.features = [9]; // Monkey-patch SEIPDv2 feature flag
+    primaryUser.selfCertification.preferredCipherSuites = [[7, 3]];
```
larabr, line 4150:
```suggestion
    primaryUser.selfCertification.preferredCipherSuites = [[openpgp.enums.symmetric.aes128, openpgp.enums.aead.gcm]];
```

PR #1785, opened 2024-08-09
[v6] Drop asmcrypto.js for AES fallbacks in favor of noble-ciphers

larabr:
Asm.js has now been deprecated for many years, and no performance gain is recorded for AES compared to vanilla JS.
The relevant AES fallback code is only used if the WebCrypto (resp. NodeCrypto) implementation is not available.

### DIFF
--- a/src/crypto/mode/cfb.js
+++ b/src/crypto/mode/cfb.js
@@ -21,7 +21,8 @@
  * @module crypto/mode/cfb
  */
 
-import { AES_CFB } from '@openpgp/asmcrypto.js/aes/cfb.js';
+import { cfb as nobleAesCfb, unsafe as nobleAesHelpers } from '@noble/ciphers/aes';
+
 import * as stream from '@openpgp/web-stream-tools';
 import util from '../../util';
 import enums from '../../enums';
@@ -174,17 +175,17 @@ class WebCryptoEncryptor {
 
       const encryptedBlocks = await this._runCBC(toEncrypt);
       xorMut(encryptedBlocks, plaintext);
-      this.prevBlock = encryptedBlocks.subarray(-this.blockSize).slice();
+      this.prevBlock = encryptedBlocks.slice(-this.blockSize);
 
       // take care of leftover data
-      if (leftover > 0) this.nextBlock.set(value.subarray(-leftover).slice());
+      if (leftover > 0) this.nextBlock.set(value.subarray(-leftover));
       this.i = leftover;
 
       return encryptedBlocks;
     }
 
     this.i += added.length;
-    let encryptedBlock = new Uint8Array();
+    let encryptedBlock;
     if (this.i === this.nextBlock.length) { // block ready to be encrypted
       const curBlock = this.nextBlock;
       encryptedBlock = await this._runCBC(this.prevBlock);
@@ -195,6 +196,8 @@ class WebCryptoEncryptor {
       const remaining = value.subarray(added.length);
       this.nextBlock.set(remaining, this.i);
       this.i += remaining.length;
+    } else {
+      encryptedBlock = new Uint8Array();
     }
 
     return encryptedBlock;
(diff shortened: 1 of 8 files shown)


src/crypto/mode/cfb.js
```diff
@@ -237,22 +240,111 @@ class WebCryptoEncryptor {
...
+    const src32 = getUint32Array(src);
+    const dst = new Uint8Array(src.length);
+    const dst32 = getUint32Array(dst);
+    for (let i = 0; i + 4 <= dst32.length; i += 4) {
+      const { s0: e0, s1: e1, s2: e2, s3: e3 } = nobleAesHelpers.encrypt(this.key, this.prevBlock[0], this.prevBlock[1], this.prevBlock[2], this.prevBlock[3]);
+      dst32[i + 0] = src32[i + 0] ^ e0;
+      dst32[i + 1] = src32[i + 1] ^ e1;
+      dst32[i + 2] = src32[i + 2] ^ e2;
+      dst32[i + 3] = src32[i + 3] ^ e3;
+      this.prevBlock = (this.forEncryption ? dst32 : src32).subarray(i, i + 4).slice();
```
twiss, line 266:
```suggestion
      this.prevBlock = (this.forEncryption ? dst32 : src32).slice(i, i + 4);
```

src/crypto/mode/cfb.js
```diff
@@ -237,22 +240,111 @@ class WebCryptoEncryptor {
...
+      const leftover = (value.length - missing) % this.blockSize;
+      const toProcess = util.concatUint8Array([
+        this.nextBlock,
+        value.subarray(missing, value.length - leftover)
+      ]);
+
+      const processedBlocks = this._runCFB(toProcess);
+
+      // take care of leftover data
+      if (leftover > 0) this.nextBlock.set(value.subarray(-leftover).slice());
```
twiss, line 286: I don't think this `.slice()` is necessary since `set` already copies it
```suggestion
      if (leftover > 0) this.nextBlock.set(value.subarray(-leftover));
```

src/crypto/mode/cfb.js
```diff
@@ -237,22 +240,111 @@ class WebCryptoEncryptor {
...
+    }
+
+    this.clearSensitiveData();
+    return result;
+  }
+
+  clearSensitiveData() {
+    this.nextBlock.fill(0);
+    this.prevBlock.fill(0);
+    this.key = null;
```
twiss, line 326: Shouldn't we be zeroing out `this.key` as well?

src/crypto/mode/cfb.js
```diff
@@ -237,22 +240,111 @@ class WebCryptoEncryptor {
   }
 }
 
+class NobleStreamProcessor {
+  constructor(forEncryption, algo, key, iv) {
+    this.forEncryption = forEncryption;
```
twiss, line 245:
```suggestion
    this.forEncryption = forEncryption;
```

src/crypto/mode/ocb.js
```diff
@@ -73,8 +73,8 @@ async function OCB(cipher, key) {
   // `encipher` and `decipher` cannot be async, since `crypt` shares state across calls,
   // hence its execution cannot be broken up.
   // As a result, WebCrypto cannot currently be used for `encipher`.
-  const encipher = block => AES_CBC.encrypt(block, key, false);
-  const decipher = block => AES_CBC.decrypt(block, key, false);
+  const encipher = block => nobleAesCbc(key, zeroBlock, { disablePadding: true }).encrypt(block);
+  const decipher = block => nobleAesCbc(key, zeroBlock, { disablePadding: true }).decrypt(block);
```
twiss, line 77: Can we reuse the cipher object here?
```suggestion
  const cbc = nobleAesCbc(key, zeroBlock, { disablePadding: true });
  const encipher = block => cbc.encrypt(block);
  const decipher = block => cbc.decrypt(block);
```

PR #1787, opened 2024-08-16
[v7] Add support for private keys which include public (sub)key packets

larabr:
As per [RFC9580](https://www.rfc-editor.org/rfc/rfc9580.html#name-transferable-secret-keys), a key is to considered private as long as it includes at least one secret (sub)key packet.

Previously, a `PrivateKey` could only include secret (sub)key packets.

To handle this better, we split `getSigningKey` into `getSigningKey` and `getVerificationKey`, so that the caller can indicate whether they require the private key material to be available. This is a breaking change, so we should release this in v7.

**!! Blocker:** a different mechanism to exclude private key material is expected to be standardized

### DIFF
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -291,8 +291,8 @@ export async function getPreferredCipherSuite(keys = [], date = new Date(), user
  * @returns {Promise<SignaturePacket>} Signature packet.
  */
 export async function createSignaturePacket(dataToSign, recipientKeys, signingKeyPacket, signatureProperties, date, recipientUserIDs, notations = [], detached = false, config) {
-  if (signingKeyPacket.isDummy()) {
-    throw new Error('Cannot sign with a gnu-dummy key.');
+  if (isPublicOrDummyKeyPacket(signingKeyPacket)) {
+    throw new Error('Cannot sign with a public or dummy key.');
   }
   if (!signingKeyPacket.isDecrypted()) {
     throw new Error('Signing key is not decrypted.');
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -554,4 +554,4 @@ export function checkKeyRequirements(keyPacket, config) {
   }
 }
 
-export const isPublicOrDummyKeyPacket = keyPacket => keyPacket.isDecrypted() === null || keyPacket.isDummy();
+export const isPublicOrDummyKeyPacket = keyPacket => !keyPacket.isPrivate() || keyPacket.isDummy();
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -553,3 +553,5 @@ export function checkKeyRequirements(keyPacket, config) {
       break;
   }
 }
+
+export const isPublicOrDummyKeyPacket = keyPacket => keyPacket.isDecrypted() === null || keyPacket.isDummy();
--- a/src/key/key.js
+++ b/src/key/key.js
@@ -257,16 +257,18 @@ class Key {
   }
 
   /**
-   * Returns last created key or key by given keyID that is available for signing and verification
+   * Returns last created key or key by given keyID that is available for signing or verification
+   * @param  {Boolean} [needPrivateKeyMaterial] - whether the key will be used for signing
    * @param  {module:type/keyid~KeyID} [keyID] - key ID of a specific key to retrieve
    * @param  {Date} [date] - use the fiven date date to  to check key validity instead of the current date
    * @param  {Object} [userID] - filter keys for the given user ID
    * @param  {Object} [config] - Full configuration, defaults to openpgp.config
    * @returns {Promise<Key|Subkey>} signing key
-   * @throws if no valid signing key was found
+   * @throws if no valid signing/verification key was found
    * @async
+   * @private
    */
-  async getSigningKey(keyID = null, date = new Date(), userID = {}, config = defaultConfig) {
+  async getSigningOrVerificationKey(needPrivateKeyMaterial = false, keyID = null, date = new Date(), userID = {}, config = defaultConfig) {
     await this.verifyPrimaryKey(date, userID, config);
     const primaryKey = this.keyPacket;
     try {
@@ -284,7 +286,7 @@ class Key {
     for (const subkey of subkeys) {
       if (!keyID || subkey.getKeyID().equals(keyID)) {
         try {
-          if (!keyID && helper.isPublicOrDummyKeyPacket(subkey.keyPacket)) {
+          if (needPrivateKeyMaterial && helper.isPublicOrDummyKeyPacket(subkey.keyPacket)) {
             throw new Error('Cannot sign with a public or dummy key');
           }
           await subkey.verify(date, config);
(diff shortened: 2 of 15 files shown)


src/key/private_key.js
```diff
@@ -83,6 +80,9 @@ class PrivateKey extends PublicKey {
     const primaryKey = this.keyPacket;
     const keys = [];
     for (let i = 0; i < this.subkeys.length; i++) {
+      if (helper.isPublicOrDummyKeyPacket(this.subkeys[i].keyPacket)) {
+        continue;
+      }
```
twiss, line 85: While this is consistent with the existing behavior, I think we could improve the error reporting by keeping this in `decryptSessionKeys`...

src/message.js
```diff
@@ -215,9 +215,6 @@ export class Message {
           // do not check key expiration to allow decryption of old messages
           const decryptionKeyPackets = (await decryptionKey.getDecryptionKeys(pkeskPacket.publicKeyID, null, undefined, config)).map(key => key.keyPacket);
           await Promise.all(decryptionKeyPackets.map(async function(decryptionKeyPacket) {
-            if (!decryptionKeyPacket || decryptionKeyPacket.isDummy()) {
-              return;
-            }
```
twiss, line 220: ...by throwing an exception here instead, and then expanding the `try/catch`es below to cover this (and the one below as well, since theoretically speaking you could have one key packet that's decrypted and one that's not).

src/key/private_key.js
```diff
@@ -77,28 +74,44 @@ class PrivateKey extends PublicKey {
...
    */
   async getDecryptionKeys(keyID, date = new Date(), userID = {}, config = defaultConfig) {
     const primaryKey = this.keyPacket;
     const keys = [];
+    let exception = null;
     for (let i = 0; i < this.subkeys.length; i++) {
+      if (helper.isPublicOrDummyKeyPacket(this.subkeys[i].keyPacket)) {
+        exception = exception || new Error('Public or gnu-dummy key packets cannot be used for decryption');
+        continue;
+      }
```
twiss, line 88: I would move this one line down, as it's only relevant if the key ID matches

src/key/helper.js
```diff
@@ -291,6 +291,9 @@ export async function getPreferredCipherSuite(keys = [], date = new Date(), user
  * @returns {Promise<SignaturePacket>} Signature packet.
  */
 export async function createSignaturePacket(dataToSign, recipientKeys, signingKeyPacket, signatureProperties, date, recipientUserIDs, notations = [], detached = false, config) {
+  if (signingKeyPacket.isDecrypted() === null) {
```
larabr: Let's merge with the if below and use the `isPublicOrDummyKeyPacket` helper instead?

src/key/key.js
```diff
@@ -248,16 +257,18 @@ class Key {
...
    * @param  {Object} [userID] - filter keys for the given user ID
    * @param  {Object} [config] - Full configuration, defaults to openpgp.config
    * @returns {Promise<Key|Subkey>} signing key
-   * @throws if no valid signing key was found
+   * @throws if no valid signing/verification key was found
    * @async
+   * @private
    */
-  async getSigningKey(keyID = null, date = new Date(), userID = {}, config = defaultConfig) {
+  async getSigningOrVerificationKey(needPrivateKeyMaterial = false, keyID = null, date = new Date(), userID = {}, config = defaultConfig) {
```
larabr, line 271: Can we make this a local helper that takes a Key input instead? So that we don't expose this function at all

Then I'd also make all params required

src/key/private_key.js
```diff
@@ -146,31 +157,30 @@ class PrivateKey extends PublicKey {
...
       signingKeyPacket = this.keyPacket;
     } else {
       /**
        * It is enough to validate any signing keys
        * since its binding signatures are also checked
        */
-      const signingKey = await this.getSigningKey(null, null, undefined, { ...config, rejectPublicKeyAlgorithms: new Set(), minRSABits: 0 });
-      // This could again be a dummy key
-      if (signingKey && !signingKey.keyPacket.isDummy()) {
+      const signingKey = await this.getSigningKey(null, null, undefined, { ...config, rejectPublicKeyAlgorithms: new Set(), minRSABits: 0 }).catch(() => {});
```
larabr, line 174:
```suggestion
      const signingKey = await this.getSigningKey(null, null, undefined, { ...config, rejectPublicKeyAlgorithms: new Set(), minRSABits: 0 }).catch(() => null);
```
To make the intention clearer

test/general/config.js
larabr, line 1: These changes don't seem correct?

src/key/helper.js
```diff
@@ -553,3 +556,5 @@ export function checkKeyRequirements(keyPacket, config) {
       break;
   }
 }
+
+export const isPublicOrDummyKeyPacket = keyPacket => keyPacket.isDecrypted() === null || keyPacket.isDummy();
```
larabr: I am thinking it'd make sense to add a `isPrivate` to the packet classes (same that we already have for PublicKey and PrivateKey). It can call `isDecrypted` internally. This null check looks quite confusing, and I think being able to check whether there is any public key in a parsed private key is a reasonable use-case

src/key/private_key.js
twiss: We have this pattern in a bunch of places and I think it's clear enough that the intention is just to ignore errors IMHO

test/general/config.js
twiss: The point of these tests is just to check that the config gets passed along. The one for `openpgp.decryptKey` was doing so in a very roundabout way by relying on the fact that decryption would fail if we don't like the hash algorithm of the self-signature. However that doesn't really make sense and is no longer the case after the changes in this PR. Now we test that `maxArgon2MemoryExponent` is passed along instead, which is a lot more relevant to `openpgp.decryptKey`.

larabr: I see; there is no need for the `aeadProtect` option though?

src/key/private_key.js
larabr: I had to have a hard look to understand what was going on, and -- more importantly -- that the fall-through was intentional; I know it was already like that, but I am not a fan... i think we should just use a three-way conditional given the new simpler conditions now?

```js
const signingKeyPacket = !helper.isPublicOrDummyKeyPacket(this.keyPacket)
  ? this.keyPacket
  /**
   * It is enough to validate any signing keys
   * since its binding signatures are also checked;
   * however, `getSigningKey` will throw if only private encryption keys are available
   */
  : (await this.getSigningKey(null, null, undefined, { ...config, rejectPublicKeyAlgorithms: new Set(), minRSABits: 0 }).catch(() => {}))?.keyPacket;
```

src/key/key.js
larabr: @twiss can we still make the params mandatory? The first one should be mandatory at least

test/general/key.js
```diff
@@ -1857,6 +1857,53 @@ S2rSjQ4JF0Ktgdr9585haknpGwr31t486KxXOY4AEsiBmRyvTbaQegwKaQ+C
 -----END PGP PRIVATE KEY BLOCK-----
 `;
 
+// public primary key with:
+// - one signing public subkey
+// - one signing private subkey
+// - one encryption public subkey
+// - one encryption private subkey
+const privateKeyWithPublicKeyPackets = `-----BEGIN PGP PRIVATE KEY BLOCK-----
```
larabr, line 1865: Can we also add a test of a private primary key with public subkeys? Since the parsing logic is separate

PR #1801, opened 2024-10-28
[v6] Switch to SHA512 as default preferred hash algo (`config.preferredHashAlgorithm`)

larabr:
This affects the preferences of newly generated keys, which by default will have SHA512 as first hash algo preference.
SHA512 will also be used when signing, as long as the signing key declares support for the algorithm.

SHA512 is usually faster than SHA256 (former default) on 64-bit platforms.

NB: SHA512 cannot be used with RSA 512-bit keys (which are insecure, but may be used e.g. for testing), so error messages have been improved for this edge case.

### DIFF
--- a/src/crypto/public_key/rsa.js
+++ b/src/crypto/public_key/rsa.js
@@ -26,6 +26,7 @@ import { uint8ArrayToB64, b64ToUint8Array } from '../../encoding/base64';
 import { emsaEncode, emeEncode, emeDecode } from '../pkcs1';
 import enums from '../../enums';
 import { bigIntToNumber, bigIntToUint8Array, bitLength, byteLength, mod, modExp, modInv, uint8ArrayToBigInt } from '../biginteger';
+import hash from '../hash';
 
 const webCrypto = util.getWebCrypto();
 const nodeCrypto = util.getNodeCrypto();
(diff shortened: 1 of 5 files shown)


src/crypto/public_key/rsa.js
```diff
@@ -264,9 +273,6 @@ async function bnSign(hashAlgo, n, d, hashed) {
   n = uint8ArrayToBigInt(n);
   const m = uint8ArrayToBigInt(emsaEncode(hashAlgo, hashed, byteLength(n)));
   d = uint8ArrayToBigInt(d);
-  if (m >= n) {
-    throw new Error('Message size cannot exceed modulus size');
-  }
```
twiss, line 269: The comment above makes me think we should still keep this error, for edge cases where the digest size fits but the encoded message size doesn't?

src/crypto/public_key/rsa.js
```diff
@@ -45,6 +46,14 @@ const _1n = BigInt(1);
  * @async
  */
 export async function sign(hashAlgo, data, n, e, d, p, q, u, hashed) {
+  if (hash.getHashByteLength(hashAlgo) >= n.length) {
+    // Throw here instead of `emsaEncode` below, to provide a clearer and consistent error
+    // e.g. if a 512-bit RSA key is used with a SHA-512 digest.
+    // The size limit is actually slightly different but here we only care about throwing
+    // on common key sizes.
+    throw new Error('digest size cannot exceed key modulus size');
```
twiss, line 54:
```suggestion
    throw new Error('Digest size cannot exceed key modulus size');
```

src/crypto/public_key/rsa.js
larabr: `emsaEncode` already checks the length and throws if needed, this is an unreachable case

twiss: OK :+1:

twiss (approved):

PR #1802, opened 2024-10-30
[v6] Determine signature hash prefs based on recipient keys instead of signing key

larabr:
Changes:
- The signature will be generated using the preferred hash algo from the recipient keys:
  - in `openpgp.encrypt`, when passing `signingKeys`, the `encryptionKeys` are treated as recipient keys
  - in `openpgp.sign`, a `recipientKeys` option has been added

  If no recipient keys are available, the signing key preferences are used (this was also the existing behavior).

- The hash algo selection logic has been reworked as follows:
if `config.preferredHashAlgo` appears in the prefs of all recipients, we pick it;
otherwise, we use the strongest supported algo (note: SHA256 is always implicitly supported by all keys), as long as it is compatible with the signing key (e.g. ECC keys require minimum digest sizes).
With this change, `config.preferredHashAlgo` is picked even if it's weaker than the preferences of the recipient keys.

Previously, only the preferences of the signing key were used to determine the hash algo to use, but this is in contrast to the RFC: https://www.rfc-editor.org/rfc/rfc9580.html#section-5.2.3.16-2 .
Also, an algo stronger than `config.preferredHashAlgo` would be used, if the signing key declared it as first preference.

### DIFF
--- a/src/key/helper.js
+++ b/src/key/helper.js
@@ -90,7 +90,7 @@ export async function createBindingSignature(subkey, primaryKey, options, config
   const signatureProperties = { signatureType: enums.signature.subkeyBinding };
   if (options.sign) {
     signatureProperties.keyFlags = [enums.keyFlags.signData];
-    signatureProperties.embeddedSignature = await createSignaturePacket(dataToSign, null, subkey, {
+    signatureProperties.embeddedSignature = await createSignaturePacket(dataToSign, [], subkey, {
       signatureType: enums.signature.keyBinding
     }, options.date, undefined, undefined, undefined, config);
   } else {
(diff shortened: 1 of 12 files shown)


src/key/helper.js
```diff
@@ -100,41 +100,88 @@ export async function createBindingSignature(subkey, primaryKey, options, config
...
+      supportedAlgosMap.set(algo, supportedAlgosMap.has(algo) ? supportedAlgosMap.get(algo) + 1 : 1);
     }
   }
-  switch (keyPacket.algorithm) {
-    case enums.publicKey.ecdsa:
-    case enums.publicKey.eddsaLegacy:
-    case enums.publicKey.ed25519:
-    case enums.publicKey.ed448:
-      prefAlgo = crypto.getPreferredCurveHashAlgo(keyPacket.algorithm, keyPacket.publicParams.oid);
+  const isSupportedHashAlgo = hashAlgo => targetKeys.length === 0 || supportedAlgosMap.get(hashAlgo) === targetKeys.length || hashAlgo === defaultAlgo;
```
twiss, line 138: I think we also need some logic to handle the case where all recipients support a hash algorithm that we don't support? Otherwise `crypto.hash.getHashByteLength` below will throw, I think?

twiss (approved):

PR #1811, opened 2024-11-21
Fix decryption support for non-standard, legacy AEAD-encrypted messages and keys that used `experimentalGCM` from OpenPGP.js v5

larabr:
This change affects only keys or messages generated using features marked as "experimental" in OpenPGP.js v5.
Messages and keys encrypted with OpenPGP.js v6 are not affected since there was a bug that prevented using the `experimentalGCM` algo; the bug is also fixed in this MR, to e.g. allow decrypting existing entities.

See commits for details.

### DIFF
--- a/src/crypto/crypto.js
+++ b/src/crypto/crypto.js
@@ -468,17 +468,22 @@ export function generateSessionKey(algo) {
 /**
  * Get implementation of the given AEAD mode
  * @param {enums.aead} algo
+ * @param {Boolean} [acceptExperimentalGCM] - whether to allow the non-standard, legacy `experimentalGCM` algo
  * @returns {Object}
  * @throws {Error} on invalid algo
  */
-export function getAEADMode(algo) {
+export function getAEADMode(algo, acceptExperimentalGCM = false) {
   switch (algo) {
     case enums.aead.eax:
       return mode.eax;
     case enums.aead.ocb:
       return mode.ocb;
     case enums.aead.gcm:
+      return mode.gcm;
     case enums.aead.experimentalGCM:
+      if (!acceptExperimentalGCM) {
+        throw new Error('Unexpected non-standard `experimentalGCM` AEAD algorithm provided in `config.preferredAEADAlgorithm`: use `gcm` instead');
+      }
       return mode.gcm;
     default:
       throw new Error('Unsupported AEAD mode');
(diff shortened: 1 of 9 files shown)


src/crypto/crypto.js
```diff
@@ -468,12 +468,26 @@ export function generateSessionKey(algo) {
...
  * Get implementation of the given AEAD mode
  * @param {enums.aead} algo
+ * @param {Boolean} [rejectExperimentalGCM] - whether to throw on non-standard, legacy `experimentalGCM` algo
  * @returns {Object}
  * @throws {Error} on invalid algo
  */
-export function getAEADMode(algo) {
-  const algoName = enums.read(enums.aead, algo);
-  return mode[algoName];
+export function getAEADMode(algo, rejectExperimentalGCM = false) {
```
twiss, line 475: Bit nitpicky, but probably it's safer to swap the meaning and the default:
```suggestion
export function getAEADMode(algo, allowExperimentalGCM = false) {
```
That way, it'll do the right thing by default if we use this function in new places (like persistent symmetric keys).

test/general/key.js
```diff
@@ -2286,7 +2286,7 @@ function versionSpecificTests() {
     openpgp.config.preferredSymmetricAlgorithm = openpgp.enums.symmetric.aes192;
     openpgp.config.preferredHashAlgorithm = openpgp.enums.hash.sha224;
     openpgp.config.preferredCompressionAlgorithm = openpgp.enums.compression.zlib;
-    openpgp.config.preferredAEADAlgorithm = openpgp.enums.aead.experimentalGCM;
+    openpgp.config.preferredAEADAlgorithm = openpgp.enums.aead.gcm;
```
twiss, line 2289: Since this is the default, to make this test do something, it's probably better to set it to a different one, i.e. ocb or eax.

src/crypto/crypto.js
```diff
@@ -468,12 +468,26 @@ export function generateSessionKey(algo) {
 /**
  * Get implementation of the given AEAD mode
  * @param {enums.aead} algo
+ * @param {Boolean} [acceptExperimentalGCM] - whether to allow on non-standard, legacy `experimentalGCM` algo
```
twiss, line 471:
```suggestion
 * @param {Boolean} [acceptExperimentalGCM] - whether to allow the non-standard, legacy `experimentalGCM` algo
```

twiss (approved):

PR #1812, opened 2024-11-22
Improve internal tree-shaking and lazy load md5

larabr:
See commits for details.

The improved internal tree-shaking for crypto functions helps with unused code elimination; also, md5 can be lazily load as it's used by legacy messages only. This is the impact on the lightweight build before and after the changes:

<details>
  <summary>Open spoiler</summary>
image
image
</details>


twiss (approved):
:+1: Cool, thanks!

PR #1850, opened 2025-05-06
Bump dev dependencies to latest versions

larabr:
ESLint update still blocked due to https://github.com/airbnb/javascript/issues/2961 . Upgrade not needed for now, but for future reference, the current config/plugins should still be usable via https://eslint.org/blog/2022/08/new-config-system-part-2/#backwards-compatibility-utility .

Chai v5 update also needs to be delayed due to missing support for older Safari versions (https://github.com/chaijs/chai/issues/1687)


twiss:
From the CI failure, it looks like one of the dependencies uses some syntax that Safari iOS 14 doesn't support; not exactly sure what. But, that probably means we should postpone one of these until we can drop support for it.

larabr:
Yeah I am looking into a fix/workaround

twiss:
It seems like even with this more modest update from 4.4 to 4.5, the build is unhappy due to the updated `type-detect` dependency

larabr:
Yeah I've seen that issue before, have to look into why it's not properly handled by rollup

PR #1854, opened 2025-05-19
Prefer subkeys with higher algorithm IDs

twiss:
In case of equal creation timestamps, pick the signing/encryption subkey with the highest algorithm ID, on the assumption that that's the most modern/secure algorithm.

### DIFF
--- a/test/general/key.js
+++ b/test/general/key.js
@@ -4659,13 +4659,14 @@ I8kWVkXU6vFOi+HWvv/ira7ofJu16NnoUkhclkUrk0mXubZvyl4GBg==
       expect(subkeyOid.getName()).to.be.equal(pkOid.getName());
       expect(subkey2.getAlgorithmInfo().algorithm).to.be.equal('eddsaLegacy');
       await subkey2.verify();
+      expect(await newPrivateKey.getSigningKey()).to.be.equal(subkey2);
     });
 
-    it('create and add a new ecdsa subkey to a eddsa key', async function() {
+    it('create and add a new ecdsa subkey to a eddsa key (equal timestamp)', async function() {
       const userID = { name: 'test', email: 'a@b.com' };
-      const { privateKey } = await openpgp.generateKey({ curve: 'ed25519Legacy', userIDs: [userID], format: 'object', subkeys:[] });
+      const { privateKey } = await openpgp.generateKey({ curve: 'ed25519Legacy', userIDs: [userID], format: 'object', subkeys: [{ sign: true }] });
       const total = privateKey.subkeys.length;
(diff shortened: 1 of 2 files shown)


test/general/key.js
```diff
@@ -4674,11 +4675,47 @@ I8kWVkXU6vFOi+HWvv/ira7ofJu16NnoUkhclkUrk0mXubZvyl4GBg==
       expect(subkey.getAlgorithmInfo().curve).to.be.equal('nistP256');
       expect(newPrivateKey.getAlgorithmInfo().algorithm).to.be.equal('eddsaLegacy');
       expect(subkey.getAlgorithmInfo().algorithm).to.be.equal('ecdsa');
+      await subkey.verify();
```
larabr, line 4678: Is this needed? The get*Key helpers should already verify signatures internally

test/general/key.js
```diff
@@ -4710,6 +4748,24 @@ I8kWVkXU6vFOi+HWvv/ira7ofJu16NnoUkhclkUrk0mXubZvyl4GBg==
...
+      let newPrivateKey = await privateKey.addSubkey({ type: 'rsa', date: new Date(+privateKey.subkeys[0].keyPacket.created + 1000) });
+      const armoredKey = newPrivateKey.armor();
+      newPrivateKey = await openpgp.readKey({ armoredKey });
+      const subkey = newPrivateKey.subkeys[total];
+      expect(subkey).to.exist;
+      expect(newPrivateKey.subkeys.length).to.be.equal(total + 1);
+      expect(subkey.getAlgorithmInfo().bits).to.be.equal(4096);
+      expect(subkey.getAlgorithmInfo().algorithm).to.be.equal('rsaEncryptSign');
+      await subkey.verify(new Date(+privateKey.subkeys[0].keyPacket.created + 1000));
+      expect(await newPrivateKey.getEncryptionKey(undefined, new Date(+privateKey.subkeys[0].keyPacket.created + 1000))).to.be.equal(subkey);
```
larabr, line 4768:
```suggestion
      expect(await newPrivateKey.getEncryptionKey(undefined, new Date(+privateKey.subkeys[0].getCreationTime() + 1000))).to.be.equal(subkey);
```

I'd avoid low-level access whenever possible

larabr (approved):

test/general/key.js
twiss: This particular one is needed yeah, because the signing key we add _isn't_ selected by the call to `getSigningKey` below, and we want to make sure it's not because it's we created an invalid subkey, but rather because we prefer the other subkey's algorithm.

Some of the other ones aren't needed, but I'd prefer to keep them for simplicity and to make it clear what's broken if something fails.

PR #1857, opened 2025-05-28
Improve type definition for the User class

kkredit:
This PR improves the TypeScript type definition for the User class. Please let me know if there are any testing steps. I tested this locally by using these type definitions in another project.


twiss (approved):
:+1: Thanks!

PR #1887, opened 2025-08-26
Switch from seek-bzip to unbzip2-stream

twiss:
To allow streaming decompression of bzip2 data.


package.json
```diff
@@ -67,9 +67,9 @@
     "@noble/curves": "^1.9.6",
     "@noble/hashes": "^1.8.0",
     "@openpgp/jsdoc": "^3.6.11",
-    "@openpgp/seek-bzip": "^1.0.5-git",
     "@openpgp/tweetnacl": "^1.0.4-2",
-    "@openpgp/web-stream-tools": "~0.1.3",
+    "@openpgp/unbzip2-stream": "file:../unbzip2-stream",
+    "@openpgp/web-stream-tools": "file:../web-stream-tools",
```
larabr, line 72:
```suggestion
    "@openpgp/unbzip2-stream": "openpgpjs/unbzip2-stream#e156342e78cfe483ea1ddac0c0b6203cbb7cf220",

    "@openpgp/web-stream-tools": "openpgpjs/web-stream-tools#717c809b81438832f616948b8456cc3b69f72e76",
```

For CI to run

src/packet/compressed_data.js
```diff
@@ -205,8 +209,12 @@ function zlib(compressionStreamInstantiator, ZlibStreamedConstructor) {
...
 function bzip2Decompress() {
   return async function(data) {
-    const { decode: bunzipDecode } = await import('@openpgp/seek-bzip');
-    return streamFromAsync(async () => bunzipDecode(await streamReadToEnd(data)));
+    const { default: Unbzip2Stream } = await import('@openpgp/unbzip2-stream');
+    if (isArrayStream(data)) {
+      data = await streamReadToEnd(data);
+    }
+    data = toStream(data);
+    return data.pipeThrough(new Unbzip2Stream());
```
larabr, line 217: This is more of a comment on the stream tools API, but I think the interaction between the helpers is confusing; I think we should have `ArrayStream` support pipeThrough; or alternatively add a (required) arg to `toStream` like `convertArrayStream` so that it takes care of it internally.

twiss: `ArrayStream` does support `pipeThrough`, however, in that case all of the data gets decompressed regardless of what happens after it, since `ArrayStream`s don't have backpressure support.

In this case, since we want to limit the output size of the stream afterwards (in a future PR), we want to use a real stream.

You're right that the docs need updating, though.

package.json
```diff
@@ -68,8 +68,8 @@
     "@noble/curves": "^1.9.7",
     "@noble/hashes": "^1.8.0",
     "@openpgp/jsdoc": "^4.0.4",
-    "@openpgp/seek-bzip": "^1.0.5-git",
     "@openpgp/tweetnacl": "^1.0.4-2",
+    "@openpgp/unbzip2-stream": "~2.0.0",
```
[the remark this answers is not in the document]
twiss: I matched the one below rather than the one above.. anyway I don't think there's a specific reason

larabr (dismissed):

larabr (approved):

PR #1914, opened 2025-10-01
Linter: update to eslint v9

larabr:
This eslint version brings breaking changes related to the eslint config.
eslint-airbnb-config has yet to release a compatible version (https://github.com/airbnb/javascript/issues/2961).
Hence we drop the dependency in favor of the built-in recommended configs.


src/key/subkey.js
```diff
@@ -139,7 +139,7 @@ class Subkey {
       this.keyPacket = subkey.keyPacket;
     }
     // update missing binding signatures
-    const that = this;
+    const that = this;
```
twiss, line 142:
```suggestion
    const that = this;
```

src/key/key.js
```diff
@@ -510,10 +510,10 @@ class Key {
...
     }
     if (!users.length) {
-      // eslint-disable-next-line @typescript-eslint/no-throw-literal
+      // eslint-disable-next-line @typescript-eslint/only-throw-error
       throw exception || new Error('Could not find primary user');
     }
-    await Promise.all(users.map(async function (a) {
+    await Promise.all(users.map(a => {
       return a.selfCertification.revoked || a.user.isRevoked(a.selfCertification, null, date, config);
     }));
```
twiss, line 518: I know that linters like to complain about "useless" uses of `async`, but I think it makes it clearer here to explain why the `Promise.all` is needed. Perhaps we could do:
```suggestion
    await Promise.all(users.map(async a => {
      a.selfCertification.revoked || await a.user.isRevoked(a.selfCertification, null, date, config);
    }));
```
(the code here is also a bit strange because we rely on the side-effect of setting the `revoked` property rather than just using the return value directly, but OK).

src/util.js
```diff
@@ -638,17 +638,18 @@ const util = {
...
+        let exception;
+        await Promise.all(promises.map(async promise => {
+          try {
+            resolve(await promise);
+          } catch (e) {
+            exception = e;
+          }
+        }));
+        reject(exception);
+      })();
```
twiss, line 652: What's the point of this IIFE? Is it just to make the linter happy? The two versions are equivalent and the old one is shorter..

test/benchmarks/memory_usage.js
```diff
@@ -105,7 +108,7 @@ class MemoryBenchamrkSuite {
  * Memory usage tests.
  * All the necessary variables must be declared inside the test function.
  */
-(async () => {
+void (async () => {
```
twiss, line 111: If this is to make the linter happy, could we just add a comment instead, to make that explicit? Adding code to make the linter happy feels very silly IMHO..
(below as well)

openpgp.d.ts
```diff
@@ -337,10 +337,9 @@ export class Message<T extends MaybeStream<Data>> {
 }
 
 /* ############## PACKET #################### */
-
-export declare abstract class BasePacket {
+export declare abstract class BasePacket<AsyncRead extends boolean = false> {
   static readonly tag: enums.packet;
-  public read(bytes: Uint8Array): void;
+  public read(bytes: Uint8Array): AsyncRead extends true ? Promise<void> : void;
```
twiss, line 342: Maybe we can avoid this complexity and just mark all of the `read` functions `async`?

src/key/key.js
larabr: The linter now throws if `awaits` are redundant (or missing) so I think it's probably more confusing to specify them unnecessarily

src/util.js
larabr: The new version makes it clear the code is not "awaited"

openpgp.d.ts
larabr: Breaking change, and also confusing

src/key/key.js
twiss: `isRevoked` is async so that one is not unnecessary

twiss: By marking the function as `async`, it still returns a promise.
`a.selfCertification.revoked`, by contrast, _isn't_ a promise, so the version I suggested is actually more type-consistent because now it always returns a promise, instead of only sometimes.

twiss (approved):

PR #1919, opened 2025-10-20
[v7] Bump `@noble` libs to v2

larabr:
See commits for more details.
The new lib versions include support for md5 and brainpool curves, so we can drop our internal implementations.

NB: the libs require bumping the min supported version from Node 18 to Node 20, hence this change will be released in  OpenPGP.js v7 .

Supersedes #1911, #1912 , #1913 .

<details><summary>Bundled lib size changes</summary>
Before:
image

After (standalone brainpool modules were ~2KB) : 
image

</details>


src/crypto/public_key/elliptic/ecdsa.js
```diff
@@ -26,7 +26,6 @@ import util from '../../../util.js';
 import { getRandomBytes } from '../../random.js';
```
larabr, line 26: Should double check this padding issue: https://github.com/openpgpjs/openpgpjs/actions/runs/29019822522/job/86123892418?pr=1919

twiss: WebKit on Windows seems to sometimes generate P-521 private keys that are one byte short. We could pad them manually but it seems safer to just add a check and fallback to JS: https://github.com/openpgpjs/openpgpjs/pull/1919/commits/79ae63bc871b407185766dfa4ad485794b6edc49

larabr: The error was thrown here actually I think: https://github.com/openpgpjs/openpgpjs/blob/2ac0048404b74a3595d503125b53f3b3d0486bec/src/crypto/public_key/elliptic/ecdh.js#L220 .

So let's just pad the key there; we already do the same on key [generation](https://github.com/openpgpjs/openpgpjs/blob/2ac0048404b74a3595d503125b53f3b3d0486bec/src/crypto/public_key/elliptic/oid_curves.js#L211) and parsing; this one-byte-off thing is not uncommon for p521. 

(I think on key generation we could/should limit it to pad at most one byte though, and only for this specific curve)

twiss: > The error was thrown here actually I think:

Yes, and that in turn calls out to `webGenKeyPair` here: https://github.com/openpgpjs/openpgpjs/blob/2ac0048404b74a3595d503125b53f3b3d0486bec/src/crypto/public_key/elliptic/oid_curves.js#L177 (even though we're in `jsPublicEphemeralKey`, Web Crypto is still used for keygen if available).

> So let's just pad the key there; we already do the same on key [generation](https://github.com/openpgpjs/openpgpjs/blob/2ac0048404b74a3595d503125b53f3b3d0486bec/src/crypto/public_key/elliptic/oid_curves.js#L211) and parsing; this one-byte-off thing is not uncommon for p521.

I think padding here is potentially dangerous. What if the implementation accidentally returns a P-256 sized key for P-521, should we ignore it too?

The format for JWK strictly requires a fixed length, here: https://www.rfc-editor.org/info/rfc7518/#section-6.2.2.1. So, the Web Crypto implementation of WebKit on Windows is just wrong and should be fixed, instead; I'll file a bug with them.

larabr: Windows webkit is not the only one that returns 1-byte short keys, I've seen the failure in Node and Safari before. If the curve is P-521 it's fine to pad the last byte, the noble lib is just being stricter than before about this it seems.

For other curves, or if the padding required is larger than 1 byte, we should just let the noble lib throw and not silently fallback to JS -- we've only been keeping fallbacks for WebCrypto for specific documented bugs.

PR #1942, opened 2025-12-03
[v7] Don't test WebKit on Windows

twiss:
It often fails, and nobody uses it.


package.json
```diff
@@ -58,8 +58,8 @@
     "benchmark-memory-usage": "node test/benchmarks/memory_usage.js",
     "prebrowsertest": "npm run build-test",
     "browsertest": "web-test-runner --config test/web-test-runner.config.js --group local --manual --open",
-    "test-browser": "web-test-runner --config test/web-test-runner.config.js --group local --playwright --browsers chromium firefox webkit",
-    "test-browser:ci": "web-test-runner --config test/web-test-runner.config.js --group headless:ci",
+    "test-browser": "web-test-runner --config test/web-test-runner.config.js --group headless:ci",
+    "test-browser:no-webkit": "web-test-runner --config test/web-test-runner.config.js --group headless:ci:no-webkit",
```
twiss, line 62: I don't often use it but maybe we should keep `test-browser` as it was and rename these to `test-browser:ci` and `test-browser:ci:no-webkit`, instead, so that there's still a command dedicated to running the tests locally?

larabr: The behavior for the two commands was actually equivalent, that's why I dropped the duplicate. I could revert to the two commands just for the sake of not having to worry about CI-specific changes (e.g. once we switch to test firefox beta in CI); but for now the `test-browser` would just be an alias of `test-browser:ci` basically

twiss: Oki :+1:

PR #1959, opened 2026-01-08
[v7] Improve whitespace handling in CSF and other armored messages

twiss:
- In cleartext signed messages, fix dash unescaping of empty lines.
- Fix the regex testing for an "empty line" to only ignore spaces and tabs, and not other whitespace.
- Simplify handling of trailing whitespace.

To be merged after https://github.com/openpgpjs/openpgpjs/pull/1884.

Diff: https://github.com/openpgpjs/openpgpjs/commit/4a1a09ddc90e86f6ae0c934252d22f48e8ca6a1a.

### DIFF
--- a/src/encoding/armor.js
+++ b/src/encoding/armor.js
@@ -34,7 +34,7 @@ import defaultConfig from '../config';
  * @private
  */
 function getType(text) {
-  const reHeader = /^-----BEGIN PGP (MESSAGE, PART \d+\/\d+|MESSAGE, PART \d+|SIGNED MESSAGE|MESSAGE|PUBLIC KEY BLOCK|PRIVATE KEY BLOCK|SIGNATURE)-----$/m;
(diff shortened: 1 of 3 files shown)


src/encoding/armor.js
```diff
@@ -36,7 +34,7 @@ import defaultConfig from '../config.ts';
  * @private
  */
 function getType(text) {
```
larabr, line 36:
```suggestion
function getType(headerLine) {
     if (headerLine.trimEnd().includes('\n') {
         throw new Error('Expected single input line');
     }
```

test/general/armor.js
```diff
@@ -64,7 +64,7 @@ export default () => describe('ASCII armor', function() {
     let msg = [
       '-----BEGIN PGP SIGNED MESSAGE-----',
       'Hash: SHA1',
-      ' \f\r\t\u00a0\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200a\u202f\u205f\u3000',
```
larabr, line 67: Shouldn't we test behavior with these chars?

test/general/armor.js
larabr, line 359: Why test different encryption types and key versions? I'd add comments to clarify what's actually being tested in the armored output

PR #2022, opened 2026-06-11
[v7] PQC: implement ML-KEM-768+X25519 and ML-DSA-65+Ed25519

twiss:
See PRs https://github.com/ProtonMail/openpgpjs/pull/10, https://github.com/ProtonMail/openpgpjs/pull/13, https://github.com/ProtonMail/openpgpjs/pull/19 for the incremental changes.

Reference: https://datatracker.ietf.org/doc/draft-ietf-openpgp-pqc/17/.


package.json
```diff
@@ -73,7 +73,9 @@
     "@noble/ciphers": "^1.3.0",
     "@noble/curves": "^1.9.7",
     "@noble/hashes": "^1.8.0",
+    "@noble/post-quantum": "^0.2.1",
     "@openpgp/jsdoc": "^4.0.4",
+    "@openpgp/seek-bzip": "^1.0.5-git",
```
larabr:
```suggestion
```
?

src/enums.js
```diff
@@ -96,7 +96,11 @@ export default {
     /** Ed25519 (Sign only) */
     ed25519: 27,
     /** Ed448 (Sign only) */
-    ed448: 28
+    ed448: 28,
+    /** Post-quantum ML-DSA-64 + Ed25519 (Sign only) */
+    pqc_mldsa_ed25519: 30,
+    /** Post-quantum ML-KEM-768 + X25519 (Encrypt only) */
+    pqc_mlkem_x25519: 35
```
twiss: IMHO we should name these
```suggestion
    /** Post-quantum ML-DSA-65 + Ed25519 (Sign only) */
    pqc_mldsa65_ed25519: 30,
    /** Post-quantum ML-KEM-768 + X25519 (Encrypt only) */
    pqc_mlkem768_x25519: 35
```
or maybe just
```suggestion
    /** ML-DSA-65 + Ed25519 (Sign only) */
    mldsa65_ed25519: 30,
    /** ML-KEM-768 + X25519 (Encrypt only) */
    mlkem768_x25519: 35
```

larabr: I'd keep the `pqc_` prefix and descriptors

larabr: Just noticed that the casing is inconsistent with the other enums, should be `pqcMldsa65Ed25519` etc

twiss: I prefer to match the spec here, this is a direct mapping of the IANA table and those don't have the PQC prefix so I wouldn't add them here either. I've updated the casing though :+1:

larabr: I think these algo names are a little obscure for users atm; that's why I'd keep PQC references explicit until the names become more popular; similarly if we also introduce SLH. We could keep aliases..? regardless, can you add back the "Post-quantum" in the comments pls?

twiss: > I'd keep PQC references explicit until the names become more popular

But we shouldn't change these enum names later, that's just introducing churn for no reason. Again, this is just a low-level mapping of the [IANA table](https://www.iana.org/assignments/openpgp/openpgp.xhtml#openpgp-public-key-algorithms) so we should match the names there. Then, for the higher-level API we can do something nicer like `type: 'pqc'`.

src/openpgp.js
```diff
@@ -38,8 +38,8 @@ import { checkKeyRequirements } from './key/helper.js';
  * The generated primary key will have signing capabilities. By default, one subkey with encryption capabilities is also generated.
  * @param {Object} options
  * @param {Object|Array<Object>} options.userIDs - User IDs as objects: `{ name: 'Jo Doe', email: 'info@jo.com' }`
- * @param {'ecc'|'rsa'|'curve448'|'curve25519'} [options.type='ecc'] - The primary key algorithm type: ECC (default for v4 keys), RSA, Curve448 or Curve25519 (new format, default for v6 keys).
- *                                                                     Note: Curve448 and Curve25519 (new format) are not widely supported yet.
+ * @param {'rsa'|'ecc'|'curve25519'|'curve448'|'pqc'} [options.type='ecc'] - The primary key algorithm type: RSA, ECC (default for v4 keys), Curve25519 (new format, default for v6 keys), Curve448 or PQC.
+ *                                                                           Note: Curve25519 (new format), Curve448 and PQC are not widely supported yet.
```
larabr, line 43: Should document that PQC means mlkem+x25519 as of now; then in a separate MR we should add a more specific selector (similar to the `curve` input)

src/enums.js
larabr: But these enums are meant to be accessed by users when setting e.g. some config options (blacklisted algorithms for instance). So it's important to keep a meaningful description in the jsdoc at least.

twiss: But we also don't document that x25519 is ECC for example. If users are getting that much into the weed that they want to blacklist PQC algorithms for some weird reason, they should be savvy enough to know which algorithms are PQC.

larabr (approved):

PR #2055, opened 2026-09-01
Fix reading an empty trailing subpacket in v6 signatures

twiss:
Drive-by: add some tests for a key containing an unknown subkey and a trailing empty subpacket, both with and without `ignoreUnsupportedPackets: false`.

### DIFF
--- a/test/general/openpgp.js
+++ b/test/general/openpgp.js
@@ -917,6 +917,103 @@ me8oJym/ILMHRGIc4JvY9+2wE5U1FBYTsze3WnVH5dP5mfA2Uk83TR5KewKANsb4kl/OEPlADWdR
 =c2mQ
 -----END PGP PRIVATE KEY BLOCK-----`;
 
+// v6 Ed25519 public key with X25519 and experimental ML-KEM-768+ECDH-NIST-P-384 (algorithm ID=100) subkeys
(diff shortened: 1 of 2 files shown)


test/general/openpgp.js
```diff
@@ -917,6 +917,101 @@ me8oJym/ILMHRGIc4JvY9+2wE5U1FBYTsze3WnVH5dP5mfA2Uk83TR5KewKANsb4kl/OEPlADWdR
 =c2mQ
 -----END PGP PRIVATE KEY BLOCK-----`;
 
+const experimentalPqcNistReadyPubKey = `-----BEGIN PGP PUBLIC KEY BLOCK-----
```
larabr, line 921: Can you maybe write a comment about what the key structure actually is? Or where are these from? On the sequoia dumper I only see X25519/Ed25519 keys

Also, once NIST support is added I guess these tests will need updating? I think creating a key with fake algo IDs might be better?

twiss: This key is from https://sequoia-pgp.gitlab.io/openpgp-interoperability-test-suite/pqc-nist-bp.html#Encrypt-Decrypt_roundtrip_with_v6_Ed25519_X25519_ML-KEM-768_ECDH-NIST-P-384_key and has a subkey with experimental algorithm ID 100, so it won't need updating unless we add support for that (which we shouldn't do in a production release).

larabr (approved):

PR #2059, opened 2026-09-11
Fix encrypting more than 4GB of data using AEAD

larabr:
And other internal improvements (see commits).

### DIFF
--- a/src/packet/sym_encrypted_integrity_protected_data.js
+++ b/src/packet/sym_encrypted_integrity_protected_data.js
@@ -323,6 +323,11 @@ export async function runAEAD(packet, fn, key, data) {
             nonce[iv.length - 8 + i] ^= chunkIndexArray[i];
           }
         }
+        // AES-GCM must not be used for more than 2^32 invocations (chunks) under the same key;
+        // OCB and EAX have much larger limits, so no equivalent check is needed for them.
+        if ((packet.aeadAlgorithm === enums.aead.gcm || packet.aeadAlgorithm === enums.aead.experimentalGCM) && chunkIndex >= 2 ** 32) {
+          throw new Error('Exceeded max number of AES-GCM chunks');
+        }
         if (!chunkIndex || chunk.length) {
           reader.unshift(finalChunk);
           cryptedPromise = modeInstance[fn](chunk, nonce, adataArray);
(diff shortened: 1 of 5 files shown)


src/packet/sym_encrypted_integrity_protected_data.js
```diff
@@ -316,7 +337,7 @@ export async function runAEAD(packet, fn, key, data) {
           // After the last chunk, we either encrypt a final, empty
           // data chunk to get the final authentication tag or
           // validate that final authentication tag.
-          adataView.setInt32(5 + chunkIndexSizeIfAEADEP + 4, cryptedBytes); // Should be setInt64(5 + chunkIndexSizeIfAEADEP, ...)
+          setInt64(adataView, 5 + chunkIndexSizeIfAEADEP, cryptedBytes);
```
twiss, line 340: [`setBigUint64`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/setBigUint64) exists nowadays, so at this point I would just change `cryptedBytes` to be a `BigInt` and use that directly, tbh.

src/packet/sym_encrypted_integrity_protected_data.js
```diff
@@ -260,7 +260,7 @@ export async function runAEAD(packet, fn, key, data) {
   const adataView = new DataView(adataBuffer);
   const chunkIndexArray = new Uint8Array(adataBuffer, 5, 8);
   adataArray.set([0xC0 | packet.constructor.tag, packet.version, packet.cipherAlgorithm, packet.aeadAlgorithm, packet.chunkSizeByte], 0);
-  let chunkIndex = 0;
+  let chunkIndex = BigInt(0);
   let latestPromise = Promise.resolve();
   let cryptedBytes = 0;
```
twiss: Probably it's safer to make this a BigInt from the start too:
```suggestion
  let cryptedBytes = BigInt(0);
```

src/packet/sym_encrypted_integrity_protected_data.js
twiss:
```suggestion
        cryptedBytes += BigInt(chunk.length - tagLengthIfDecrypting);
```

src/packet/sym_encrypted_integrity_protected_data.js
```diff
@@ -316,7 +316,7 @@ export async function runAEAD(packet, fn, key, data) {
           // After the last chunk, we either encrypt a final, empty
           // data chunk to get the final authentication tag or
           // validate that final authentication tag.
-          adataView.setInt32(5 + chunkIndexSizeIfAEADEP + 4, cryptedBytes); // Should be setInt64(5 + chunkIndexSizeIfAEADEP, ...)
+          adataView.setBigUint64(5 + chunkIndexSizeIfAEADEP, BigInt(cryptedBytes));
```
twiss:
```suggestion
          adataView.setBigUint64(5 + chunkIndexSizeIfAEADEP, cryptedBytes);
```

src/packet/sym_encrypted_integrity_protected_data.js
larabr: I was actually just seeing that converting numbers to bigints is faster than working with the latter; I don't think we need to worry about the max-safe-integer limit in this setting, so I'd just go back to numbers for the index too

twiss (approved):

PR #2060, opened 2026-09-12
Fix secret key packet containment

liyander:
## Summary

- use the existing `secretKey` and `secretSubkey` packet enum names when tracking key-packet boundaries
- extend the unknown-subkey regression test to verify trailing binding signatures are discarded for both public and secret keys

## Testing

- `npm test -- --grep "unknown subkey"` (2 passing)
- `npm test` (1040 passing, 3 pending; one unrelated Windows line-ending failure in the Node streaming roundtrip test)


larabr (approved):
Thanks for spotting and fixing this!

PR #2061, opened 2026-09-14
Fix parsing of four-octet packet lengths of 2**31 and above

### LINKED ISSUE #2058: Packet lengths ≥ 2^31 parse as negative (signed `<< 24`), breaking decryption of messages > 2 GiB
# Body

- OpenPGP.js version: 6.3.1 (also present on `main` @ `d35228c`)
- Affected platform (Browser or Node.js version): Node.js v25.9.0 — but the code path is platform-independent

## Description

Both 4-byte packet-length parsers in `src/packet/packet.js` build the length with signed 32-bit bitwise operators:

```js
packetLength = (await reader.readByte() << 24) | (await reader.readByte() << 16) | (await reader.readByte() << 8) | await reader.readByte();
```

- `src/packet/packet.js:183` — old-format, length-type 2 (RFC 4880 §4.2.1)
- `src/packet/packet.js:218` — new-format, 5-octet length (RFC 4880 §4.2.2.3)

When the first length octet has its high bit set (declared length ≥ 2^31), `<< 24` produces a **negative** result. The subsequent `if (packetLength > 0)` check is then false, so the packet body is never read, and parsing fails with a misleading `TypeError` rather than a length error.

Note that `util.readNumber()` — used for the 4-byte `type === 255` case in the same file — already avoids this by accumulating with `256 ** i` multiplication instead of shifts, so the two paths disagree about whether a 4-byte length may exceed 2^31.

## Steps to reproduce

No large file is needed: the defect is in parsing the *declared* length, so a 13-byte packet that merely claims 2^31 bytes triggers it.

```js
import { readMessage } from 'openpgp';

const header = Buffer.alloc(5);
header[0] = 0x80 | (11 << 2) | 2;                  // old format, tag 11, length type 2
const body = Buffer.from([0x62, 0x00, 0, 0, 0, 0, 0x41, 0x41]); // 'b', no filename, date, "AA"

for (const declared of [0x7fffffff, 0x80000000]) {
  header.writeUInt32BE(declared, 1);
  try {
    await readMessage({ binaryMessage: new Uint8Array(Buffer.concat([header, body])) });
  } catch (err) {
    console.log(declared, '->', err.message);
  }
}
```

The body is deliberately shorter than declared, so the only correct outcome in both cases is a truncation error.

Refer to [repro.js](https://github.com/user-attachments/files/32088688/repro.js) for details.

## Actual output

```
2147483647 -> Parsing LiteralDataPacket failed: Unexpected end of packet
2147483648 -> Parsing LiteralDataPacket failed: Cannot read properties of undefined (reading 'length')
```

## Expected output

Both should report the truncation, e.g. `Unexpected end of packet`. At 2^31 the length itself is
parsed incorrectly, so the error describes a symptom rather than the cause.

## Real-world impact

This surfaced while stream-decrypting files in AWS Lambda. `gpg`-produced ciphertext above 2^31 - 1 bytes (2,147,483,647) fails, while a 2,131,602,763-byte file with the same key and command decrypts fine. In Lambda the failure is an uncatchable process abort ~450 ms in (`Runtime.ExitError`, 128 MB used, no OOM and no timeout) — nothing allocates, because the body is skipped rather than read. That makes it hard to diagnose: the visible error points at `LiteralDataPacket` and mentions no length at all.

## Suggested fix

Coercing both sites to unsigned resolves it:

```js
packetLength = ((await reader.readByte() << 24) | (await reader.readByte() << 16) | (await reader.readByte() << 8) | await reader.readByte()) >>> 0;
```

Using `util.readNumber()` at both sites would also work and would match the existing
`type === 255` path.

I verified the `>>> 0` change locally against `gpg`-produced fixtures: a 2,148,386,963-byte ciphertext that previously failed decrypts correctly (~155 s, peak RSS 833 MB – 1.5 GB, i.e. memory stays bounded and streaming is unaffected), and reverting the change reproduces the failure. Happy to open a PR with a regression test built on the 13-byte fixture above if that would help.

## Possibly related

#1012 ("Streaming decryption fails on files larger than 2GiB", closed) reported a >2 GiB streaming failure with a different symptom (`Invalid typed array length`). I could not find a linked PR or fix commit on that issue, so this length-parsing path may be an unresolved part of the same underlying limit.

AaronLuo3296:
Both four-octet packet length parsers built the length with signed 32-bit bitwise operators, so a declared length with the high bit set came out negative. The `packetLength > 0` check below then skipped the packet body entirely, and parsing failed with an unrelated TypeError rather than reading the packet or reporting a length problem.

This affects old-format length type 2 (RFC 4880 section 4.2.1) and new-format five-octet lengths (section 4.2.2.3), and makes messages declaring 2 GiB or more undecryptable.

Fixes #2058


twiss (approved):

larabr:
Thanks for the fix, I took it over to apply some minor changes, to include this in the upcoming patch release.

PR #2062, opened 2026-09-20
fix(message): pass date to createSignaturePackets in signDetached

u9822914324-web:
## The bug

`createSignaturePackets` is declared as:

```js
createSignaturePackets(literalDataPacket, signingKeys, recipientKeys, signature, signingKeyIDs, date, signingUserIDs, recipientUserIDs, notations, detached, config)
```

`Message.signDetached` declares an extra `recipientKeyIDs` parameter in position 5 and forwards it where `date` belongs, shifting the last three arguments by one. So calling it with defaults passes `[]` as the signing date:

```js
const message = await createMessage({ text: 'hello' });
await message.signDetached([privateKey]);
// Error: Could not find valid key signature in key <id>:
//        Signature creation time is in the future
```

`Message.sign` is otherwise identical and forwards correctly, which is why `message.sign([privateKey])` works on the same key at the same instant.

## Why this is a drift, not a design choice

The JSDoc directly above `signDetached` already documents the parameters in `createSignaturePackets` order — `signingKeys, recipientKeys, signature, signingKeyIDs, date, signingUserIDs, recipientUserIDs, notations, config` — with no `recipientKeyIDs`. The implementation had drifted from its own documentation. `recipientKeyIDs` was never a parameter of the helper, so it had no effect beyond causing the shift; nothing is lost by removing it.

## Effect on the public API

`src/openpgp.js:432` is the only in-tree caller, and it already passes arguments in `Message.sign` order. That meant `sign({ detached: true })` worked by coincidence: the `date` it passed in position 5 landed in the helper's `date` slot *through* the shifted parameter. Two offsetting mistakes cancelled out.

After this change that call site is correct by construction, and its behaviour is unchanged — verified both before and after.

## Testing

- New regression test in `test/general/openpgp.js`. It fails on `main` with the exact error above and passes with this change.
- Full suite: **1042 passing, 0 failing** (1041 before — the delta is the new test).
- `eslint` on both touched files: 0 errors.
- Manually confirmed `sign({ detached: true })` and `sign()` still produce valid signatures.

Prepared with AI assistance (Claude); the argument mismatch was found by diffing `signDetached` against `sign` and confirmed against `createSignaturePackets`, then verified by the failing/passing regression test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


test/general/openpgp.js
```diff
@@ -2283,6 +2283,24 @@ BdPq
...
+
+      // signDetached used to forward `recipientKeyIDs` into the `date` parameter of
+      // createSignaturePackets, so leaving `date` unset passed an array where a Date was
+      // expected and threw 'Signature creation time is in the future'.
+      const signature = await message.signDetached([privateKey], undefined, undefined, undefined, undefined, undefined, undefined, undefined, config);
+      expect(signature.packets.filterByTag(openpgp.enums.packet.signature)).to.have.length(1);
+      const verified = await openpgp.verify({ message, signature, verificationKeys: privateKey, expectSigned: true, config });
+      expect(verified.data).to.equal(text);
+    });
+
```
twiss: This test was added between two tests that are strongly related to each other, but that this test isn't really related to. Arguably it shouldn't go in this file at all since it isn't related to the high-level `openpgp.*` API. How about putting it in test/general/signature.js instead?

twiss (commented):
Thanks for the PR! The core of it looks good to me; one nitpick about the added test below.

twiss (approved):

larabr:
Thanks for the MR, I took it over to release in the upcoming patch; dropped the test since I don't think it generalizes enough to meaningfully test functionality, aside from confirming the specific fix. Still, I followed up with https://github.com/openpgpjs/openpgpjs/pull/2065 to try and mitigate this class of issues in the future, as we transition to TS which is the ultimate safeguard.

PR #2063, opened 2026-09-21
types: declare applyRevocationCertificate, getPrimarySelfSignature, verifyDetached

u9822914324-web:
## What

Five public methods are implemented and documented but absent from `src/index.d.ts`, so TypeScript consumers cannot call them without casting:

| Method | Implementation |
|---|---|
| `Key.applyRevocationCertificate` | `src/key/key.js:643` |
| `Key.getPrimarySelfSignature` | `src/key/key.js:465` |
| `Message.verifyDetached` | `src/message.js:646` |
| `Message.decryptSessionKeys` | `src/message.js:170` |
| `Message.compress` | `src/message.js:539` |

## How the gaps were found

By diffing the methods defined on every class in `src/**/*.js` against the corresponding class blocks in `src/index.d.ts`, rather than from the issue tracker. The raw diff produced a longer list which I filtered down:

- `isPrivate`, `toPublic`, `armor` — inherited from `class Key`, already declared
- `validate`, `hasSameFingerprintAs` — already declared
- `clone`, `packetListToStructure`, `toPacketList` — internal, deliberately left out
- `clearPrivateParams` — already covered by #2029
- `generateSessionKey`, `encryptSessionKey` — module-level functions, not methods on `Message`

`Message.decryptSessionKeys` and `Message.compress` were easy to miss because the *names* resolve in the file: a module-level `decryptSessionKeys` function is declared, and `compress` appears in `preferredCompressionAlgorithms`. Neither is the method on the class.

## Accuracy

Signatures come from the implementations, not the JSDoc. That distinction matters for `verifyDetached`: it is implemented as `verifyDetached(signature, verificationKeys, date, config)`, while its JSDoc lists the verification keys first. The declaration follows the implementation.

`applyRevocationCertificate` returns `this.clone()` with the revocation appended, so it is declared `Promise<this>` to preserve `PublicKey` / `PrivateKey` through the call rather than widening to `Key`.

`decryptSessionKeys` is declared as returning `DecryptedSessionKey[]`, matching the declaration already used for the module-level function.

## Scope note

`Message.signDetached` is also undeclared, but its parameter list is wrong in a way that is being corrected separately in #2062. I left it out of this PR so the two can be reviewed independently; it can be declared once that lands.

## Testing

Assertions added to `test/typescript/definitions.ts`, which is both type-checked and executed by `test-type-definitions`. Verified against a clean `main` build (no dependency on #2062):

```
> tsc --project test/typescript/tsconfig.test.json && ts-node ... definitions.ts
TypeScript definitions are correct
```

Types-only; no runtime code touched.

Prepared with AI assistance (Claude); each declaration was checked against the implementation it describes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


larabr:
Hi, only a subset of methods are covered by TS, we don't necessarily want to expose all internal methods. Do you have a use case for these functions you added?

